PR #127, opened 2023-07-18
Minor fixes and additions

### LINKED ISSUE #112: Smoothing over a region border causes artifact
Smoothing over a region border causes artifact.

In the meantime, sculptors can use the flatten tool to erase the artifact, and another tool like reduce to get a similar smoothing effect.

image

### LINKED ISSUE #120: Add option to turn off shadows
While it good to have shadows for a better looking terrain, the performance can be worst and make a difference

### LINKED ISSUE #126: Remove terrain node pointer from storage
Bad design.
Instead make '_update_regions()' emit a signal and connect that to Terrain node to update AABBs.

outobugi:
Closes #126
Closes #120
Fixes #112

Removed _update_visibility, _update_world and put all that into a new _update_instances method.

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -207,61 +207,51 @@ void Terrain3D::_destroy_collision() {
 
 /**
  * Make all mesh instances visible or not
+ * Update all mesh instances with the new world scenario so they appear
  */
-void Terrain3D::_update_visibility() {
-	if (!is_inside_tree() || !_valid) {
+void Terrain3D::_update_instances() {
+	if (!is_inside_tree() || !_valid || !_is_inside_world) {
 		return;
 	}
+	if (_static_body.is_valid()) {
+		RID _space = get_world_3d()->get_space();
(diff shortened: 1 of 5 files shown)


TokisanGames (approved):
Very good, thanks. Renamed visual_layers and layers to render_layers, _shadow_casting_setting to _shadow_casting and grouped them, and other renames. Tested collision & render layers and shadow casting.

e.g. since we have both render layers and collision layers, they should be explicitly named. 

Also the render layer is not a mask in the way defined by other places in godot docs. The rendererserver is inconsistent. Layers are where objects live (renderer or physics). Masks are filters that define what layers they interact with. 

The object and camera both have their own render layer they live on. The object has no render mask filter. The camera has it, which determines which layers it sees (interaction). 

Collision objects have both layers and mask filters as they both live somewhere and choose what to interact with.

Confusingly all layers and masks are bit**masks** as each bit defines which layer they live on or filter. So I renamed things according to Godot's own definitions of layers and masks, though their rendererdocs and function calls are inconsistent with the rest.

Issue #108, opened 2023-07-07
Add false color debug modes to shader


TokisanGames:
Options include:
* ~~Grid ( w/ scale)~~
* ~~Grey albedo~~
* ~~Color map~~
* ~~Control map~~
* ~~Roughness map~~
* ~~Height: `ALBEDO = vec3(get_height(UV2/100.);`~~

* Optional Terrain normals w/ any of the following:
* Terrain normal map
* Texture normal maps
* Slope

There are several here
https://github.com/TokisanGames/gterrain/blob/570951f8fa16a94ea186df8ca3791fee5c984375/addons/gterrain/shaders/editor.gdshaderinc#L40

~~Best to wait until #117 is done~~ Done.

PR #144, opened 2023-07-21
Extract shader from C++ to separate files

### LINKED ISSUE #117: Move shader in C++ to standalone file
Let's put the shader into a standalone file with either:

* `static const String vertex_shader_text { #include "glmain.vert" }; `
* `R"( #version 110 )"`

The file will likely need sentinels that we can find and replace with other snippets like noise. These could be separate files. We might also be able to take advantage of godot's shader includes.

https://stackoverflow.com/questions/43256465/include-a-file-as-a-string (comments by x4444)
https://stackoverflow.com/questions/410980/include-a-text-file-in-a-c-program-as-a-char/25021520#25021520

TokisanGames:
This PR:
* Moves all shader code out of C++ to external files
* Files are loaded by the preprocessor
* Optional Inserts can be declared or inserted in glsl through `//INSERT: ID` tags

e.g.
 In `world_noise.glsl`, and `debug_views.glsl`, inserts like `//INSERT: WORLD_NOISE1` are defined and loaded into memory
In `main.glsl`, the inserts are placed in the base shader. The same tag is used, but only applied in this file.

* The show_grid checkbox turns it on. The first of many [debug views](#108). If all surfaces are deleted, or a new terrain storage is created, show_grid turns on automatically. However the user can turn it off again, showing black.

All of this works as is. The only thing that is a bit hacky is the getting the grid to turn on when creating a new storage. I want show_grid to default to false. However if it is set to true in the header, constructor, or at any time before a certain unknown point, it will set the default to true. I had to place it in Terrain3D::set_storage() to initiate it, and used update_surface_values() there as the easiest way in to do so, placing the surface count check down in update_surface_data.

Fixes #117

### DIFF
--- a/src/shaders/debug_views.glsl
+++ b/src/shaders/debug_views.glsl
@@ -1,0 +1,13 @@
+R"(
+
+//INSERT: GRID
+	// Display a checkered grid
+	vec2 p = UV * 1.0; // scale
+	vec2 ddx = dFdx(p);
+	vec2 ddy = dFdy(p);
+	vec2 w = max(abs(ddx), abs(ddy)) + 0.01;
+	vec2 i = 2.0 * (abs(fract((p - 0.5 * w) / 2.0) - 0.5) - abs(fract((p + 0.5 * w) / 2.0) - 0.5)) / w;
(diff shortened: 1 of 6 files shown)


outobugi (approved):
Nice.

Issue #164, opened 2023-07-26
Export breaks on surfaces without textures


TokisanGames:
Export breaks if any surface is missing textures. All textures are gone, rendered as the albedo color only plus the colormap, and the console reports:
```
ERROR: Condition "!_image_compress_bc_func" is true. Returning: ERR_UNAVAILABLE
   at: compress_from_channels (core/io/image.cpp:2637)
ERROR: Condition "p_layers[i]->get_format() != valid_format" is true.
   at: texture_2d_layered_initialize (servers/rendering/renderer_rd/storage_rd/texture_storage.cpp:884)
ERROR: Condition "!_image_compress_bc_func" is true. Returning: ERR_UNAVAILABLE
   at: compress_from_channels (core/io/image.cpp:2637)
ERROR: Condition "!_image_compress_bc_func" is true. Returning: ERR_UNAVAILABLE
   at: compress_from_channels (core/io/image.cpp:2637)
ERROR: Condition "p_layers[i]->get_format() != valid_format" is true.
   at: texture_2d_layered_initialize (servers/rendering/renderer_rd/storage_rd/texture_storage.cpp:884)
ERROR: Attempting to use an uninitialized RID
   at: (./core/templates/rid_owner.h:199)
ERROR: Attempting to use an uninitialized RID
   at: (./core/templates/rid_owner.h:199)
```

Upstream issue
https://github.com/godotengine/godot/issues/79932

TokisanGames:
Godot devs report the compression libraries are not in the exports. They are discussing the option to include them. In the meantime, we will just have to ensure invalid surfaces are not included. #166 adds warnings in the editor, and a note was added to the wiki. This ticket can remain open to track the upstream issue.

PR #166, opened 2023-07-26
Warn about null textures breaking exported games

TokisanGames:
Any null textures in surfaces breaks exports. Releases crash, debugs turn textures white. This PR mitigates #164, for now.

This PR:
* Warns about null textures breaking exported games in the console
* Warns in the scenetree on the terrain3d node
* Triggers surface list updates (and warnings check) on every inspector change


TokisanGames:
I found a better solution. Superseded by #168.

PR #170, opened 2023-07-31
Fix normals

### LINKED ISSUE #169: Fragment & Texture Normals are wrong
Update: 

TLDR:
* Terrain binormals/tangents were wrong, being calculated off of viewspace instead of worldspace. 
* We can't calculate terrain B/N/T in vertex() without creating an artifact on the lod0/1 border. The other differences are very subtle; a very slight reshaping of shadows on steep inclines that I can't say is better, just different and irrelevant.
* TextureGrad and Texture can be interchanged with identical results, as shown above in update 2. Through some casual testing, Texture might be up to 4% faster. Not the problem here. Saving for #26.
* Rotation is part of the problem. Even when set to 0 it rotates the textures 90 with incorrect normals. Removing.
* Normalizing texture normals is the remainder of the problem. Removing.

All fixed in #170.

----

A cube with the standard material compared to the terrain a few cm lower. Top perspective.

image

Side angle comparing the same rock

image

On the standard material the normals on this rock do not respond to camera angle. The light is always the same. On the terrain, the light subtly changes based upon camera angle. The amount and coverage of the light should be identical between the left and right images, as it is for the cube. Every face should be identically lit.

image

What I found is that if I used a basic terrain shader:
```
fragment() 
{
	float scale=.2;
	vec4 alb = texture(texture_array_albedo, vec3(UV*scale, 11));
	ALBEDO = alb.rgb;
	vec4 nrm = texture(texture_array_normal, vec3(UV*scale, 11));
	NORMAL_MAP = nrm.rgb;
	ROUGHNESS = nrm.a;
}

```

No fragment normals. and in vertex make sure tangent and binormal are calculated

```
vertex() {
	NORMAL = vec3(0, 1, 0);
	TANGENT = cross(NORMAL, vec3(0, 0, 1));
	BINORMAL = cross(NORMAL, TANGENT);
}

```
Then we get the proper look:
image

image

So it appears there's a problem with our terrain tangents and binormals in fragment, which we do need.

Also we can provide the option to use height as AO for free.

--------------
Update:
I've further found that binormal and tangent can be calculated in fragment, however they must do so before being converted to view space.
```
fragment()
	vec3 out_tangent, out_binormal;
	vec3 normal = get_normal(UV2, out_tangent, out_binormal);
	NORMAL = mat3(VIEW_MATRIX) * normal;
	TANGENT = mat3(VIEW_MATRIX) * out_tangent;
	BINORMAL = mat3(VIEW_MATRIX) * out_binormal;
```

And this now gives us proper looking vertex and fragment normals using a terrain covered with lookups from `texture()`

image

However, the current shader looks up with `texturegrad()` plus a weighting for texture blending. Using the above n/t/b calculations, the normals are more correct, but they're a bit flat. I can double the NORMAL_MAP_DEPTH but it's not a perfect match.

image

image

image

-------- 
Update 2:

I've found that these two produce identical results, and that normalizing the normals which fixed an artifact before, is incorrect. 

```
	float scale=.2;

	vec2 matUV = UV * scale;
	vec2 ddx = dFdx(matUV);
	vec2 ddy = dFdy(matUV);
	vec4 albd = vec4(1.0);
	vec4 norm = vec4(0.5);
	albd = textureGrad(texture_array_albedo, vec3(matUV, 11), ddx, ddy);
	norm = textureGrad(texture_array_normal, vec3(matUV, 11), ddx, ddy);
//	norm.rgb = normalize(norm.rgb);
	NORMAL_MAP = norm.rgb;
	ALBEDO = albd.rgb;
	ROUGHNESS = norm.a;
```
```
	vec4 alb = texture(texture_array_albedo, vec3(UV*scale, 11));
	ALBEDO = alb.rgb;
	vec4 nrm = texture(texture_array_normal, vec3(UV*scale, 11));
	NORMAL_MAP = nrm.rgb;
	ROUGHNESS = nrm.a;

```

In get_material the normals are normalized, and as mentioned, fixed an artifact before but kills normals.  You can see it greatly weakens the normal map. It's clear at the bottom of the hill.

image
image

image

We can remove the normalization, but that will allow the previous artifact to surface. Perhaps we can normalize  later after the weighting is applied.

TokisanGames:
I found three areas causing broken normals:
* fragment() Terrain binormals/tangent
* get_material() rotation
* get_material() normalizing texture normals

Fixes #169 

If you merge this instead of approving, please use `Merge pull request: Create a merge commit`.

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -94,27 +94,22 @@ vec3 get_normal(vec2 uv, out vec3 tangent, out vec3 binormal) {
 	return normal;
 }
 
-// One big mess here.Optimized version of what it was in my GDScript terrain plugin.- outobugi
-// Using 'else' caused fps drops.If - else works the same as a ternary, where both outcomes are evaluated. Right?
 vec4 get_material(vec2 uv, vec4 index, vec2 uv_center, float weight, inout float total_weight, inout vec4 out_normal) {
 	float material = index.r * 255.0;
 	float materialOverlay = index.g * 255.0;
 	float r = random(uv_center) * PI;
 	float rand = r * texture_uv_rotation_array[int(material)];
 	float rand2 = r * texture_uv_rotation_array[int(materialOverlay)];
-	vec2 rot = vec2(sin(rand), cos(rand));
-	vec2 rot2 = vec2(sin(rand2), cos(rand2));
+	vec2 rot = vec2(cos(rand), sin(rand));
+	vec2 rot2 = vec2(cos(rand2), sin(rand2));
 	vec2 matUV = rotate(uv, rot.x, rot.y) * texture_uv_scale_array[int(material)];
 	vec2 matUV2 = rotate(uv, rot2.x, rot2.y) * texture_uv_scale_array[int(materialOverlay)];
 	vec2 ddx = dFdx(matUV);
 	vec2 ddy = dFdy(matUV);
-	vec4 albedo = vec4(1.0);
-	vec4 normal = vec4(0.5);
 
-	albedo = textureGrad(texture_array_albedo, vec3(matUV, material), ddx, ddy);
+	vec4 albedo = textureGrad(texture_array_albedo, vec3(matUV, material), ddx, ddy);
 	albedo.rgb *= texture_color_array[int(material)].rgb;
-	normal = textureGrad(texture_array_normal, vec3(matUV, material), ddx, ddy);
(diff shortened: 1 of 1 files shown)


outobugi (changes requested):
Removing the rotation isn't a proper solution for fixing normals. :(

TokisanGames:
This version restores rotation, but completely nullifies it when rotation==0, along with the other two changes to fix normals.

outobugi (approved):

PR #176, opened 2023-08-06
Add makeshift baker

SlashScreen:


TokisanGames:
Can you post instructions for use here so I can set it up in the wiki? 
Thanks.

SlashScreen:
Here are the instructions:
Terrain3D does not yet have a proper implementation for navmeshing and colliders. However, there is a workaround solution- a multithreaded navigation baker that will bake the entire terrain at once. Here's how to use it:
1. Create a `ChunkBaker` node, and place all of the geometry you want to be included in the bake as children underneath it, like you would a regular `NavigationRegion3D`. This includes the terrain you wish to bake.
2. Provide a template navmesh resource. Set it up as you would any conventional bake, apart from a few settings:
- In Geometry > Parsed Geometry Type, select Static Colliders or Both.
- In Filters > Baking AABB, change the `w`, `h`, and `d` values to select the size of the chunks that will be baked. Only the `w` and `d` values matter, but the `h` value should be tall enough to encompass the entire height of your terrain. The larger the chunk size, the more time it will take for each chunk to bake, but longer each chunk will take. A good starting point is (64, 512, 64).
3. Provide the rest of the settings.
- Bake Path: This is the folder where it will dump the baked navmesh once complete.
- Terrain: This a reference to the terrain node you want to bake.
- Bake Region: This is the region (Large square you can add in the terrain menu) you want to bake, referred to by index. -1 will bake all regions. Index 0 is the center.
- Pretty Print: Bake progress will be printed in a fancy way into the output as the bake progresses. Godot will rarely be running smooth enough to watch this go by in-editor, however, so if you want to save the processing time required to print, you can safely turn it off.
4. Hit the "Bake terrain navigation" button at the top. Godot will freeze for a bit as the scene geometry is parsed on the main thread. Then, your computer will become barely usable for a few minutes, as all cores will be maxed to get the bake done as soon as possible. On an 8 core Ryzen 7 5800X, one region of high complexity took 30 seconds to a minute. Each region is done sequentially, so bake time scales linearly with regions. Godot will become unstable during the bake.
5. Once the bake is finished, It has created a resource file in your bake path. This is a large bundle of the navmeshes your chunks have created. **There is no need to include this in an exported game**. 
6. Create a `NavmeshBundleUnpacker` node in your scene, It requires one resource: the bundle you just created. Assign it to the node, and hit "Unpack". After some time, you will find it has created a small army of `NavigationRegion3D`s, each corresponding to a baked chunk. Your terrain navmesh is now in the scene. You bundle is no longer necessary to keep, if you don't want to.

The baking process is now complete. Congrats!

TokisanGames:
Superseded by #253 which has been merged. Thanks for the initial work. I'll credit you both in the release notes.

SlashScreen:
Good, nobody has to use it anymore.

PR #212, opened 2023-09-22
Improve installation instructions

svencan:
This adds the link to the releases in the installation sections, as people who skip the "run the demo" part tend to download the main branch instead of a release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,15 @@ See the [Wiki](https://github.com/TokisanGames/Terrain3D/wiki) for more details 
 6. If the demo scene doesn't open automatically, open `demo/Demo.tscn`. You should see a terrain. Run it. 
 
 ### Install Terrain3D in your own project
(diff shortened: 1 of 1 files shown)


TokisanGames:
Great, thanks. Please add this to the first line `, or [build the plugin from source](https://github.com/TokisanGames/Terrain3D/wiki/Building-From-Source).`

Generally I'm not so concerned about squashing, but if you were submitting to Godot they are religious about it. It will be good practice to learn and for fixing a change like this. Make the above change and commit. Then look at the PR workflow document to do an interactive rebase `git rebase -i HEAD~3`, change the first word of the second and third commits to `fixup` so it merges them all into one, then do a force push to overwrite the server branch.

svencan:
Changed as requested and squashed the commits. Thanks for the hint!

TokisanGames:
Great, thanks!

PR #219, opened 2023-09-30
Add iOS Support

### LINKED ISSUE #218: Add iOS Support
### Overview

I am new to Godot and Terrain3D, but have been experimenting.  I managed to mostly get iOS working without much trouble, maybe solving 2 of 3 issues/steps, after some setup from [building from source instructions](https://github.com/TokisanGames/Terrain3D/wiki/Building-From-Source).

### Steps

1. [Local Fix] Building iOS libraries
```
scons platform=ios target=template_debug
[...]
scons platform=ios target=template_release
```
Note: there are some deprecation warnings.

2. [Local Fix] Updating terrain.gdextension:

```
[...]
ios.debug = "res://addons/terrain_3d/bin/libterrain.ios.debug.universal.dylib"
ios.release = "res://addons/terrain_3d/bin/libterrain.ios.release.universal.dylib"
```

3. [WIP] Fixing textures

With the first two changes and the normal iOS export templates from the Godot editor, I am seeing a "chess squares" shader issue (as if it had no textures?) actually on device.  I have a separate C++ GDExtension moving the character and camera slightly to confirm code is running.  Also note shadows and occlusion look to be working.

image1
image0 (1)

### Other Thoughts
**Rendering**: This is with Godot Project Settings->General->Renderering->Renderer->Rendering Method options: forward_plus or mobile.  gl_compatibility shows an all grey plane, no height at all or squares, which is a different issue.

**Exporting**: I am not an artist, but one [suggestion](https://github.com/TokisanGames/Terrain3D/issues/217#issuecomment-1735106187) from that other issue was to investigate the texture exports themselves.   I followed [Texturing the Terrain](https://github.com/TokisanGames/Terrain3D/wiki/Setting-Up-Textures#texturing-the-terrain), using the highly recommended DDS settings using GIMP-2.10.  It renders fine in editor, export to mac and windows.

Screenshot 2023-09-26 at 9 27 17 PM

**Architectures**: I did this on an M1 Mac and updated to the latest XCode.  But I am not sure that matters as it builds universal.  arm64 seems to build fine:
```
scons platform=ios target=template_debug arch=arm64
[...]
scons platform=ios target=template_release arch=arm64
```

Though the resulting libraries are exactly the same size as the universal libraries.  I haven't tried with the App Store either. The mac libraries look to not use architecture, though now I wonder if that would help arm macs.

directedchaossoftware:
- Adds iOS universal library paths to terrain.gdextension
- Adds icon and project reference for iOS
- Modifies terrain_3d_texture to allow

Reviewer Edit: 
Fixes #218


TokisanGames:
Thank you, I will review and test a bit more. We won't be defaulting the demo project to etc2, but we can document that it's needed for ios and android so people can setup their own files.

Unity recommends:
Desktop: BC7 or DXT1 (BC1) / DXT5 (BC3) (Godot only supports up to DXT5)
IOS: ASTC or ETC2
Android: ASTC or ETC2
https://docs.unity3d.com/Manual/class-TextureImporterOverride.html

And Android says ETC2, which covers 87% of devices. ETC1 doesn't support alpha so it's not an option. ASTC is 77% of devices.
https://android-developers.googleblog.com/2020/09/improve-your-game-with-texture.html

directedchaossoftware:
Thanks for the help!  In particular I don't know much about texture formats (compression etc.), I just tinker to get something working.  It makes sense not to default the demo to etc2 (and you may want to change/remove the icon too?) -- there are more setup steps for mobile anyway.  I went ahead and removed the etc2 project change.  Let me know if you want something else to help get this in, but otherwise I'm quite happy to keep tinkering with this change in my project.

TokisanGames:
FYI, I will work on this and get it in as part of 0.8.4, after I'm done with changing the control map format, adding holes, and merging in the shader rewriting I've been doing. I need to test etc2 / astc and the image formats.

You can remove the icon and squash your commits.

I'll need to write up a wiki page for Mobile and let people know they need to enable this setting in their project settings: `textures/vram_compression/import_etc2_astc=true`. And other steps from your issue #218.

I also need to create a github action to build IOS automatically.

I have an old iphone 7 w/ v15.7.9 and no mac. If I have github build it, is there a way I can export the demo from my windows pc and load it on my phone?

Thanks

directedchaossoftware:
Cool, I'm happy to try those things.  I can sync/squash next week, but I think the icon is required to export to iOS.  I don't know about github build, but I don't think you can build for iPhone from Windows.  I think you need XCode on Mac OS and developer credentials.

TokisanGames:
Ah, you don't want to merge. There's almost never a reason to use `git merge`. If you used the `sync` button on github desktop, that creates a merge commit and you shouldn't use it. 

In the future, what you want to do is a rebase off of the latest commit as described in the godot pull request workflow.

https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#updating-your-branch


However, what you want to do now is an interactive rebase and delete that merge commit. Then you might need to do the rebase above if it is still needed.

https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase

TokisanGames:
I tested this today. I was delaying because I needed to test the ` && format != Image::FORMAT_RGBA8)` and was working on other areas.

This change allows the user to insert PNGs that have not been imported according to the docs. By default PNGs are listed as "Lossless" and are straight RGBA8. However when used this way, they don't work at all and produce errors in the console, so we can't allow that. 

When PNGs are properly imported as VRAM Compressed, an s3tc.ctex file is created. This is a DXT5 file.
```
$ file project/.godot/imported/rock030_alb_ht.png-ac17816f8b127f0c153387663d699b8f.s3tc.ctex
Godot 4 texture v1: 1024 x 1024, 10 mipmaps, DXT5
```

If the import etc2 is selected in the project settings, it also makes an etc2.ctex file. Both the DXT5 and ETC2 file exist in the `.godot/imported` folder. 
```
$ file project/.godot/imported/rock030_alb_ht.png-ac17816f8b127f0c153387663d699b8f.etc2.ctex
Godot 4 texture v1: 1024 x 1024, 10 mipmaps, ETC2_R11S
```

Godot has a format for this: `FORMAT_ETC2_R11S = 27`. Does this work in that format validation? 

There's also a variety of other options. I assume you can't get a console on ios. But you could load an exported PNG file and display the format on screen in the UI if 27 doesn't work.

In the inspector when I doubleclick the png, the DXT5 file displays. So the etc2 file must only be used on export. So yes, we can leave it on in the project settings as it won't affect desktop at all. In the documentation we can let them know etc2 must be enabled in project settings and on export.

I exported a PCK file as a zip and looked for rock028*. There are no ctex files for it. There are only dds files because they don't get imported(converted). So there is no DXT5 -> ETC2 conversion. Since DXT doesn't work on mobiles, that's why you got a checkered display.

So mobile games need to use PNG imported as VRAM Compressed. Not DDS.

The only thing that's confusing for me is how you got the DXT5 grass texture to show up in your [screenshot](https://github.com/TokisanGames/Terrain3D/issues/218#issuecomment-1741767166)? Did you convert the DDS file to another format in photoshop? It is not RGBA8, so why does your console report it as 5? 

Your export .pck is a zip file. What rock028* files do you have in .godot and in the file system? If it's just the rock028_alb_ht.dds then I'm totally confused how. I expect there should be an etc2.ctex file and a PNG version of the file or something in there.

TokisanGames:
I updated this PR:
* Removed merge commit
* Rebased
* Removed image format check, which has been superseded by #295. Since that PR allows mostly any format, without breaking the terrain, the contents of this PR should allow IOS to work.

Thanks for all of the testing before to produce this PR.

We need more testing on IOS now that these PRs are in.

PR #229, opened 2023-10-14
Custom uniforms

### LINKED ISSUE #86: Support custom uniforms
> How do I control the uniforms? The inspector needs to expose these. Perhaps you should make the override a ShaderMaterial instead.

Custom shaders were added in #83, also see #39. But we need a way to add custom uniforms that support any format regular shaders do: textures, colors, etc.

This should also expose any useful variables in the current shader. Such as UV scaling.

TokisanGames:
* Custom uniforms in shaders are now detected and displayed in the Terrain3DMaterial inspector. Uniforms that begin with _ are considered private and not displayed.
* All of the hard coded functions for worldnoise parameters are removed and instead the above pulls out the public world noise settings from the shader. It displays them whether they are part of a custom shader or the built in one.
* Uniforms work as the native data types (int, float, vec*, mat*, etc) and texture objects. I tested ImageTexture (dds), CompressedTexture (png), and NoiseTexture
* Shader includes work

Fixes #86

### DIFF
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -88,7 +88,7 @@ String Terrain3DMaterial::_parse_shader(String p_shader, String p_name, Array p_
 String Terrain3DMaterial::_generate_shader_code() {
 	LOG(INFO, "Generating default shader code");
 	Array excludes;
-	if (!_noise_enabled) {
+	if (!_world_noise_enabled) {
 		excludes.push_back("WORLD_NOISE1");
 		excludes.push_back("WORLD_NOISE2");
 	}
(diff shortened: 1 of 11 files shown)


src/terrain_3d_material.cpp
```diff
@@ -231,65 +236,112 @@ void Terrain3DMaterial::_update_texture_arrays(const Array &p_args) {
...
 		RS->material_set_shader(_material, _shader_override->get_rid());
+		LOG(DEBUG, "Mat rid: ", _material, ", _shader_override rid: ", _shader_override->get_rid());
 	} else {
 		RS->shader_set_code(_shader, _generate_shader_code());
 		RS->material_set_shader(_material, _shader);
+		LOG(DEBUG, "Mat rid: ", _material, ", _shader rid: ", _shader);
+	}
+	// Update custom shader params in RenderingServer
+	// Populate _active_params
+	_get_property_list(&(List<PropertyInfo>()));
```
phire, line 266: This results in a compile error on my machine

`src/terrain_3d_material.cpp:266:21: error: taking the address of a temporary object of type 'List<PropertyInfo>' [-Waddress-of-temporary]`

This works: 

```suggestion
	{
		auto temp_list = List<PropertyInfo>();
		_get_property_list(&temp_list);
	}
```
```c++

TokisanGames: Thanks for testing. Godot devs don't like auto, but I can put in a typed temp var. What compiler are you using? Apple clang?

phire: Yeah, Apple clang 15. Just the default setting from the godot_cpp SConstruct

Issue #30, opened 2023-01-28
Support doubles


TokisanGames:
Update: #232 has added the ability to build Terrain3D with double precision. 

To get it working you need to:
* Build Godot with `scons precision=double`
* [Regenerate godot-cpp bindings](https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/gdextension_cpp_example.html#building-the-c-bindings) with this new executable
* Build Terrain3D (which includes godot-cpp) with `scons precision=double custom_api_file=YOUR_CUSTOM_FILE`

@zuriel13 reported the demo is working fine using Godot 4.2.1 and Terrain3D 0.9.1-dev.

Using a single precision engine, the terrain and shader look fine when moving the camera even 100k units away from the origin. This may break down at extremes though and bears testing.

Shaders do not support double precision, however the engine does include some double emulation. Clayjohn wrote an article describing how to emulate double precision in shaders for positional elements. He writes that the camera and model transform matrices needed to be emulated, and that is now done automatically in the engine.

https://godotengine.org/article/emulating-double-precision-gpu-render-large-worlds/

It's possible that other elements of the shader will need to emulate double precision, but no issues are apparent, yet.

Related to regenerating bindings: https://github.com/godotengine/godot-cpp/issues/1386

-----

Include a build flag that supports doubles in both the shader and C++ where relevant. Probably only locations. This is done in godot source:

```
//math_defs.h
/**
 * The "Real" type is an abstract type used for real numbers, such as 1.5,
 * in contrast to integer numbers. Precision can be controlled with the
 * presence or absence of the REAL_T_IS_DOUBLE define.
 */
#ifdef REAL_T_IS_DOUBLE
typedef double real_t;
#else
typedef float real_t;
#endif

// Elsewhere
#ifdef REAL_T_IS_DOUBLE
	if (p_feature == "double") {
		return true;
	}
#else
	if (p_feature == "single") {
		return true;
	}
#endif // REAL_T_IS_DOUBLE
```

10k out
![image](https://user-images.githubusercontent.com/632766/215282486-d3c626d5-4eff-417c-95ba-6d6f0339bea0.png)

1M out, corruption and the mesh jitters all over the place
![image](https://user-images.githubusercontent.com/632766/215282470-6e6c1304-7970-43a4-959c-f0a830a78d68.png)

TokisanGames:
Double support has been tested and is reported working fine. Updated instructions in the first post.

[...]

PR #232, opened 2023-10-19
Change c++ code to work with doubles

SlashScreen:
solves half of issue #30 - doesn't touch shader though

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -564,7 +564,7 @@ void Terrain3D::set_cast_shadows(GeometryInstance3D::ShadowCastingSetting p_shad
 	_update_instances();
 }
 
-void Terrain3D::set_cull_margin(float p_margin) {
+void Terrain3D::set_cull_margin(real_t p_margin) {
 	LOG(INFO, "Setting extra cull margin: ", p_margin);
 	_cull_margin = p_margin;
 	update_aabbs();
@@ -603,10 +603,10 @@ void Terrain3D::snap(Vector3 p_cam_pos) {
 	int tile = 0;
 
 	for (int l = 0; l < _clipmap_levels; l++) {
-		float scale = float(1 << l);
+		real_t scale = real_t(1 << l);
 		Vector3 snapped_pos = (p_cam_pos / scale).floor() * scale;
-		Vector3 tile_size = Vector3(float(_clipmap_size << l), 0, float(_clipmap_size << l));
-		Vector3 base = snapped_pos - Vector3(float(_clipmap_size << (l + 1)), 0, float(_clipmap_size << (l + 1)));
+		Vector3 tile_size = Vector3(real_t(_clipmap_size << l), 0, real_t(_clipmap_size << l));
+		Vector3 base = snapped_pos - Vector3(real_t(_clipmap_size << (l + 1)), 0, real_t(_clipmap_size << (l + 1)));
 
 		// Position tiles
 		for (int x = 0; x < 4; x++) {
@@ -634,7 +634,7 @@ void Terrain3D::snap(Vector3 p_cam_pos) {
 		}
 
 		if (l != _clipmap_levels - 1) {
-			float next_scale = scale * 2.0f;
+			real_t next_scale = scale * 2.0f;
 			Vector3 next_snapped_pos = (p_cam_pos / next_scale).floor() * next_scale;
 
 			// Position trims
@@ -646,9 +646,9 @@ void Terrain3D::snap(Vector3 p_cam_pos) {
 				r |= d.x >= scale ? 0 : 2;
 				r |= d.z >= scale ? 0 : 1;
 
-				float rotations[4] = { 0.0, 270.0, 90, 180.0 };
+				real_t rotations[4] = { 0.0, 270.0, 90, 180.0 };
 
-				float angle = UtilityFunctions::deg_to_rad(rotations[r]);
+				real_t angle = UtilityFunctions::deg_to_rad(rotations[r]);
 				Transform3D t = Transform3D().rotated(Vector3(0, 1, 0), -angle);
 				t = t.scaled(Vector3(scale, 1, scale));
 				t.origin = tile_center;
@@ -657,7 +657,7 @@ void Terrain3D::snap(Vector3 p_cam_pos) {
 
 			// Position seams
 			{
-				Vector3 next_base = next_snapped_pos - Vector3(float(_clipmap_size << (l + 1)), 0, float(_clipmap_size << (l + 1)));
+				Vector3 next_base = next_snapped_pos - Vector3(real_t(_clipmap_size << (l + 1)), 0, real_t(_clipmap_size << (l + 1)));
 				Transform3D t = Transform3D().scaled(Vector3(scale, 1, scale));
 				t.origin = next_base;
 				RS->instance_set_transform(_data.seams[edge], t);
@@ -727,7 +727,7 @@ Vector3 Terrain3D::get_intersection(Vector3 p_position, Vector3 p_direction) {
 	Vector3 test_point = p_position;
 	p_direction.normalize();
 
-	float highest_dotp = 0.f;
+	real_t highest_dotp = 0.f;
 	Vector3 highest_point;
 
 	if (_storage.is_valid()) {
@@ -736,7 +736,7 @@ Vector3 Terrain3D::get_intersection(Vector3 p_position, Vector3 p_direction) {
 			test_point.y = _storage->get_height(test_point);
 			Vector3 test_vec = (test_point - p_position).normalized();
 
-			float test_dotp = p_direction.dot(test_vec);
+			real_t test_dotp = p_direction.dot(test_vec);
 			if (test_dotp > highest_dotp) {
 				highest_dotp = test_dotp;
 				highest_point = test_point;
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -54,7 +54,7 @@ void Terrain3DEditor::_operate_region(Vector3 p_global_position) {
 	}
 }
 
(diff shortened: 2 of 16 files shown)


SlashScreen:
Don't merge this until I've gone through everything

src/terrain_3d.cpp
```diff
@@ -564,7 +564,7 @@ void Terrain3D::set_cast_shadows(GeometryInstance3D::ShadowCastingSetting p_shad
 	_update_instances();
 }
 
-void Terrain3D::set_cull_margin(float p_margin) {
+void Terrain3D::set_cull_margin(real_t p_margin) {
```
TokisanGames, line 567: float for all usages of set/get cull_margin

src/terrain_3d.cpp
```diff
@@ -603,10 +603,10 @@ void Terrain3D::snap(Vector3 p_cam_pos) {
 	int tile = 0;
 
 	for (int l = 0; l < _clipmap_levels; l++) {
-		float scale = float(1 << l);
+		real_t scale = real_t(1 << l);
 		Vector3 snapped_pos = (p_cam_pos / scale).floor() * scale;
-		Vector3 tile_size = Vector3(float(_clipmap_size << l), 0, float(_clipmap_size << l));
-		Vector3 base = snapped_pos - Vector3(float(_clipmap_size << (l + 1)), 0, float(_clipmap_size << (l + 1)));
+		Vector3 tile_size = Vector3(real_t(_clipmap_size << l), 0, real_t(_clipmap_size << l));
+		Vector3 base = snapped_pos - Vector3(real_t(_clipmap_size << (l + 1)), 0, real_t(_clipmap_size << (l + 1)));
```
TokisanGames, line 609: All of this can be float. clipmap size will always be <1000

src/terrain_3d_storage.h
```diff
@@ -20,7 +20,7 @@ class Terrain3DStorage : public Resource {
 
 	// Constants & Definitions
 
-	static inline const float CURRENT_VERSION = 0.84;
+	static inline const real_t CURRENT_VERSION = 0.84;
```
TokisanGames, line 23: float

src/terrain_3d_storage.h
```diff
@@ -63,7 +63,7 @@ class Terrain3DStorage : public Resource {
 
 	// Storage Settings & flags
 
-	float _version = CURRENT_VERSION;
+	real_t _version = CURRENT_VERSION;
```
TokisanGames, line 66: float

TokisanGames:
You know, thinking about it I think its fine that all instances here are switched to real_t. I was thinking using set_cull_margin(real_t) is a waste of space using a double when only a float is needed. But if we're doing it from gdscript what is coming in is a double anyway. 

The only concern I have is with saved variables. Say version for instance. If it's saved as a float into Storage. Then later read by a double engine, is it seamlessly converted? If it's saved as a double, can it be opened by a float engine without issue? If version and all of the other data can be read and written by both engines then it's fine to do everything as real_t.

Also include a line somewhere like __ready() that prints `LOG(ERROR, "Float: ", sizeof(float), " Real_t: ", sizeof(real_t));` and include the result here so we can make sure they are different.

SlashScreen:
Asked AThousandShips: 

```
Me:
If a double (f64) value is serialized into a text file, and then opened using a single-precision compiled engine, what happens?
is it just truncated as a float, as if you were casting
And what about the reverse?

AThousandShips: 
Again most are stored as double, just some types rely on the precision setting
Like Vector2/3/4, Basis, Transform2D/3D, etc.
Most cases of float in the engine are f64, and serialisation isn't affected unless you're storing a type that depends
```
Shouldn't be an issue.

SlashScreen:
```
SlugFiller:
Also, serialization into a text file truncates floating point way below the limit of single. There's an issue, and I believe, also a PR about it
```

src/terrain_3d.cpp
```diff
@@ -75,7 +75,7 @@ void Terrain3D::_initialize() {
 
 void Terrain3D::__ready() {
 	_initialize();
-	LOG(WARN, "Float: ", sizeof(float), " Real_t: ", sizeof(real_t));
+	LOG(WARN, "Float: ", Variant(sizeof(float)), " Real_t: ", Variant(sizeof(real_t)));
```
TokisanGames, line 78: We don't need these in the repo, I just want you to run it once to confirm that building with precision=double does something, and I'd like you to post your results in the comments.

SlashScreen: Ah, Ok

SlashScreen:
Got it to build, as I said in discord. Let me know if I need to change anything.

godot-cpp
TokisanGames, line 1: Remove from PR. I will update godot-cpp later. Current commit tag is godot-4.1.1-stable

project/project.godot
```diff
@@ -12,7 +12,7 @@ config_version=5
 
 config/name="Terrain3D"
 run/main_scene="res://demo/Demo.tscn"
-config/features=PackedStringArray("4.1")
+config/features=PackedStringArray("4.1", "Double Precision")
```
TokisanGames, line 15: Remove from PR. Users can enable this

src/terrain_3d.cpp
```diff
@@ -75,6 +75,7 @@ void Terrain3D::_initialize() {
 
 void Terrain3D::__ready() {
 	_initialize();
+	// LOG(WARN, "Float: ", Variant(int(sizeof(float))), " Real_t: ", Variant(int(sizeof(real_t))));
```
TokisanGames, line 78: remove

src/util.cpp
```diff
@@ -58,7 +58,7 @@ Vector2 Util::get_min_max(const Ref<Image> p_image) {
 }
 
 /**
- * Returns a Image of a float heightmap normalized to RGB8 greyscale and scaled
+ * Returns a Image of a real_t heightmap normalized to RGB8 greyscale and scaled
```
TokisanGames, line 61: **Returns a scaled, RGB8 greyscale Image of a heightmap normalized to its heights**

SlashScreen:
oh dear. Will me removing godot-cpp from this PR cause issues in the main branch? The CI is freaking out right now.

src/terrain_3d_editor.cpp
```diff
@@ -191,27 +191,27 @@ void Terrain3DEditor::_operate_map(Vector3 p_global_position, float p_camera_dir
 					storage->update_heights(destf);
 
 				} else if (map_type == Terrain3DStorage::TYPE_CONTROL) {
-					float alpha_clip = (brush_alpha < 0.1f) ? 0.0f : 1.0f;
-					int index_base = int(src.r * 255.0f);
-					int index_overlay = int(src.g * 255.0f);
+					real_t alpha_clip = (brush_alpha < real_t(0.1f)) ? 0.0f : 1.0f;
```
TokisanGames, line 194: Was there any sort of warning from the compiler comparing real_t brush_alpha and 0.1f?

If not, let's remove all of the literal conversions. I now think we only need the real_t conversions for ints. For floats to doubles it is free and unnecessary, unless there are warnings. Let's remove them all from here down to the last lerp on line 233.

https://stackoverflow.com/questions/13045485/division-of-double-with-a-float

The only need I see is where it is casting dest_index

SlashScreen: There were warnings when it came to function parameters; math.lerp didn't like an argument list of mixed floats and real_ts. I can remove the ones outside of those parameters, though.

TokisanGames:
I looked closer at the godot source. Excluding modules & thirdparty code, .h/.cpp lines using float: 10212, real_t: 6091.

There are only 2 usages of float in all of the physics servers, none in navigation. 

Rendering is 99% float, needed only for locations or distance. The other uses are unnecessary.

The moviewriter is a new addition and has 3 floats, 0 real_t.

So my conclusions are:
1) it appears to be used where it's needed, so hopefully doubles actually does work.
2) Usage is all over the place and not very disciplined. When real_t is used in files, it is often used on inconsequential things and who knows if double support is completed and actually works.
3) Blanket search and replace will waste space for those using it, but unlikely to cause a problem.
4) Literals like  `5.0` are doubles, and `5.0f` are floats. Godot has `real_t x = 1.0f;` all over, so I don't think we need an explicit conversion in many cases unless the compiler produces a warning.

This link says float to double conversion on x87 fpu is free, and I've seen that other places as well.
https://stackoverflow.com/questions/13045485/division-of-double-with-a-float

So I think for consistency like the physics and navigation servers, we can use real_t everywhere. Let's wrap up this and merge it in. The shader portion can come separately.

Please make the following changes:
* Address all the individual comments I made
* Remove the changes to godot-cpp and project.godot. I will update the former soon along with a version upgrade. The godot-cpp tag is godot-4.1.1-stable
* Squash your changes somewhat or completely
* Rebase

src/terrain_3d_editor.cpp
TokisanGames: Add any casts to remove warnings. Otherwise, we don't need it for float to double.

SlashScreen:
I've addressed the individual comments. I will soon remove explicit conversions, squash (hopefully I don't screw it up this time), and rebase.

TokisanGames:
> oh dear. Will me removing godot-cpp from this PR cause issues in the main branch? The CI is freaking out right now.

I don't mean *removing* godot-cpp from your repository, aka, deleting the folder and pushing that change. I mean reverting your godot-cpp to the previous version (`git checkout godot-4.1.1-stable`) and pushing that. That will revert the change you made and reset it back to the version in the main branch.

TokisanGames:
Great job, thank you. In the future don't use `git merge`. You can do anything else like rebase, but let github do all merging.

SlashScreen:
For the record, I never did git merge, I let GitHub do all of that. And, I never found any “merge” actions when I was rebasing.

PR #242, opened 2023-11-03
Add terrain occlusion baking to the editor

### LINKED ISSUE #177: Support occlusion baking
From @SlashScreen

Looking into it, it looks like adding an occlusion mesh to the rendering server to allow the terrain to act as an occluder may be relatively simple. You can [add a mesh directly to the rendering server](https://docs.godotengine.org/en/stable/classes/class_renderingserver.html#class-renderingserver-method-occluder-create) with this function and the one below it. It could go somewhere in [these blocks](https://github.com/outobugi/Terrain3D/blob/b56cda5323e8e2d17a126e6d93375c2340195918/src/terrain_3d.cpp#L202), and just upload the static mesh. However, having it update too often could cause performance issues, since it has to build a BVH.

-------

My comments
It may not be necessary due to the nature of a clipmap terrain that already lods the distant terrain and hides everything beyond. But it's worth considering if it is worth it. Occluders could hide our enemies and cities between mountains. But manual shapes under mountains can also accomplish this.

tcoxon:
When a Terrain3D node is selected, this adds a new button to the editor for baking an occlusion mesh for the terrain.
image

When clicked, a dialog appears asking for a LOD. This value determines the granularity of the occlusion mesh, and therefore the number of vertices. Baking an occluder at a lower level of detail (higher number) will reduce opportunities for culling, but make occlusion testing quicker (and the mesh will use less memory).

Baking pauses the editor for a few seconds. It has to read every pixel on the height map once to make sure that the generated occluder doesn't extend above / outside the clipmap (at any level of detail).

Here's some draft documentation for the wiki: [occlusion-baking.md](https://github.com/TokisanGames/Terrain3D/files/13253121/occlusion-baking.md)

There's a second button this PR adds ("Bake ArrayMesh") which uses the same LOD dialog. It "comes for free" with occlusion mesh baking since it's generated the same way, but if you prefer I can strip this out of the PR. I've found this function useful for baking low-detail terrain meshes for unreachable distant islands and minimap generation, that are otherwise difficult to do cheaply with the clipmap. The mesh it creates has no material.

Fixes #177

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -6,6 +6,7 @@
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/height_map_shape3d.hpp>
 #include <godot_cpp/classes/rendering_server.hpp>
+#include <godot_cpp/classes/surface_tool.hpp>
 #include <godot_cpp/classes/time.hpp>
 #include <godot_cpp/classes/v_box_container.hpp> // for get_editor_main_screen()
 #include <godot_cpp/classes/viewport.hpp>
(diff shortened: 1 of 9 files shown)


src/terrain_3d.cpp
```diff
@@ -465,6 +466,37 @@ void Terrain3D::_update_instances() {
...
+/**
+ * Creates a vertex for a 'static' mesh (i.e. ArrayMesh/Occluder3D baking, NOT a clipmap).
+ * p_filter:
+ *  BAKE_MESH_FILTER_NEAREST: Samples the height map at the exact coordinates given.
+ *  BAKE_MESH_FILTER_MINIMUM: Samples p_step * p_step heights around the given coordinates and returns the lowest.
+ * p_x, p_z: X and Z coordinates of the vertex. Heights will be sampled around these coordinates.
+ * p_step: The width and height of the area around the X and Z coordinates to sample.
+ */
+Vector3 Terrain3D::_bake_mesh_vertex(Terrain3D::BakeMeshFilter p_filter, int32_t p_x, int32_t p_z, int32_t p_step) const {
+	Vector3 v = Vector3(p_x, 0.0, p_z);
```
TokisanGames, line 478: Add `LOG(INFO, "Calculating vertex location");` before this line

src/terrain_3d.cpp
```diff
@@ -714,6 +746,58 @@ void Terrain3D::update_aabbs() {
...
+ * Generates a static ArrayMesh for the terrain.
+ * p_lod: Determines the granularity of the generated mesh.
+ * p_filter: Controls how vertices' Y coordinates are generated from the height map.
+ *  BAKE_MESH_FILTER_NEAREST: Samples the height map in a 'nearest neighbour' fashion.
+ *  BAKE_MESH_FILTER_MINIMUM: Samples a range of heights around each vertex and returns the lowest.
+ *   This takes longer than ..._NEAREST, but can be used to create occluders, since it can guarantee the
+ *   generated mesh will not extend above or outside the clipmap at any LOD.
+ */
+Ref<Mesh> Terrain3D::bake_mesh(int p_lod, BakeMeshFilter p_filter) const {
+	Ref<Mesh> result;
```
TokisanGames, line 759: Add `LOG(INFO, "Baking mesh at lod: ", p_lod, " with filter: ", p_filter);`

src/terrain_3d.cpp
```diff
@@ -465,6 +466,37 @@ void Terrain3D::_update_instances() {
...
+ * p_x, p_z: X and Z coordinates of the vertex. Heights will be sampled around these coordinates.
+ * p_step: The width and height of the area around the X and Z coordinates to sample.
+ */
+Vector3 Terrain3D::_bake_mesh_vertex(Terrain3D::BakeMeshFilter p_filter, int32_t p_x, int32_t p_z, int32_t p_step) const {
+	Vector3 v = Vector3(p_x, 0.0, p_z);
+	real_t height = 0.0;
+	switch (p_filter) {
+		case Terrain3D::BAKE_MESH_FILTER_NEAREST: {
+			height = _storage->get_height(v);
+		} break;
```
TokisanGames, line 483: I haven't seen this styling before, but ok. I didn't realize before today that the case statements share scope.

src/terrain_3d.cpp
```diff
@@ -714,6 +746,58 @@ void Terrain3D::update_aabbs() {
...
+ * p_filter: Controls how vertices' Y coordinates are generated from the height map.
+ *  BAKE_MESH_FILTER_NEAREST: Samples the height map in a 'nearest neighbour' fashion.
+ *  BAKE_MESH_FILTER_MINIMUM: Samples a range of heights around each vertex and returns the lowest.
+ *   This takes longer than ..._NEAREST, but can be used to create occluders, since it can guarantee the
+ *   generated mesh will not extend above or outside the clipmap at any LOD.
+ */
+Ref<Mesh> Terrain3D::bake_mesh(int p_lod, BakeMeshFilter p_filter) const {
+	Ref<Mesh> result;
+	ERR_FAIL_COND_V(!_storage.is_valid(), result);
+	ERR_FAIL_COND_V(p_lod < 0, result);
```
TokisanGames, line 761: Rather than return nothing, how about we just clamp p_lod below between 0 and 4 that way they'll still get something and we get an upper cap.

src/terrain_3d.cpp
TokisanGames: `Terrain3D::BakeMeshFilter` can be `BakeMeshFilter` now.

src/terrain_3d.cpp
tcoxon: This is the standard styling the engine source code uses for switch-case statements.

TokisanGames:
Excellent, thank you.

PR #247, opened 2023-11-13
logger: Add command line switch --terain3d-debug

### LINKED ISSUE #230: Enable debugging on the command line
Debugging should be able to be turned on via command line so that it begins at startup, rather than having to compile it.

https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-get-cmdline-args

In `Terrain3D()`
grab the OS commandline and setup an option for --terrain3d-debug <N>
Levels are: 
```
#define ERROR 0
#define WARN 99 // Higher than DEBUG_MAX so doesn't impact gdscript enum
#define INFO 1
#define DEBUG 2
#define DEBUG_CONT 3
```

lw64:
Fixes #230

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -5,6 +5,7 @@
 #include <godot_cpp/classes/editor_script.hpp>
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/height_map_shape3d.hpp>
+#include <godot_cpp/classes/os.hpp>
 #include <godot_cpp/classes/rendering_server.hpp>
 #include <godot_cpp/classes/surface_tool.hpp>
 #include <godot_cpp/classes/time.hpp>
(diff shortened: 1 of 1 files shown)


TokisanGames (changes requested):
Very good. 

* Please format it with braces and indents like this:
```c++
if (value == "MESG") {
    set_debug_level(MESG);
} else if (value == "ERROR") {
     set_debug_level(ERROR);
}
```

* You can remove MESG.
* Remove space after `begins_with`
* Install clang-format and install the hooks onto your system. That will help with automatic formatting whenever you commit.

https://github.com/TokisanGames/Terrain3D/blob/main/CONTRIBUTING.md#clang-format

TokisanGames:
Did you commit and push your changes to this? It looks exactly the same. I don't see that any of the points above were addressed.

TokisanGames:
Very good, thank you.

PR #250, opened 2023-11-18
#233: Replaced PackedByteArray with PackedInt32Array

### LINKED ISSUE #233: Importing 256th region doesn't work
When importing a height map (16384x16384), an error occurred at the location shown in the figure, and the error was constantly reported when selecting terrain nodes.
jpg

rds1983:
The #233 was caused by the fact that PackedByteArray was used to store regions' indices.
The code incremented actual index by 1.
Hence the 255th region had index 256 in the _region_map. However 256 overlows 8-bit, hence the bug occured.
I replaced PackedByteArray with PackedInt32Array, which solved the issue.

Fixes #233


TokisanGames:
Excellent, thank you.

PR #254, opened 2023-11-26
editor: move baking tools into its own menu

lw64:
I thought about the UI and it made sense to put it where the baking button currently are. The panel at the left only contains editing options for the terrain, while these tools are operating on the whole at once.

Bildschirmfoto vom 2023-11-26 01-23-35


TokisanGames:
Looks good @lw64 , thanks!

Issue #264, opened 2023-12-03
Navigation: agents sometimes take indirect routes


tcoxon:
### Terrain3D version

0.9

### Godot version

4.1.3.stable.official

### System information

Godot v4.1.3.stable - Arch Linux #1 SMP PREEMPT_DYNAMIC Tue, 10 Oct 2023 21:10:21 +0000 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 970 (nvidia; 535.113.01) - AMD Ryzen 5 2600 Six-Core Processor (12 Threads)

### Is the issue reproducable in the demo?

Yes

### Issue description

Originally noted by @TokisanGames on #253. See also: https://github.com/godotengine/godot/issues/85548

Agents sometimes take indirect routes to their targets. See the video below, where an agent could walk directly to the player, but instead walks to a point near the edge of the navmesh and then back:

https://private-user-images.githubusercontent.com/632766/287239527-82e43622-61c3-4af1-b5c8-88329af08975.mp4

This bug is not limited to Terrain3D: it can be reproduced without Terrain3D at all. However it may be worth considering what we can do to help users with this, in the form of documentation, or a built-in workaround.

## Cause

When certain navmeshes enter the navigation map (which is usually when a new scene is entered that contains a NavigationRegion3D), an error is logged:

```
E 0:00:01:0091   sync: Navigation map synchronization error. Attempted to merge a navigation mesh polygon edge with another already-merged edge. This is usually caused by crossing edges, overlapping polygons, or a mismatch of the NavigationMesh / NavigationPolygon baked 'cell_size' and navigation map 'cell_size'.
  <C++ Source>   modules/navigation/nav_map.cpp:856 @ sync()
```

This is logged when the navmesh contains vertices closer than the resolution of the navigation map allows. This can cause pathfinding to take strange routes (see smix8's comment on the godot issue).

![image](https://github.com/TokisanGames/Terrain3D/assets/165720/41e30dca-20a1-4f3c-90a0-fdf3b55b6ec9)

These vertices are created by the navmesh baking in-engine, not by Terrain3D. The source geometry created by Terrain3D is much more spaced out:

![image](https://github.com/TokisanGames/Terrain3D/assets/165720/91b315c6-5617-42ef-8e4d-adda22fc38a7)

## User Workarounds

[...]

PR #266, opened 2023-12-05
Postprocess navmeshes to work around engine nav mesh baking bugs

tcoxon:
Stop-gap fix for #264 until the underlying engine issues are fixed.

### DIFF
--- a/project/addons/terrain_3d/editor/components/baker.gd
+++ b/project/addons/terrain_3d/editor/components/baker.gd
@@ -168,6 +168,8 @@ func _bake_nav_region_nav_mesh(p_nav_region: NavigationRegion3D) -> void:
 	
 	NavigationMeshGenerator.bake_from_source_geometry_data(nav_mesh, source_geometry_data)
 	
+	_postprocess_nav_mesh(nav_mesh)
+	
 	# Assign null first to force the debug display to actually update:
 	p_nav_region.set_navigation_mesh(null)
 	p_nav_region.set_navigation_mesh(nav_mesh)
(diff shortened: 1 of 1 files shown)


TokisanGames:
Thank you. I'll merge it in. Functionally this is better and addresses the broken nav mesh generation and the errors. However cosmetically, this creates holes in the nav mesh that aren't there without this post processing. They're all over the place. 

image

They are cosmetic only, as the enemy will happily traverse them.

https://github.com/TokisanGames/Terrain3D/assets/632766/da927b65-7143-40e4-a72d-9ace61eeb467

Do you want to close #264, or mark it as waiting for godot?

tcoxon:
Those holes are just where the navmesh debug polygons drop below the surface of the terrain. If it appears to be new, it's probably just because vertices have moved by up to 0.25 units.

> Do you want to close #264, or mark it as waiting for godot?

Whatever you prefer! We could close it. I'm subscribed to the Godot issue, so when there's a fix I'll get a notification and we can reopen #264 or create a new issue.

TokisanGames:
I tested with and without, and the visual holes are definitely there only after post processing. They make it difficult to determine where the nav mesh is not generated due to slope. `¯\_ (ツ)_/¯`

image

> vertices have moved by up to 0.25 units

Where does this happen?

tcoxon:
> Where does this happen?

`_postprocess_nav_mesh_round_vertices` rounds vertex coordinates to the nearest cell_size/cell_height. The default value for cell_size & cell_height is 0.25.

They're not really holes, they're just the terrain poking through the nav mesh.

PR #271, opened 2023-12-12
Noise texture sampler filter now matches the albedo and normal filter setting

jacobcoughenour:
A bit of a follow up to my previous PR: https://github.com/TokisanGames/Terrain3D/pull/265

I made it so the noise texture sampler follows the same texture filtering setting as the albedo and normal.

Here is what it looks like with the Texture Filtering option set to Nearest now
image

I thought about adding a noise texture filtering option to TerrainMaterial3D but I can't think of a use case where someone would want smooth textures and sharp noise or vice versa.


TokisanGames:
Looks good, thanks. What does it look like with the noise set to linear?

PR #272, opened 2023-12-12
Baker improvements

tcoxon:
A couple of little improvements to nav mesh / occluder / mesh baking:

- Navmesh vertex positions get rounded instead of floored, which keeps it from z-fighting with terrain and looking like it has holes. I've tested this on all the meshes I have on hand and it doesn't reintroduce any of the previous errors.
- Baking occluders and meshes reuses existing nodes, instead of creating new MeshInstance3D/OccluderInstance3D every time.

### DIFF
--- a/project/addons/terrain_3d/editor/components/baker.gd
+++ b/project/addons/terrain_3d/editor/components/baker.gd
@@ -215,7 +215,7 @@ func _postprocess_nav_mesh_round_vertices(p_nav_mesh: NavigationMesh) -> PackedV
 	
 	var vertices: PackedVector3Array = p_nav_mesh.get_vertices()
 	for i in range(vertices.size()):
-		vertices[i] = (vertices[i] / round_factor).floor() * round_factor
+		vertices[i] = (vertices[i] / round_factor).round() * round_factor
 	return vertices
 
 
--- a/project/addons/terrain_3d/editor/components/baker.gd
+++ b/project/addons/terrain_3d/editor/components/baker.gd
@@ -52,18 +52,33 @@ func _bake_mesh() -> void:
 	if !mesh:
 		push_error("Failed to bake mesh from Terrain3D")
 		return
+
 	var undo: EditorUndoRedoManager = plugin.get_undo_redo()
-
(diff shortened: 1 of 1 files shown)


TokisanGames:
I get this when baking in the navigation demo.
```
WARNING: NavigationMesh baking process will likely fail.
Source geometry is suspiciously big for the current Cell Size and Cell Height in the NavMesh Resource bake settings.
If baking does not fail, the resulting NavigationMesh will create serious pathfinding performance issues.
It is advised to increase Cell Size and/or Cell Height in the NavMesh Resource bake settings or reduce the size / scale of the source geometry.
```

With this PR merged in, I get this error on every load/run:
```
ERROR: Navigation map synchronization error. Attempted to merge a navigation mesh polygon edge with another already-merged edge. This is usually caused by crossing edges, overlapping polygons, or a mismatch of the NavigationMesh / NavigationPolygon baked 'cell_size' and navigation map 'cell_size'.
```

Reverting back to `floor` the sync error goes away.

TokisanGames:
Oh, sorry, I'm using a modified map for the video I'm making. The file is only one region larger than the repo demo. Here are the files if you want to look.

I'm using 4.1.3. Baking gives the `suspiciously big` message. I increased the cell size/height to 0.5, but that complained saying it expected 0.25. It seems to work, though there is a noticable lag when the enemy is far away and needs to find a long path.

When I load or run it gives the nav sync error.

https://we.tl/t-rikUJhMnQk

tcoxon:
The `suspiciously big` warning comes from here: https://github.com/godotengine/godot/pull/60945/commits/79511af7c94b447409d10194239069facef9c4a0

Recast (the navmesh baker) builds a grid that covers the entire navigable area. Supposedly, when that grid exceeds 30,000,000 cells, baking breaks, regardless of how little of it is navigable. Increasing the cell size reduces how big the grid needs to be (although the cell size in ProjectSettings also needs to match, which is why it complained when you tried that).

The recommendation for large maps seems to be runtime baking: https://github.com/godotengine/godot/issues/60907#issuecomment-1122465223

Issue #270, opened 2023-12-11
Emit a signal when the height map is altered


tcoxon:
### Description

I'm working on level design tools for my project. One useful function I'd like to have is the ability for props manually placed in the world to automatically move up/down when the terrain height map is edited. This would help avoid objects accidentally ending up floating, or buried underground.

What I would need from Terrain3D, to be able to implement this, would be a signal that is emitted when the height map changes, with an AABB of the area edited as a parameter. This could be on either the Terrain3D node itself, on the Terrain3DStorage. ProtonScatter integration could also benefit from this - to automatically trigger regeneration of scatter nodes.

Without the signal, the only options are to either poll the terrain periodically and update every prop in the scene frequently - slow in large scenes - or hack Terrain3D's plugin scripts and find a way to emit the signal that way.

### Are you willing to help create this feature?

Yes

TokisanGames:
It's fine to issue signals. There are several update functions that could emit them when things are regenerated.

> with an AABB of the area edited as a parameter

Everything gets regenerated on changes. We should adjust that to mark modified regions as dirty (See https://github.com/TokisanGames/Terrain3D/issues/165) so only those are allocated for undo/redo. I can see the signal issuing a list of modified regions. Issuing an AABB though would be only for your specific tool which seems like an odd design choice for us to do unless there's another way to justify its inclusion?

tcoxon:
> Everything gets regenerated on changes.

Everything gets regenerated? I probably wasn't very clear. This is what I'm doing:

[Screencast from 2023-12-11 19-01-49.webm](https://github.com/TokisanGames/Terrain3D/assets/165720/dda737b5-d857-40d5-8eae-2b5422762f9d)

My script moves manually-placed nodes up/down when the heightmap is edited. I could do this in scenes with many times more nodes if my scripts could determine when an edit occurs, and the actual positions that were edited. The Terrain3DEditor class has the information I need - it only changes pixels within [-brush_size/2, brush_size/2] of the position clicked.

The AABB would be useful to editor scripts other than mine, including the project_on_terrain script in the Terrain3D repo provided for ProtonScatter integration. If you'd rather not have it though I'm happy to just keep this in my own fork.

TokisanGames:
> Everything gets regenerated? I probably wasn't very clear. This is what I'm doing:

On every pass of the brush, `force_update_maps()` is called, which rebuilds all height regions if a height brush is used.
 
[...]

PR #274, opened 2023-12-13
Add terrain_edited signal to Terrain3DEditor

tcoxon:
As discussed in #270.

Terrain3DEditor becomes a singleton, and gains a new `terrain_edited(edited_area: AABB)` signal. The use case is integration with other level design tools and scripts - see the linked issue.

The signal is emitted whenever:
- a region is added or removed,
- a region map is altered with a brush tool,
- or any of the above are undone / redone.

In the end I decided not to include the map_type parameter in the signal, as it seems like this would not be relevant to non-brush-related tools (region add/remove, etc.).

### DIFF
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -7,6 +7,8 @@
 #include "terrain_3d_editor.h"
 #include "util.h"
 
+Terrain3DEditor *Terrain3DEditor::_singleton = nullptr;
+
 ///////////////////////////
 // Subclass Functions
 ///////////////////////////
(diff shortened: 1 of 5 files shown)


src/terrain_3d_editor.cpp
```diff
@@ -301,11 +340,22 @@ void Terrain3DEditor::_operate_map(Vector3 p_global_position, real_t p_camera_di
 				}
 
 				map->set_pixelv(map_pixel_position, dest);
+
+				if (map_type == Terrain3DStorage::TYPE_HEIGHT) {
+					edited_position.y = dest.r;
+					edited_area = edited_area.expand(edited_position);
+				}
```
TokisanGames, line 347: This can be moved up to the map_type=height section after new line 248.

src/terrain_3d_editor.cpp
```diff
@@ -342,14 +392,16 @@ void Terrain3DEditor::_setup_undo() {
 	}
 	LOG(INFO, "Setting up undo snapshot...");
 	_undo_set.clear();
-	_undo_set.resize(Terrain3DStorage::TYPE_MAX + 2);
+	_undo_set.resize(Terrain3DStorage::TYPE_MAX + 3);
```
TokisanGames, line 395: Please add `5: edited AABB` in the function comment above, L386

src/terrain_3d_editor.cpp
```diff
@@ -504,4 +571,6 @@ void Terrain3DEditor::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("stop_operation"), &Terrain3DEditor::stop_operation);
 
 	ClassDB::bind_method(D_METHOD("apply_undo", "maps"), &Terrain3DEditor::_apply_undo);
+
+	ADD_SIGNAL(MethodInfo("terrain_edited", PropertyInfo(Variant::AABB, "edited_area")));
```
TokisanGames, line 575: Please add signal documentation (when it's fired in your PR description)

In `doc/classes/Terrain3DEditor.xml`, between `methods` and `constants` paste and fill out:
```
    <signals>
        <signal name="terrain_edited">
            <param index="0" name="edited_area" type="AABB" />
            <description>
            </description>
        </signal>
    </signals>
```

TokisanGames (changes requested):
Why does this class need to be a singleton in order to fire a signal? Can't normal classes do that?

tcoxon:
> Why does this class need to be a singleton in order to fire a signal? Can't normal classes do that?

So that tool scripts and other addon scripts besides the Terrain3D addon can access it. Currently Terrain3DEditor is created and the reference held by the EditorPlugin, but other scripts can't access this.

tcoxon:
I've updated my branch to address your other comments.

TokisanGames:
Thanks!

src/terrain_3d_editor.cpp
```diff
@@ -123,6 +153,10 @@ void Terrain3DEditor::_operate_map(Vector3 p_global_position, real_t p_camera_di
 	}
 	Object::cast_to<Node>(_terrain->get_plugin()->get("ui"))->call("set_decal_rotation", rot);
 
+	AABB edited_area;
+	edited_area.position = p_global_position - Vector3(brush_size, 0.0, brush_size) / 2;
```
TokisanGames, line 157: Shouldn't 2 be float just to avoid unnecessary conversions? I'm changing this in the new PR

PR #278, opened 2023-12-17
Generate Collision Dynamically 

### LINKED ISSUE #106: Update debug collision on operations
When adding/removing regions, sculpting, undo/redo the debug collision shapes are not updated automatically.
Currently the user must turn _show_debug_collision off and on.

Currently updating destroys and rebuilds and takes .5-1s to update, so it's not good to update on every painting. It could update during the undo/redo setup, but that's still a bit slow.

What would be faster to mark dirty regions and only update those. It also needs to add and remove CollisionShapes as regions are added or removed.

### LINKED ISSUE #161: Generate Collision Dynamically
Collision for all regions currently builds when a scene runs using the physics server.

There is a debug/show_collision option that builds collision for all regions right then, in the editor and it also works in game. This creates a static body rather than the physics server. If view gizmos is on, you can see it in the editor. You also need this if you want to enable view collision shapes in game.

Collision takes up a lot of memory and is a bit slow to generate. Each region takes 145ms to generate on my system. And running 16k x 8k collision takes about 6gb, or about 48mb per region.

What I'd like is keep the two options of build all collision at runtime, build all collision in the editor, and add dynamic options that will build a much smaller collision mesh around the camera.

The code is already structured into `Terrain3D::_build_collision`, `_update_collision`, `_destroy_collision`, but it's not fully setup to be dynamic yet. So most of the work will be in the middle one. It currently builds a 1025x1025 collision mesh. 1025 because we need one more vertex to get 1024 squares.

* Provide an option to set dynamic collision mesh size, defaulting to 65x65. I would give fixed options: 33, 65, 127. (Though you could call them 32, 64, 128 as that is the number of squares across). In the set size function, you'll have to destroy_collision and build_collision, if its different and appropriate for the mode (eg don't build in editor if an in game is specified).
* In build, create the heightmapshape of the specified size.
* `Terrain3D::__process` increase the distance_to a bit, 5-10 units, up to a maximum of 50% of `clipmap_size` so it calls snap less frequently.
* `Terrain3D::snap()`, after it sets the position of the meshes, call update_collision to set the position and vertices of the heightmap shape.
* `collision/enabled` and `set_collision_enabled` can be used to toggle it on or off in realitme. If collision isn't built and it's enabled, create collision. Otherwise if collision is built, just toggle it on or off on the physics server, calling `PhysicsServer3D.body_set_shape_disabled()`. Keep the current logic of if it's disabled, don't build collision.
* In the collision properties for the inspector (`Terrain3D::bind_methods`) `debug/show_collision` can be moved in with the collision group and change to an enum for collision_mode: dynamic in game, dynamic in editor, full in game, full in editor, with the following meanings:
  * dynamic in game (new default, generate 65x65 around player in game w/ physics server)
  * dynamic in editor (generate 65x65 around player in editor & game w/ static body)
  * full in game (current default, build for all regions w/ physics server)
  * full in editor (current debug/show_collision, build for all regions in editor & game w/ a static body)

### LINKED ISSUE #341: Fatal error (Index p_index = 0 is out of bounds) when closing the game
### Terrain3D version

v0.9.1-beta

### System information

Godot v4.2.1.stable - Windows 10.0.22631 - Vulkan (Forward+) - dedicated AMD Radeon RX 7900 XT (Advanced Micro Devices, Inc.; 31.0.24019.1006) - AMD Ryzen 9 7900 12-Core Processor (24 Threads)

No

### Issue description

When closing the scene preview of a scene that contains only a Camera and Terrain3D node, the Godot console shows:

```
ERROR: FATAL: Index p_index = 0 is out of bounds (shapes.size() = 0).
   at: get_shape (servers/physics_3d/godot_collision_object_3d.h:124)
```

I have manually compiled the Terrain3D and was able to grab the stack trace in WinDbg:
```
0:000> k
 # Child-SP          RetAddr               Call Site
00 000000bb`cfffea60 00007ff6`331deb71     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x256f185
01 000000bb`cfffeac0 00007ffa`e3ace50a     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x23da2f1
02 000000bb`cfffeaf0 00007ffa`e3bc6b26     libterrain_windows_debug_x86_64!godot::internal::_call_native_mb_ret<godot::RID,godot::RID const *,__int64 *>+0x7a [F:\Projekty\GodotProjects\Terrain3D\godot-cpp\include\godot_cpp\core\engine_ptrcall.hpp @ 62] 
03 000000bb`cfffeb50 00007ffa`e394033e     libterrain_windows_debug_x86_64!godot::PhysicsServer3D::body_get_shape+0x1a6 [F:\Projekty\GodotProjects\Terrain3D\godot-cpp\gen\src\classes\physics_server3d.cpp @ 477] 
04 000000bb`cfffebe0 00007ffa`e393d733     libterrain_windows_debug_x86_64!Terrain3D::_destroy_collision+0x17e [F:\Projekty\GodotProjects\Terrain3D\src\terrain_3d.cpp @ 477] 
05 000000bb`cfffee10 00007ffa`e3946e70     libterrain_windows_debug_x86_64!Terrain3D::_clear+0x4f3 [F:\Projekty\GodotProjects\Terrain3D\src\terrain_3d.cpp @ 248] 
06 000000bb`cfffef70 00007ffa`e3938ed0     libterrain_windows_debug_x86_64!Terrain3D::_notification+0x530 [F:\Projekty\GodotProjects\Terrain3D\src\terrain_3d.cpp @ 1095] 
07 000000bb`cffff2e0 00007ff6`33f5c91f     libterrain_windows_debug_x86_64!Terrain3D::notification_bind+0x50 [F:\Projekty\GodotProjects\Terrain3D\src\terrain_3d.h @ 25] 
08 000000bb`cffff320 00007ff6`3232331d     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x315809f
09 000000bb`cffff360 00007ff6`323231c0     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x151ea9d
0a 000000bb`cffff470 00007ff6`323231c0     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x151e940
0b 000000bb`cffff580 00007ff6`32324112     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x151e940
0c 000000bb`cffff690 00007ff6`323799ea     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x151f892
0d 000000bb`cffff6f0 00007ff6`305b74c4     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x157516a
0e 000000bb`cffff750 00007ff6`34bd56c3     Godot_v4_2_1_stable_win64+0x874c4
0f 000000bb`cffffc20 00007ff6`305312ee     Godot_v4_2_1_stable_win64!ZN6embree13TaskScheduler12startThreadsEv+0x3dd0e43
10 000000bb`cffffc70 00007ff6`305313e6     Godot_v4_2_1_stable_win64+0x12ee
11 000000bb`cffffcc0 00007ffb`9f85257d     Godot_v4_2_1_stable_win64+0x13e6
12 000000bb`cffffcf0 00007ffb`a03eaa58     KERNEL32!BaseThreadInitThunk+0x1d
13 000000bb`cffffd20 00000000`00000000     ntdll!RtlUserThreadStart+0x28
```

With this and looking at Godot's `servers/physics_3d/godot_collision_object_3d.h:124`, it looks like `Terrain3D::_destroy_collision()` should check the body shape count before calling into `body_get_shape(_static_body, 0)` [here](https://github.com/TokisanGames/Terrain3D/blob/v0.9.1-beta/src/terrain_3d.cpp#L476).

To confirm, I have added extra information at [terrain_3d.cpp:475](https://github.com/TokisanGames/Terrain3D/blob/v0.9.1-beta/src/terrain_3d.cpp#L475):
`LOG(INFO, "Freeing physics body, shape count ", PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body));`

With that change, I see the logs contain the following:
`Terrain3D::_destroy_collision: Freeing physics body, shape count 0`

I have attached the log without my custom changes. I can try to grab more information if required.

### Logs

```
 bialp  D:  Programy  Godot_v4.2.1-stable_win64.exe  .\Godot_v4.2.1-stable_win64_console.exe --terrain3d-debug=DEBUG --path F:/Projekty/GodotProjects/CodeGenesis --position "1344,720" res://Game.tscn
Godot Engine v4.2.1.stable.official.b09f793f5 - https://godotengine.org
Vulkan API 1.3.277 - Forward+ - Using Vulkan Device #0: AMD - AMD Radeon RX 7900 XT

ERROR: Attempt to register extension class 'Terrain3D', which appears to be already registered.
   at: (core/extension/gdextension.cpp:386)
ERROR: Attempt to register extension class 'Terrain3DMaterial', which appears to be already registered.
   at: (core/extension/gdextension.cpp:386)
ERROR: Attempt to register extension class 'Terrain3DStorage', which appears to be already registered.
   at: (core/extension/gdextension.cpp:386)
ERROR: Attempt to register extension class 'Terrain3DTexture', which appears to be already registered.
   at: (core/extension/gdextension.cpp:386)
ERROR: Attempt to register extension class 'Terrain3DTextureList', which appears to be already register.
   at: (core/extension/gdextension.cpp:386)
ERROR: Attempt to register extension class 'Terrain3DEditor', which appears to be already registered.
   at: (core/extension/gdextension.cpp:386)
Terrain3D::_clear: Clearing the terrain
Terrain3D::_initialize: Checking material, storage, texture_list, signal, and mesh initialization
Terrain3D::_initialize: Creating blank material
Terrain3D::_initialize: Creating blank texture list
Terrain3D::_initialize: Connecting texture_list.textures_changed to _material->_update_texture_arrays()
Terrain3D::_initialize: Connecting region_size_changed signal to _material->_set_region_size()
Terrain3D::_initialize: Connecting regions_changed signal to _material->_update_regions()
Terrain3D::_initialize: Connecting height_maps_changed signal to update_aabbs()
Terrain3D::set_material: Setting material
Terrain3DMaterial::~Terrain3DMaterial: Destroying material
Terrain3D::_clear: Clearing the terrain
Terrain3D::_initialize: Checking material, storage, texture_list, signal, and mesh initialization
Terrain3D::_initialize: Connecting texture_list.textures_changed to _material->_update_texture_arrays()
Terrain3D::_initialize: Connecting region_size_changed signal to _material->_set_region_size()
Terrain3D::_initialize: Connecting regions_changed signal to _material->_update_regions()
Terrain3D::set_texture_list: Setting texture list
Terrain3D::_clear: Clearing the terrain
Terrain3D::_initialize: Checking material, storage, texture_list, signal, and mesh initialization
Terrain3D::_initialize: Connecting texture_list.textures_changed to _material->_update_texture_arrays()
Terrain3D::_notification: NOTIFICATION_ENTER_WORLD
Terrain3D::_notification: NOTIFICATION_ENTER_TREE
Terrain3D::_initialize: Checking material, storage, texture_list, signal, and mesh initialization
Terrain3DMaterial::initialize: Initializing material
Terrain3DMaterial::_preload_shaders: Loaded shader insert: uniforms
Terrain3DMaterial::_preload_shaders: Loaded shader insert: TEXTURE_SAMPLERS_LINEAR
Terrain3DMaterial::_preload_shaders: Loaded shader insert: TEXTURE_SAMPLERS_NEAREST
Terrain3DMaterial::_preload_shaders: Loaded shader insert: world_noise
Terrain3DMaterial::_preload_shaders: Loaded shader insert: WORLD_NOISE1
Terrain3DMaterial::_preload_shaders: Loaded shader insert: WORLD_NOISE2
Terrain3DMaterial::_preload_shaders: Loaded shader insert: auto_shader
Terrain3DMaterial::_preload_shaders: Loaded shader insert: AUTO_SHADER_UNIFORMS
Terrain3DMaterial::_preload_shaders: Loaded shader insert: AUTO_SHADER_TEXTURE_ID
Terrain3DMaterial::_preload_shaders: Loaded shader insert: TEXTURE_ID
Terrain3DMaterial::_preload_shaders: Loaded shader insert: dual_scaling
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DUAL_SCALING_UNIFORMS
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DUAL_SCALING_VERTEX
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DUAL_SCALING_BASE
Terrain3DMaterial::_preload_shaders: Loaded shader insert: UNI_SCALING_BASE
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DUAL_SCALING_OVERLAY
Terrain3DMaterial::_preload_shaders: Loaded shader insert: debug_views
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_CHECKERED
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_GREY
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_HEIGHTMAP
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_COLORMAP
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_ROUGHMAP
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_CONTROL_TEXTURE
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_CONTROL_BLEND
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_AUTOSHADER
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_TEXTURE_HEIGHT
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_TEXTURE_NORMAL
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_TEXTURE_ROUGHNESS
Terrain3DMaterial::_preload_shaders: Loaded shader insert: DEBUG_VERTEX_GRID
Terrain3DMaterial::_preload_shaders: Loaded shader insert: editor_functions
Terrain3DMaterial::_preload_shaders: Loaded shader insert: EDITOR_NAVIGATION
Terrain3DMaterial::_preload_shaders: Loaded shader insert: main
Terrain3DMaterial::_set_region_size: Setting region size in material: 1024
Terrain3DMaterial::initialize: Mat RID: RID(4114578669577), _shader RID: RID(4118873636876)
Terrain3DMaterial::_update_shader: Updating shader
Terrain3DMaterial::_generate_shader_code: Generating default shader code
Terrain3DMaterial::_update_shader: Material rid: RID(4114578669577), shader rid: RID(4118873636876)
Terrain3DMaterial::_update_shader: _active_params: [&"_region_size", &"_region_texel_size", &"_mesh_ver]
Terrain3DUtil::print_dict: Printing Dictionary: _shader_params
Terrain3DUtil::print_dict: Key: _mouse_layer, Value: 2147483648
Terrain3DUtil::print_dict: Key: blend_sharpness, Value: <null>
Terrain3DUtil::print_dict: Key: height_blending, Value: <null>
Terrain3DUtil::print_dict: Key: macro_variation1, Value: <null>
Terrain3DUtil::print_dict: Key: macro_variation2, Value: <null>
Terrain3DUtil::print_dict: Key: noise1_angle, Value: <null>
Terrain3DUtil::print_dict: Key: noise1_offset, Value: <null>
Terrain3DUtil::print_dict: Key: noise1_scale, Value: <null>
Terrain3DUtil::print_dict: Key: noise2_scale, Value: <null>
Terrain3DUtil::print_dict: Key: noise3_scale, Value: <null>
Terrain3DUtil::print_dict: Key: noise_texture, Value: <NoiseTexture2D#-9223372009977674545>
Terrain3DMaterial::set_mesh_vertex_spacing: Setting mesh vertex spacing in material: 1
Terrain3DMaterial::_update_regions: Updating region maps in shader
Terrain3DMaterial::_update_regions: Height map RID: RID(0)
Terrain3DMaterial::_update_regions: Control map RID: RID(0)
Terrain3DMaterial::_update_regions: Color map RID: RID(0)
Terrain3DMaterial::_update_regions: _region_map.size(): 256
Terrain3DMaterial::_update_regions: Region map
Terrain3DMaterial::_update_regions: Region_offsets size: 0 []
Terrain3DMaterial::_generate_region_blend_map: Regenerating (512, 512) region blend map
Terrain3DGeneratedTex::create: RenderingServer creating Texture2D
Terrain3DUtil::dump_gen: Generated blend_map RID: RID(4252017623075), dirty: false, image: <Image#-9223>
Terrain3DTextureList::update_list: Reconnecting texture signals
Terrain3DMaterial::_update_texture_arrays: Updating texture arrays in shader
Terrain3D::_setup_mouse_picking: Setting up mouse picker and get_intersection viewport, camera & screend
Terrain3D::set_mouse_layer: Setting mouse layer: 32 (2147483648) on terrain mesh, material, mouse camerd
Terrain3D::set_render_layers: Setting terrain render layers to: 2147483649
Terrain3DMaterial::set_shader_param: Setting shader parameter: _mouse_layer
Terrain3D::_build: Building the terrain meshes
Terrain3DGeoClipMap::generate: Generating meshes of size: 48 levels: 7
Terrain3DGeoClipMap::_create_mesh: Creating mesh via the Rendering server
Terrain3DGeoClipMap::_create_mesh: Setting custom aabb: (0, 0, 0), (49, 0.1, 49)
Terrain3DGeoClipMap::_create_mesh: Creating mesh via the Rendering server
Terrain3DGeoClipMap::_create_mesh: Setting custom aabb: (-96, 0, -96), (193, 0.1, 193)
Terrain3DGeoClipMap::_create_mesh: Creating mesh via the Rendering server
Terrain3DGeoClipMap::_create_mesh: Setting custom aabb: (-97.5, 0, -97.5), (194.5, 0.1, 194.5)
Terrain3DGeoClipMap::_create_mesh: Creating mesh via the Rendering server
Terrain3DGeoClipMap::_create_mesh: Setting custom aabb: (-97.5, 0, -97.5), (194.5, 0.1, 194.5)
Terrain3DGeoClipMap::_create_mesh: Creating mesh via the Rendering server
Terrain3DGeoClipMap::_create_mesh: Setting custom aabb: (-97.5, 0, -97.5), (291.5, 0.1, 291.5)
Terrain3D::_build: Creating mesh instances
Terrain3D::_build_collision: Building collision with physics server
Terrain3D::_update_collision: Collision creation time: 0 ms
Terrain3D::_notification: NOTIFICATION_READY
Terrain3D::_initialize: Checking material, storage, texture_list, signal, and mesh initialization
Terrain3D::__process: camera is null, getting the current one
Terrain3D::_grab_camera: Connecting to the in-game viewport camera
Terrain3D::_notification: NOTIFICATION_EXIT_TREE
Terrain3D::_clear: Clearing the terrain
Terrain3D::_destroy_collision: Freeing physics body
ERROR: FATAL: Index p_index = 0 is out of bounds (shapes.size() = 0).
   at: get_shape (servers/physics_3d/godot_collision_object_3d.h:124)
```

### LINKED ISSUE #397: Enabling collision increases memory consumption each time the scene is reloaded.
### Terrain3D version

v0.9.1-beta

### System information

Godot_v4.2.2-stable　Vulkan Forward+

Yes

### Issue description

Each time I reload a scene with terrain placed, memory consumption increases.　Disabling collisions does not cause this problem. The same problem occurred when exporting to an executable file.
https://github.com/TokisanGames/Terrain3D/assets/172058970/7bcd7477-8c79-4e1f-8a7d-46aa58bf89f9

### Logs

No errors displayed in console

lw64:
Admin edit: 
- [x] Should address #106
- [x] Test #341 
- [x] Test #397
- [x] Test Not disabling collision first when generating from code (causes  `ERROR: FATAL: Index p_index = 0 is out of bounds (shapes.size() = 0) in import_images`, disabling it beforehand makes it work.
 
- [x] Test #329 
- [x] Test #328 

Fixes #161 
Fixes #106
Fixes #341 
Fixes #397

See https://github.com/TokisanGames/Terrain3D/pull/278#issuecomment-2619955778


project/demo/Demo.tscn
```diff
@@ -87,4 +87,5 @@ texture_list = ExtResource("8_g2of2")
 render_cull_margin = 1000.0
 collision_mask = 3
 mesh_size = 64
+debug_show_collision = true
```
TokisanGames, line 90: Please leave this disabled in the demo.

src/terrain_3d.cpp
```diff
@@ -1075,6 +1139,7 @@ void Terrain3D::_bind_methods() {
 
 	ADD_GROUP("Collision", "collision_");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "collision_enabled"), "set_collision_enabled", "get_collision_enabled");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "dynamic_collision"), "set_dynamic_collision", "get_dynamic_collision");
```
TokisanGames, line 1142: This variable name broke the inspector group. It needs to begin with "collision_" to keep all of these in one group.

Instead, look at the last bullet point of #161 and make this and debug_show_collision one enum called `collision_mode` with the four listed meanings.

src/terrain_3d.cpp
```diff
@@ -293,111 +293,163 @@ void Terrain3D::_update_collision() {
...
+					shape_data["max_height"] = min_max.y;
+					PhysicsServer3D::get_singleton()->shape_set_data(shape, shape_data);
+					PhysicsServer3D::get_singleton()->body_add_shape(_static_body, shape);
+					PhysicsServer3D::get_singleton()->body_set_shape_transform(_static_body, i, xform);
+					PhysicsServer3D::get_singleton()->body_set_collision_mask(_static_body, _collision_mask);
+					PhysicsServer3D::get_singleton()->body_set_collision_layer(_static_body, _collision_layer);
+					PhysicsServer3D::get_singleton()->body_set_collision_priority(_static_body, _collision_priority);
+				} else {
+					CollisionShape3D *debug_col_shape;
+					debug_col_shape = memnew(CollisionShape3D);
```
TokisanGames, line 436: If you play this and watch your process monitor, the game continually consumes more and more memory as the player moves around. All memory allocation needs to be moved to `_build_collision` and all freeing moved to `_destroy_collision`.

src/terrain_3d.cpp
```diff
@@ -293,111 +293,163 @@ void Terrain3D::_update_collision() {
...
-		int region = _storage->get_region_index(Vector3(global_pos.x + region_size, 0, global_pos.z));
-		if (region >= 0) {
-			map_x = _storage->get_map_region(Terrain3DStorage::TYPE_HEIGHT, region);
-			cmap_x = _storage->get_map_region(Terrain3DStorage::TYPE_CONTROL, region);
-		}
-		region = _storage->get_region_index(Vector3(global_pos.x, 0, global_pos.z + region_size));
-		if (region >= 0) {
-			map_z = _storage->get_map_region(Terrain3DStorage::TYPE_HEIGHT, region);
-			cmap_z = _storage->get_map_region(Terrain3DStorage::TYPE_CONTROL, region);
+	float maximum_distance = 64.0;
```
TokisanGames, line 310: This should also be a parameter. 
* Rename collision_region_size to collision_shape_size
* Rename maximum_distance to collision_distance 
* Make both settable variables accessible to gdscript

src/terrain_3d.cpp
```diff
@@ -293,111 +293,163 @@ void Terrain3D::_update_collision() {
...
+			memfree(child);
 		}
-		region = _storage->get_region_index(Vector3(global_pos.x + region_size, 0, global_pos.z + region_size));
-		if (region >= 0) {
-			map_xz = _storage->get_map_region(Terrain3DStorage::TYPE_HEIGHT, region);
-			cmap_xz = _storage->get_map_region(Terrain3DStorage::TYPE_CONTROL, region);
+	}
+	if (_static_body.is_valid() && PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body) > 0) {
+		for (int i = PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body) - 1; i >= 0; i--) {
+		    RID shape = PhysicsServer3D::get_singleton()->body_get_shape(_static_body, i);
```
TokisanGames, line 334: Rather than query the shapes from the server, just create an array with all of the predefined collision shapes in build_collision. Then empty the array in destroy_collision.

src/terrain_3d.cpp
```diff
@@ -293,111 +293,163 @@ void Terrain3D::_update_collision() {
...
+					for (int x = 0; x < shape_size; x++) {
+						// Choose array indexing to match triangulation of heightmapshape with the mesh
+						// https://stackoverflow.com/questions/16684856/rotating-a-2d-pixel-array-by-90-degrees
+						// Normal array index rotated Y=0 - shape rotation Y=0 (xform below)
+						// int index = z * shape_size + x;
+						// Array Index Rotated Y=-90 - must rotate shape Y=+90 (xform below)
+						int index = shape_size - 1 - z + x * shape_size;
+
+						int x_in_region = x + jx * collision_region_size;
+						int z_in_region = z + jz * collision_region_size;
```
TokisanGames, line 388: Throughout you are using `region` to refer to the collision shape. It's confusing to read and understand what these variables are for. Please go through and rename all of the variables you've added that reference your collision shapes. Call them something to do with `shape`. Use `region` only for things that deal with the entire 1024x1024 region. Here x_in_region is the x shape size, plus the shape global location. So it should be called `shape_global_x`.

src/terrain_3d.cpp
```diff
@@ -293,111 +293,163 @@ void Terrain3D::_update_collision() {
...
 	int region_size = _storage->get_region_size();
-	int shape_size = region_size + 1;
+	int collision_region_size = region_size;
+	if (_dynamic_collision) {
+		if (!UtilityFunctions::is_instance_valid(_camera) || !_camera->is_inside_tree()) {
+			return;
+		}
+		collision_region_size = 16; // TODO parameter
+	}
+	int divisor = region_size / collision_region_size;
```
TokisanGames, line 303: Divisor is the number of collision shapes of size 16x16 that fit in a region of size 1024x1024, or 64. So it should be named appropriately, eg `shapes_per_region`

TokisanGames (changes requested):
Alright, good job. I think once the memory management is moved to build collision and destroy collision it will be on par or better than a single 65x65 collision shape. Worse because it has to manage 50-60 shapes, but better because it only needs to update a few smaller shapes instead of all 65x65.

We're pretty close. Make the modifications, rebase and it should be finished or very close.
Thanks!

src/terrain_3d.cpp
```diff
@@ -293,123 +293,201 @@ void Terrain3D::_update_collision() {
...
-					if (map_x.is_valid()) {
-						map_data[index] = (Util::is_hole(cmap_x->get_pixel(0, z).r)) ? hole_const : map_x->get_pixel(0, z).r;
+				PackedFloat32Array map_data;
+				CollisionShape3D *debug_col_shape = nullptr;
+				RID col_shape = RID();
+				if (_dynamic_collision && !_unused_collision_shapes.is_empty()) {
+					if (!_show_debug_collision) {
+						col_shape = _unused_collision_shapes.pop_back();
+						Dictionary shape_data = PhysicsServer3D::get_singleton()->shape_get_data(col_shape);
+						map_data = shape_data["heights"];
```
lw64, line 361: I am not entirely sure this avoids allocations. But I also don't see another method. the map_data gets assigned back later again.

src/terrain_3d.cpp
```diff
@@ -293,123 +293,201 @@ void Terrain3D::_update_collision() {
...
+					int shape_id = PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body);
+					PhysicsServer3D::get_singleton()->body_set_shape(_static_body, shape_id, col_shape);
+					PhysicsServer3D::get_singleton()->body_set_shape_transform(_static_body, shape_id, xform);
+					PhysicsServer3D::get_singleton()->body_set_collision_mask(_static_body, _collision_mask);
+					PhysicsServer3D::get_singleton()->body_set_collision_layer(_static_body, _collision_layer);
+					PhysicsServer3D::get_singleton()->body_set_collision_priority(_static_body, _collision_priority);
+				} else {
+					Ref<HeightMapShape3D> hshape;
+					if (_dynamic_collision && debug_col_shape != nullptr) {
+						hshape = debug_col_shape->get_shape();
```
lw64, line 455: I implemented the chaching for both the debug and non-debug modes, but I don't know if that actually makes sense.

src/terrain_3d.cpp
lw64: I couldn't move the memory allocation fully to `_build_collision`, as it is not defined at build time, how many collision shapes need to be allocated. the `distance` is variable, and for example for `64.0` the shape count varies between 48 and 52 in my experience, depending on the camera position.

lw64: I made it now allocate new ones on demand, but that should only happen during the first `_update_collision` call.

lw64:
One thing I observed is that when you change the collision_mode from full to dynamic, the collision shapes aren't freed. But I guess its not a problem since this isn't changed at runtime. And in the editor you can just reopen the scene

src/terrain_3d.cpp
TokisanGames: The allocations and deallocations we want to avoid in update_collision are `memnew` and `memfree`. Creating a PackedFloat32Array here is fine. This gets created on the stack which is designed for frequent, rapid creation/destruction. The other two allocate memory on the heap, which should not be as frequent.

TokisanGames:
> I couldn't move the memory allocation fully to _build_collision, as it is not defined at build time, how many collision shapes need to be allocated.

* `Array.resize()` will resize arrays dynamically.
shapes_per_region is calculated in update_collision, but doesn't need to be, it can be calculated in build_collision.

* In game modes (physics server) are broken:
```
ERROR: Index p_index = 0 is out of bounds (shapes.size() = 0).
   at: set_shape (servers/physics_3d/godot_collision_object_3d.cpp:52)
ERROR: Index p_index = 0 is out of bounds (shapes.size() = 0).
   at: set_shape_transform (servers/physics_3d/godot_collision_object_3d.cpp:63)
```

* Three related issues don't rebuild collision properly. Upon set_collision_mode, set shape size, & distance it should destroy and build collision to free allocations and set everything up properly, just like set_collision_enabled does.
  * Changing from `Full editor` to `Dynamic editor` all collision meshes are huge and don't get resized down.

  * Changing shape size (on dynamic in editor mode) crashes godot.

  * Changing from an editor mode to a game mode does not remove the shapes, but should. 

I didn't evaluate all of the new code since there are things that aren't working. Let me know if you need help with anything or want to work on things together.

I added a PR that renames and organizes some things.

src/terrain_3d.cpp
```diff
@@ -283,148 +293,232 @@ void Terrain3D::_update_collision() {
...
+						shape_data["min_height"] = min_max.x;
+						shape_data["max_height"] = min_max.y;
+						PhysicsServer3D::get_singleton()->shape_set_data(col_shape, shape_data);
 					}
-				} else if (x == region_size && z == region_size) {
-					if (map_xz.is_valid()) {
-						map_data[index] = (Util::is_hole(cmap_xz->get_pixel(0, 0).r)) ? hole_const : map_xz->get_pixel(0, 0).r;
+					int shape_id = PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body);
+					PhysicsServer3D::get_singleton()->body_add_shape(_static_body, col_shape);
+					PhysicsServer3D::get_singleton()->body_set_shape_transform(_static_body, shape_id, xform);
```
lw64, line 458: The problem here is the shape id. There is no way to get the correct one added with `add()`, and the on I guessed using `shape_count` seems to be wrong.

TokisanGames:
In dynamic editor mode, collision shapes do not have unique positions and are doubling up

image


Though this solution is an interesting method, I feel like this is an overly complicated solution. 

What I proposed was using a small 65x65 collision square around the player, then on every snap, update the vertices of the shape, so 4225 vertex updates every snap, (which is based on movement, not frames). That is how my [gterrain code](https://github.com/TokisanGames/gterrain/blob/570951f8fa16a94ea186df8ca3791fee5c984375/addons/gterrain/src/GTerrain.gd#L161) worked, and is very short and simple. It will work fine with the static body and physics server, and does not continually allocate and deallocate memory.

This solution has improves upon it by breaking down the collision shape into 17x17 vertex chunks, and only new chunks are updated. So when moving laterally, say 8 chunks are updated, or 2312 vertices updated.

The cost of it though is a rather complicated chunking system that needs to track 50-64 collision shapes and move them around accurately without making memory non-contiguous by constant allocations and deallocations. This hasn't been achieved quite yet, so additional cost is that we need to spend more time to preallocate all of the shapes in build_collision, and identify that shapes aren't doubling up. 

Then we have to maintain this code over time.

And the physics server needs to calculate based on 64 shapes instead of 1. 

Though I started working on this PR to clean up and fix things, once I discovered the doubling up, I think the scales tipped for me in the cost/benefit analysis. 

I'm not sure the added overhead of us and the physics server managing all of the extra shapes, and the effort we've put in and still need to put in is worth the savings of ~2000 vertex assignments per snap. Whether it's 2000 or 4000 adjustments, both are inconsequential.

I think we should revert most of these changes and go back to the simpler solution proposed in #161 and demonstrated in my [gterrain code](https://github.com/TokisanGames/gterrain/blob/570951f8fa16a94ea186df8ca3791fee5c984375/addons/gterrain/src/GTerrain.gd#L161). Though the editor collision modes and settable collision shape size are good.

What do you think?

lw64:
So I addressed your issue with the doubled collision shapes and its fixed now. Also I finished the non-editor collision, which was previously not really done yet. The issues I had before with missing API was resolved by finding the correct API xD I tested this a bit and for me it doesn't leak any memory and there shouldn't be any allocations as far as I can see. So I think its good to go now.

src/terrain_3d.cpp
lw64: Which other two? `Dictionary`?

TokisanGames: By other two, I mean the general usage of memnew and memfree. Not referring to lines specifically.

TokisanGames:
Alright, I tested more and now it's all working well. Good job. There are a few more improvements that I made and some more I've left for you.

**Items I adjusted in my commits:**

* If Dynamic shape size is odd, the vertices don't line up with the terrain. If the size is not a power of 2, there are gaps at the region boundaries. If the size is too small relative to distance, say 2 and 64, 8 and 256 the whole engine slows down (would be fixed by redesign at bottom). The numbers can get negative or very large. I enforced sane values so they won't break

* Renamed variables and added comments
* Replaced append w/ push_back. They are aliases but the latter is used everywhere else
* Replaced pop_at(i) on negative forloops with pop_back, which is faster. The former recreates the array.
* Stripped camera detection out of update_collision
* Enforced collision setup in build_collision and removed from update_collision
* Changed update calculation time to usec and print it on DEBUG_CONT error level

**There are two other issues I found and have not fixed.**

---------

### Invalid RID
I get this error message multiple times on DYNAMIC_GAME, after running the game, moving the camera, then quitting.

```
ERROR: Invalid ID.
   at: (servers/physics_3d/godot_physics_server_3d.cpp:1619)

or using Jolt:

ERROR: Failed to free RID: The specified RID has no owner.
   at: (src\servers\jolt_physics_server_3d.cpp:1782)
```

This is caused by this code in destroy_collision, 
```
		for (int i = PhysicsServer3D::get_singleton()->body_get_shape_count(_static_body) - 1; i >= 0; i--) {
			RID shape = PhysicsServer3D::get_singleton()->body_get_shape(_static_body, i);
			PhysicsServer3D::get_singleton()->free_rid(shape);
		}
```

This means we can't rely on the server to house the RIDs but must track them ourselves, as we do with `_static_body` and other RID variables. This is typical and expected. The solution is addressed by the redesign below.

-------

### Inefficient Loops

At even the default values for dynamic size/distance with 50-64 shapes in existing, using `existing.has()` to check if the shape is there within 3 for loops means checking: 
* 3 vector integers for each of 64 shapes
* shapes per region squared - default 64
* regions - up to 256

So up to 3 * 64 * 64 * 256 = 3.1 million integer comparisons just to tell if the collision shape already exists. Both the checking if it exists and the overall for loop need to be better designed. We don't need to check for all regions, for all shapes per region x & z, and we don't need to iterate through the array using has.

What should happen in a chunking system is:

**_build_collision**
  * number of shapes needed identified
  * all shapes are created, memory allocated
  * all shapes tracked in three arrays:
    * All shapes - fixed size identified from above
    * Used shapes - fixed size - identified by calculation of distance_size as a grid around camera, some indices are always empty
    * Unused shapes - dynamic size used with push_back / pop_back

**_update_collision**
  * We assume there is a fixed grid where shape locations can be (like an integer grid, or every 4 units, etc)
  * Rather than evaluating all possible points on the grid for all regions, we just take a small area around the camera to evaluate
  * Rather than checking the entire existing array each time, we just translate the grid location to an array index. Then we go directly to that array index and if something is already there, we skip it. This is like how the region offsets work, see `storage.get_region_offset()` and `get_region_index()`. They don't iterate over all regions, they just translate position into an array index. Our _region_map is a fixed size 16x16 array with many empty slots. This isn't the exact code we need, but we need to use the same concept. 
  * Anything unused is disabled and stored in the unused array
  * The only thing that happens in this function is moving shapes between arrays, moving shape locations, and setting vertices. All in a small area around the provided update location. No allocations.

**_destroy_collision**
  * All shapes in our arrays are destroyed without asking the physics server or the tree (children) for them. Fixes the invalid RID problem since we track and free all of our own RIDs, which is our responsibility.

This logic is the same for physics server shapes or instanced shapes. There's very little code difference between the two. The main difference is that we get the RID of the shape and send it to the server, or how shapes are created and destroyed.

This should reduce the update time a bit, but even if not, it's a cleaner design that will help us maintain the code in the future. It won't take much modification from what's already there. But those crucial for loops need to be reworked. I can help you implement this if you like.

src/terrain_3d.cpp
```diff
@@ -244,187 +248,363 @@ void Terrain3D::_build(int p_mesh_lods, int p_mesh_size) {
...
+						if (shape_global_x < 0) {
+							shape_global_x = region_size + shape_global_x;
+						}
+						int shape_global_z = z + (global_shape_location.z % region_size);
+						if (shape_global_z < 0) {
+							shape_global_z = region_size + shape_global_z;
+						}
+						// Set heights on local map, or adjacent maps if on the last row/col
+						if (shape_global_x < region_size && shape_global_z < region_size) {
+							map_data[index] = (Util::is_hole(cmap->get_pixel(shape_global_x, shape_global_z).r)) ? hole_const : map->get_pixel(shape_global_x, shape_global_z).r;
```
TokisanGames, line 486: As soon as I switched to Dynamic / Editor running the demo crashes on this line. The camera is outside of the terrain. The reason is:
`get_pixel` is called with `(896, 464)`, however cmap, cmap_x, cmap_xz, cmap_z are all null

Region is currently -1, which means no region. So if region == -1 after line 427, you can skip all of the code down to L558 probably. Though you may want to disable that shape.

TokisanGames: Adding this in L428 is sufficient to have it work. However I haven't looked yet to see if this properly disables the shapes or if it needs to do more.
```
				if (region < 0) {
					continue;
				}

```

TokisanGames:
I'm going to look at the design a bit more later, but here's some initial testing of functionality.

* It was working very well before, just not the ideal design. Now as it's being redesigned, the issues are normal. Just make sure to not squash your PR yet so we can refer back to commit d720d55f86efeacaf8901e08176e1f0ad2659622 for reference.

* I recently realized the use of `memfree()` in `_destroy_collision()` is wrong. It should be `memdelete()` with `memnew`. `memfree` goes with `memalloc`. Changing these dramatically reduces the number of `RID Allocations leaked at exit` messages when godot closes, which it should. We weren't removing any of them. 

* In Dynamic/editor mode, it seems to be generating only 1/4th of the circle so it doesn't work in game since it's off center.
image

* Full / editor mode only generates for the first region.


* On a couple of editor startups I got this error message. Other times it didn't happen. I didn't troubleshoot, so look out for it.
 
```
ERROR: Terrain3D::_update_collision: Collision not initialized, returning
   at: push_error (core/variant/variant_utility.cpp:1091)
```

TokisanGames:
Note, I fixed the memory leaks for the up coming release.

https://github.com/TokisanGames/Terrain3D/commit/c5115fec2140f5ecff53a0fd11ab6b913e7b28c1

Saul2022:
An error i found checking this pr is that  when trying to make play the demo, it doesn´t start and leaves the long messages
 that update coll is not initialized

 core\variant\variant_utility.cpp:1091 - Terrain3D::_update_collision: Collision not initialized, returning

editor_screenshot_2024-02-19T112621

TokisanGames:
My current thinking for the foliage instancer is to generate MultiMeshes in chunks. I will build it off of the chunk infrastructure you're building here. I'll be ready to work on foliage in a week or two, so if this isn't done by then I'll get involved to move it along.

TokisanGames:
This PR needs address https://github.com/TokisanGames/Terrain3D/issues/341#issuecomment-2022062079, and make sure it works without error when running and quitting without regions.

TokisanGames:
Rebased and consolidated down to three commits. 1st is the previously identified working commit. It works fine but has some errors on cleanup. 2nd is a first pass rewrite that also seems to work, with errors out of bounds. 3rd is a separation into separate classes, which is currently broken due to changes in the main tree not considered in this branch. Unknown if it works.

Next step is to audit the new design, consolidate the new classes down to 1 or 2, and review and adjust the code.

lw64:
@TokisanGames yes, thanks for the rebase! Please give some feedback on the system design and APIs

TokisanGames:
@lw64 I'll take over this PR and work on it for my next priority. As you requested, I'll give you feedback along the way.

Some notes I have already:

* The number of classes and level of abstraction is good for a project that's going to scale very big, or an academic project. For our needs, pairing that down to 1 or 2 classes will do. The instancer handles its own chunk system all in one function.

* Like Godot code, all the headers will have the same copyright. You are included in Contributors and noted on the front page of the repo and in Authors.

Next I need to understand what changes are made in commits 2 and 3, and what is the best way to structure it.

src/chunk_manager.cpp
```diff
@@ -0,0 +1,132 @@
...
+		p_size--;
+		p_size |= p_size >> 1;
+		p_size |= p_size >> 2;
+		p_size |= p_size >> 4;
+		p_size |= p_size >> 8;
+		p_size |= p_size >> 16;
+		p_size++;
+	}
+	p_size = CLAMP(p_size, 8, 256);
+	_chunk_size = p_size;
```
TokisanGames, line 130: This is duplicate code and if it's not already available in the engine, should be put in Terrain3DUtil

TokisanGames:
We have 5 classes: 3 manager types, 2 chunk types. 

Each derivative class should represent a unique object we will create. So if we were going have chunks for say instances, instance collision, ground collision that might warrant 3 child classes and 1 base class for chunks.

However, one manager could process all of those classes. It would be designed to work with the base chunk so it could handle all child types. I don't see the reasoning for having 3 manager types, and certainly not for having more manager types than managee types (chunks).

The design pattern that would be a better fit for this might be a mixture of the composite and template patterns. Something like this, which is exactly how our ProximityManager is constructed. In this case the manager might move the baseobject, then pass the transform or other data to the object run process which could handle generating ground collision, or generating an MMI, or generating collision for objects in the area. 

```
class Manager {
    std::vector<BaseObject*> objects;

public:
    void registerObject(BaseObject* object) {
        objects.push_back(object);
    }

    void processAll() {
        for (auto o : objects) {
            o->process();
        }
    }
};

class BaseObject {
public:
    virtual void process() = 0;
    virtual ~BaseObject() = default;  // Virtual destructor for proper cleanup
};

class Box : public BaseObject {
public:
    void process() override {
        std::cout << "I'm a box." << std::endl;
    }
};

class Sphere : public BaseObject {
public:
    void process() override {
        std::cout << "I'm a sphere." << std::endl;
    }
};
```

lw64:
> However, one manager could process all of those classes. It would be designed to work with the base chunk so it could handle all child types. I don't see the reasoning for having 3 manager types, and certainly not for having more manager types than managee types (chunks).

I went for the most simple approach for the start, which is each type of chunk has its own very simple manager, so simple even that they could be in the same file, and the main logic is in the base manager class.

But I can see a lot of potential for optimization there, and having only one manager instance handling everything can enable quite a lot. Also something to keep in mind is doing more complex chunking for example with quadtrees later.

But I think the most complicated part about having one single manager instance is that each chunk type might have different properties, gras chunks are smaller and have shorter visibility, while tree chunks should be bigger and have very far visibility. I think this is possible though, and could avoid a lot of overhead, since not every manager instance does the same work again and again.

lw64:
@TokisanGames and thanks for taking over! :D

TokisanGames:
The overall hierarchical structure of the later commits isn't bad. I just think it is not a good match for our use case. 

base_chunk would be more useful if we were to have multiple class types derive from it. mesh_chunk, collision_shape_chunk, instancer_chunk, etc, each with their own unique functionality, but also enough shared code like positioning, signal responding, creation, deletion (a lot that is already handled by Node3D).

chunk_manager would be useful if we were to have multiple types of managers with their own unique code, but also a lot of shared code.

I think these many classes might be used more if we were making a chunked terrain. In our case a lot of work is already done by engine classes or the clipmap nature.

In a chunked terrain, a chunk might contain a group of nodes with several mesh instance LODs, a physics body, and a collision shape. It may also have a subscene with the assets at that location. It's a lot of unique functionality not already present in the engine.

What functionality do we need in our chunks?

In the instancer, we store the transforms in a data structure, then we create MMIs and place them on the terrain in a grid. The MMIs are the objects we instance that perform a special function (rendering), hence they are the chunk. We don't need another any other functionality in the chunk beyond what the MMI and it's inherited Node3D and Object classes do, so we don't need another class to contain it.

For collision, the object we want to instance and move around is a collision shape. The only functionality we need is to store the data and move it. You have the chunk molding itself to the heightmap. That functionality can easily be done by the manager. I think either is a fine design choice, but in this case, since forming the shape is the only thing it might do, we might as well have the manager do it so we can make the chunk slimmer and remove the containing class entirely.

So we can get away with an instancer management class that manages its MMIs, and a single collision manager class that manages its collision shapes. Both managers creates, moves, configures, and destroys their chunks, but the amount of code to handle those isn't similar enough that it's worth abstracting into a super class. And the chunks already share abstract code in Node3D and Object, so I don't think we need to wrap them any further. 

This is the direction I'm going to move in. Consolidating the classes into a single manager with the collision shape as a chunk and using the first commit as a guide, since it works very well.

TokisanGames:
Update: Dynamic / Editor and Dynamic / Game modes are working with the new design. Remaining to do are the static modes.

TokisanGames:
Size/range have been more constrained and dynamic modes now work with vertex spacing 0.25-100. Editor nodes have been removed from the scenetree.

TokisanGames:
This PR is ready for testing and a careful code review. I'll work on documentation next while waiting for reviews.

To test:

* All options in the collision group, game modes, collision layers/mask/priority
* Vertex spacing
* All demos
* Editor tests in both editor collision modes
  * Adding/removing regions - should rebuild in full, update in dynamic
  * Sculpting - full ignored, update in dynamic/editor
  * Undo/redo - should rebuild/update
* The bugs found in the tracked tickets in the top post

Download this artifact to test: https://github.com/TokisanGames/Terrain3D/actions/runs/13018553729


Notes:

* collision.update() can be called in any mode to update collision, but will skip dynamic slots in the correct spot. Adding (true) will force all collision to be updated at runtime

* Updating dynamic collision as fast as 60fps while sculpting or making holes with gizmos on costs < 0.1ms CPU time (at 16s/64m). So realtime destruction is feasible now. In fact, I have enabled updating from Terrain3DEditor while sculpting in dynamic / editor mode. I don't know if this is worth keeping, but it's practically free, so maybe.

* All Dynamic shape size/distance combos work. The sliders auto adjust their sizes so non-working ratios are not settable.

* The Edge region fix was reimplemented in 034eae9b929cf59dbe2a0dfa8ec27fc9e300c8b0

src/terrain_3d_collision.cpp
```diff
@@ -0,0 +1,550 @@
...
+	ClassDB::bind_method(D_METHOD("get_layer"), &Terrain3DCollision::get_layer);
+	ClassDB::bind_method(D_METHOD("set_mask", "mask"), &Terrain3DCollision::set_mask);
+	ClassDB::bind_method(D_METHOD("get_mask"), &Terrain3DCollision::get_mask);
+	ClassDB::bind_method(D_METHOD("set_priority", "priority"), &Terrain3DCollision::set_priority);
+	ClassDB::bind_method(D_METHOD("get_priority"), &Terrain3DCollision::get_priority);
+	ClassDB::bind_method(D_METHOD("get_rid"), &Terrain3DCollision::get_rid);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "mode", PROPERTY_HINT_ENUM, "Disabled,Dynamic / Game,Dynamic / Editor,Full / Game,Full / Editor"), "set_mode", "get_mode");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "shape_size", PROPERTY_HINT_RANGE, "8,64,8"), "set_shape_size", "get_shape_size");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "radius", PROPERTY_HINT_RANGE, "16,128,16"), "set_radius", "get_radius");
```
Xtarsia, line 546: Should be Variant::INT ?

Xtarsia (commented):
Tested both Godot and Jolt physics, all modes.

The only error i could find was when using Dynamic/Editor mode, and occasionally getting this:

image

I was unable to get this to occur in dynamic/game, or other modes.

I think 128m might be a bit on the small side, since when testing earlier 256m with larger tile sizes worked fine, maybe allow the or-greater flag (at the users own risk)?

TokisanGames:
> The only error i could find was when using Dynamic/Editor mode, and occasionally getting this:
> `core/templates/sort_array.h:255 - bad comparison function; sorting will be broken`

That line is here:
```
	inline void unguarded_linear_insert(int64_t p_last, T p_value, T *p_array) const {
		int64_t next = p_last - 1;
		while (compare(p_value, p_array[next])) {
			if (Validate) {
				ERR_BAD_COMPARE(next == 0); // <----
			}
```

Which might be called from linear_insert(), unguarded_linear_sort(), or linear_sort(), which are all in class SortArray. This class is used by: 
* Navigation math a-star
* Triangle mesh math
* Scene Tree
* Viewport
* Rendering
* Physics collision solver, shape3d

It's not directly used by Dictionary or Array. It could be something totally unrelated like rendering the gizmos or handling the tree children. I have seen this error before, but I'm not able to reproduce it on demand, even with size/range abuse in game and not. Do you recall if these errors occurred in the editor or in game w/ view collision shapes on or off? Or what size/range you were using?

TokisanGames:
> I think 128m might be a bit on the small side, since when testing earlier 256m with larger tile sizes worked fine, maybe allow the or-greater flag (at the users own risk)?

I expanded range to 256 and if >128, bump min shape_size to 16.

TokisanGames:
> I think 128m might be a bit on the small side, since when testing earlier 256m with larger tile sizes worked fine, maybe allow the or-greater flag (at the users own risk)?

I expanded range to 256 and if >128, bump min shape_size to 16.

TokisanGames:
@lw64 Thanks for your help on this. Finally merged.

JekSun97:
I don't quite understand what this does, but from what I understand, it will generate collision around the camera/player, but other creatures in the distance can fall through the terrain?

TokisanGames:
Other creatures in the distance probably shouldn't be processing and wasting cycles. But if they must, read the docs:
https://terrain3d.readthedocs.io/en/latest/docs/collision.html#enemy-collision

jordanmichaelrushing:
@TokisanGames what about multiple local players and multiple computer players. Thinking specifically a racing game where there is couch versus with computer players. Is this setup to handle dynamic multi humans and computer players?

I tried setting terrain camera for each player but it seems like it only is targeting the last player that set it. Obviously that's a problem if one player is on one side of the race track and humans/computers are on the other side of the race track. Should I just switch fully to raycasting for humans and computers?

TokisanGames:
You should use discord or Discussions for unrelated questions. By posting here you've emailed everyone who's ever subscribed on this ticket.

Terrain3D and all clipmap terrains aren't great at split screen. The terrain can only be in one place (the clipmap target) per frame. If the player cameras are spread far, the non-active one will display lower lods and won't match collision. You can increase mesh size to help.

As for calculating physics, you need to either increase the dynamic collision size large enough to cover all players, or use full collision mode, or get_height(), or GPU based raycasting (see get_intersection() code and create your own). Physics raycasting won't solve the issue.

Follow up on discord for gamedev advice unrelated to this development PR.

PR #295, opened 2024-01-06
Moves responsibility of checking texture compatibility to developer.

lfxu:
Redoing #294. When setting a "New ImageTexture", a valid image of FORMAT_MAX is created. Now logs an error "Expects an actual texture file. See documentation for format suggestions."

### DIFF
--- a/src/terrain_3d_texture.cpp
+++ b/src/terrain_3d_texture.cpp
@@ -20,8 +20,8 @@ bool Terrain3DTexture::_is_texture_valid(const Ref<Texture2D> &p_texture) const 
 	if (img.is_valid()) {
 		format = img->get_format();
 	}
-	if (format != Image::FORMAT_DXT5) {
-		LOG(ERROR, "Invalid format. Expected channel packed DXT5 RGBA8. See documentation for format.");
+	if (format < 0 || format >= Image::FORMAT_MAX) {
+		LOG(ERROR, "Expects an actual texture file. See documentation for format specification.");
 		return false;
 	}
 
--- a/src/terrain_3d_texture.cpp
+++ b/src/terrain_3d_texture.cpp
@@ -21,7 +21,7 @@ bool Terrain3DTexture::_is_texture_valid(const Ref<Texture2D> &p_texture) const 
 		format = img->get_format();
 	}
 	if (format < 0 || format >= Image::FORMAT_MAX) {
-		LOG(ERROR, "Expects an actual texture file. See documentation for format specification.");
+		LOG(ERROR, "Invalid texture format. See documentation for format specification.");
 		return false;
 	}
 
--- a/src/terrain_3d_texture_list.cpp
+++ b/src/terrain_3d_texture_list.cpp
@@ -56,6 +56,9 @@ void Terrain3DTextureList::_update_texture_data(bool p_textures, bool p_settings
 		Vector2i albedo_size = Vector2i(0, 0);
 		Vector2i normal_size = Vector2i(0, 0);
 
+		Image::Format alb_format = Image::FORMAT_MAX;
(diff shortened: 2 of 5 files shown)


src/terrain_3d_texture.cpp
```diff
@@ -20,8 +20,8 @@ bool Terrain3DTexture::_is_texture_valid(const Ref<Texture2D> &p_texture) const
 	if (img.is_valid()) {
 		format = img->get_format();
 	}
-	if (format != Image::FORMAT_DXT5) {
-		LOG(ERROR, "Invalid format. Expected channel packed DXT5 RGBA8. See documentation for format.");
+	if (format < 0 || format >= Image::FORMAT_MAX) {
+		LOG(ERROR, "Expects an actual texture file. See documentation for format suggestions.");
```
TokisanGames, line 24: Please change `suggestions` to `specification`.

src/terrain_3d_texture_list.cpp
```diff
@@ -72,6 +75,12 @@ void Terrain3DTextureList::_update_texture_data(bool p_textures, bool p_settings
 				} else {
 					ERR_FAIL_COND_MSG(tex_size != albedo_size, "Albedo textures do not have same size!");
 				}
+				Image::Format format = alb_tex->get_image()->get_format();
+				if (alb_format == Image::FORMAT_MAX) {
+					alb_format = format;
+				} else {
+					ERR_FAIL_COND_MSG(alb_format != format, "Albedo textures do not have same format! See documentation for format suggestions");
```
TokisanGames, line 82: Please change `suggestions` to `specification.` with period, and repeat below.

TokisanGames:
Thanks for putting this in another branch. This fixes the crashing problem. This issue remains:

* When inserting a non-DXT5/RGBA8 image, then adding a subsequent texture, the texturing breaks when the new texture slot is created because the system generates a texture with DXT5/RGBA w/ mipmaps. The new texture already matches the previous texture size. It also needs to match the format.

TokisanGames (changes requested):
Thanks for putting this in another branch. This fixes the crashing problem. This issue remains:

* When inserting a non-DXT5/RGBA8 image, then adding a subsequent texture, the texturing breaks when the new texture slot is created because the system generates a texture with DXT5/RGBA w/ mipmaps. The new texture already matches the previous texture size. It also needs to match the format.

lfxu:
I did some investigation. The actual cause is:
1. Clicking the plus sign to add a texture triggers `_update_texture_data()` without triggering `_is_texture_valid()`.
2. `alb_tex.is_valid()` is false in `_update_texture_data()`, but only on the first call. If you add another texture the second texture `alb_tex.is_valid()` is true. This shows the checkerboard texture is created after `_update_texture_data()` is called.

lfxu:
Thanks for clarifying. By "the system generates" I thought Godot did that.

Looks like there is not a straightforward way to convert an image to a specified format. One will have to deal with `CompressMode` etc. I also found https://github.com/godotengine/godot/issues/79932. At this point, it is starting to feel impractical / not worth it at this point. Creating a blank texture seems feasible enough though.

TokisanGames:
This only occurs in the editor, so that ticket is not a concern.
If the first texture exists, you can get its format then create an image with that same format.
The only challenging part is if its a compressed format. You could have a list of compressed formats, and if it's in that list, you can use the compression function. It should work fine for 80-90% of texture types. 

If they have a rare compressed texture that godot can't create, then they'll get an error and a broken display until they drop in a new texture, which is acceptable since it will happen in a small subset of rare textures people will only use <1% of the time.

TokisanGames:
Updated the PR after a lot more testing. Thanks!

Issue #121, opened 2023-07-17
Cursor flickers, is unstable and inaccurate up close


TokisanGames:
The cursor is still so clunky when doing close up, detailed work. The decal disappears often or flickers as the mouse position isn't accurately detected. If it stays on the terrain, it only moves vertex by vertex.

The recent updates are not enough:
38d015735706c0fb0d1634028a91f6f175af7806

One option is to generate physics in the local area and use intersect_ray. Also we can review zylann's code for what he did.

See #88

TokisanGames:
* In get_intersection(), read the map directly rather than use get_height, which can be slow for many iterations. Maybe 3000 isn't enough to matter.

* I think the algorithm can be improved by using ray marching instead of the current ray intersection method.

TokisanGames:
There's a problem exacerbated by vertex spacing #296  in that the mouse cursor jumps over the 3000 unit limit before it transitions to the other tracking method.

https://github.com/TokisanGames/Terrain3D/blob/b95f7f81646ccb98f9611bef9ce7da62d3057c41/project/addons/terrain_3d/editor/editor.gd#L126

https://github.com/TokisanGames/Terrain3D/assets/632766/5a80ae42-1338-4dbb-8b86-2f429f8a1626

TokisanGames:
I have implemented a working prototype of the GPU detection method in #313. When finalized it will close this issue.

PR #296, opened 2024-01-06
Implements vertex spacing (mesh density)

### LINKED ISSUE #131: Adjustable mesh density
Our mesh vertices are every 1 unit. The Witcher 3 uses 0.5 for their clipmap terrain, which I'd like to copy. Lower end hardware might benefit from 2+.

This would allow using a 1024px map for a 512m world, or a 2048m world instead of the current 1024px = 1024m world.

### LINKED ISSUE #137: Android Texture Artifacts
m4gr3d ran terrain3d on android (samsung galaxy tab) and it works fine, except the shader looks like it has terrible floating point precision.

https://chat.godotengine.org/channel/android?msg=Xo95Qcs52bv3BAF2X

image

https://github.com/TokisanGames/Terrain3D/assets/914968/06828f51-1da4-4df1-a0e4-265e81f3143c

https://github.com/TokisanGames/Terrain3D/assets/914968/02d5b916-00ec-47f1-b7f2-90604dc002ad

@calinou suggested 
> I guess you need a highp annotation in some places

lfxu:
Admin edit:
Fixes #131
Fixes #137 

------

Implements #131.

Opted for naming "vertex spacing" because it has the least ambiguity.

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -23,6 +23,8 @@ render_mode blend_mix,depth_draw_opaque,cull_back,diffuse_burley,specular_schlic
 // Private uniforms
 uniform float _region_size = 1024.0;
 uniform float _region_texel_size = 0.0009765625; // = 1/1024
+uniform float _mesh_vertex_spacing = 1.0;
+uniform float _mesh_vertex_density = 1.0; // = 1/_mesh_vertex_spacing
 uniform int _region_map_size = 16;
 uniform int _region_map[256];
 uniform vec2 _region_offsets[256];
@@ -60,10 +62,12 @@ struct Material {
 	float blend;
 };
 
-varying vec3 v_vertex;	// World coordinate vertex location
-varying vec3 v_camera_pos;
-varying float v_vertex_dist;
+varying flat vec3 v_vertex;	// World coordinate vertex location
+varying flat vec3 v_camera_pos;
+varying flat float v_vertex_dist;
 varying flat ivec3 v_region;
+varying flat vec2 v_uv_offset;
+varying flat vec2 v_uv2_offset;
 
 ////////////////////////
 // Vertex
@@ -110,7 +114,7 @@ void vertex() {
 	v_vertex = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
 	
 	// UV coordinates in world space. Values are 0 to _region_size within regions
-	UV = v_vertex.xz;
+	UV = round(v_vertex.xz * _mesh_vertex_density);
 
 	// Discard vertices if designated as a hole or background disabled. 1 lookup.
 	v_region = get_region_uv(UV);
@@ -128,6 +132,12 @@ void vertex() {
 		v_vertex_dist = length(v_vertex - v_camera_pos);
 //INSERT: DUAL_SCALING_VERTEX
 	}
+
+	// Transform UVs to local to avoid poor precision during varying interpolation.
+	v_uv_offset = MODEL_MATRIX[3].xz * _mesh_vertex_density;
+	UV -= v_uv_offset;
+	v_uv2_offset = v_uv_offset * _region_texel_size;
+	UV2 -= v_uv2_offset;
 }
 
 ////////////////////////
@@ -140,8 +150,8 @@ vec3 get_normal(vec2 uv, out vec3 tangent, out vec3 binormal) {
 	float right = get_height(uv + vec2(_region_texel_size, 0));
 	float back = get_height(uv + vec2(0, -_region_texel_size));
 	float front = get_height(uv + vec2(0, _region_texel_size));
-	vec3 horizontal = vec3(2.0, right - left, 0.0);
-	vec3 vertical = vec3(0.0, back - front, 2.0);
+	vec3 horizontal = vec3(2.0 * _mesh_vertex_spacing, right - left, 0.0);
+	vec3 vertical = vec3(0.0, back - front, 2.0 * _mesh_vertex_spacing);
 	vec3 normal = normalize(cross(vertical, horizontal));
 	normal.z *= -1.0;
 	tangent = cross(normal, vec3(0, 0, 1));
@@ -248,16 +258,20 @@ float blend_weights(float weight, float detail) {
 }
 
 void fragment() {
+	// Recover UVs
+	vec2 uv = UV + v_uv_offset;
+	vec2 uv2 = UV2 + v_uv2_offset;
+
 	// Calculate Terrain Normals. 4 lookups
 	vec3 w_tangent, w_binormal;
-	vec3 w_normal = get_normal(UV2, w_tangent, w_binormal);
+	vec3 w_normal = get_normal(uv2, w_tangent, w_binormal);
 	NORMAL = mat3(VIEW_MATRIX) * w_normal;
 	TANGENT = mat3(VIEW_MATRIX) * w_tangent;
 	BINORMAL = mat3(VIEW_MATRIX) * w_binormal;
 
 	// Idenfity 4 vertices surrounding this pixel
-	vec2 texel_pos = UV;
-	highp vec2 texel_pos_floor = floor(UV);
+	vec2 texel_pos = uv;
+	highp vec2 texel_pos_floor = floor(uv);
 	
 	// Create a cross hatch grid of alternating 0/1 horizontal and vertical stripes 1 unit wide in XY 
 	vec4 mirror = vec4(fract(texel_pos_floor * 0.5) * 2.0, 1.0, 1.0);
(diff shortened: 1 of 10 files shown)


TokisanGames:
If this doesn't scale the overall terrain size, then it means 1px != 1 vertex, and we're losing detail. Also I don't see it changing the collision shape at all.

Turn on debug collision so you can see it. It needs to be pixel perfect.

Saul2022:
Where is the option at? i built it from source, with your branch and it doesn´t seem to appear neither in the mesh or the renderer section.

lfxu:
Rendering and editor are now scaled. Navigation and collision todo.

lfxu:
Navigation and collision should now work too. Let me know if anything is still missing.

lfxu:
> Where is the option at? i built it from source, with your branch and it doesn´t seem to appear neither in the mesh or the renderer section.

It's in the mesh section, under LOD and size. Make sure you replace the right files in your project and restart Godot.

Saul2022:
Tested ( now after  finally being able to make it build after fail and error)  and it works as expected, the terrain  m scales by it double when changing from for instance, 1 to 2, even with multiple regions it seem´s to work well.

src/terrain_3d.cpp
```diff
@@ -1077,6 +1096,7 @@ void Terrain3D::_bind_methods() {
 	ADD_GROUP("Mesh", "mesh_");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "mesh_lods", PROPERTY_HINT_RANGE, "1,10,1"), "set_mesh_lods", "get_mesh_lods");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "mesh_size", PROPERTY_HINT_RANGE, "8,64,1"), "set_mesh_size", "get_mesh_size");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "mesh_vertex_spacing", PROPERTY_HINT_RANGE, "0.5,50,0.5"), "set_mesh_vertex_spacing", "get_mesh_vertex_spacing");
```
TokisanGames, line 1099: Please change the range to `"0.25,10.0,0.05,or_greater"`

src/terrain_3d.h
```diff
@@ -35,6 +35,7 @@ class Terrain3D : public Node3D {
 	// Terrain settings
 	int _mesh_size = 48;
 	int _mesh_lods = 7;
+	float _mesh_vertex_spacing = 1;
```
TokisanGames, line 38: Change all uses of `float` to `real_t`, except where it must be a 32-bit float (control map) and the shaders. Even where it doesn't matter like this small value. And format literals like `1.0f` for assignment, elsewhere they can just be `1.0`. All literals should have a decimal so they are visually apparent.

src/terrain_3d.cpp
```diff
@@ -604,6 +607,18 @@ void Terrain3D::set_mesh_size(int p_size) {
 	}
 }
 
+void Terrain3D::set_mesh_vertex_spacing(float p_spacing) {
```
TokisanGames, line 610: clamp it to sane values like 0 and 100.

TokisanGames (changes requested):
Thank you for this. I tested it and it works well with collision, navigation, and occlusion. Good job so far! This is a highly requested feature. Here are some issues I found.

* The cursor decal disappears at some angles and heights, and the current tool doesn't work or is offset from the decal at different angles and locations. I tested and found these issues with both 0.5 and 2.0. Use the height brush on the demo and you'll see it disappear above a certain height. Rotate the camera around and add new regions to see it.

* Please explain in detail the changes you made to the shader.

* The normals are off. The first image here has a directional light facing directly down. Here the mesh has a scale of 1.

image

Here is a different section where the mesh is scaled at 10. I sculpted it to have a similar height and slope as the other. With the light straight down it should look like the above.

image

lfxu (commented):
Issues should be fixed now. Please take a look.

TokisanGames:
* These new commits change the behavior dramatically. In main, we currently have 1px == 1 vertex and 1 vertex == 1m.

Before these commits, it scaled the terrain laterally, changing only the latter ratio: 1v != 1m, and keeping the first equation, which was my expectation.

After these commits, now both equations are variable. 1px != 1 vertex and 1 vertex != 1m. The problem with this is we have a 1024x1024 height data, but with a vertex spacing of 2. So, we are ignoring 75% of the height data and wasting memory. We need to maintain 1px == 1 vertex.

So this does "fix" the normals issue, but only by not doing any scaling.


* When I build the demo and adjust vertex spacing, it changes the region gizmo size, but they no longer line up with the sculpted area. At 0.25, the region gizmos indicate a very small region space, but the actual regions with sculpted data haven't moved.
image

src/shaders/main.glsl
```diff
@@ -61,9 +63,11 @@ struct Material {
 };
 
 varying vec3 v_vertex;	// World coordinate vertex location
-varying vec3 v_camera_pos;
+varying flat vec3 v_camera_pos;
```
TokisanGames, line 66: Why is this flat?

src/terrain_3d_material.h
```diff
@@ -63,6 +63,7 @@ class Terrain3DMaterial : public Resource {
 	// Cached data from Storage
 	int _texture_count = 0;
 	int _region_size = 1024;
+	real_t _mesh_vertex_spacing = 1;
```
TokisanGames, line 66: 1.0f

src/terrain_3d_material.cpp
```diff
@@ -508,6 +508,13 @@ Variant Terrain3DMaterial::get_shader_param(const StringName &p_name) const {
 	return value;
 }
 
+void Terrain3DMaterial::set_mesh_vertex_spacing(real_t p_spacing) {
+	LOG(INFO, "Setting mesh vertex spacing in material: ", p_spacing);
+	_mesh_vertex_spacing = p_spacing;
+	RS->material_set_param(_material, "_mesh_vertex_spacing", p_spacing);
+	RS->material_set_param(_material, "_mesh_vertex_density", (real_t)1.0f / p_spacing);
```
TokisanGames, line 515: I don't think (real_t) is needed here. if p_spacing is a double the literal will be promoted to a double for free. Float or double will match the appropriate Variant constructor, which gets converted back to a float anyway.

TokisanGames:
Thanks for fixing the highlighted issues. Please address the individual inline notes I mentioned, as well the following and I'll merge it in:

* Fix region grid and vertex grid in debug_views.glsl, and any other usages of UV/UV2.
* Restore v_vertex height
* Rebase and resolve conflicts

src/shaders/main.glsl
lfxu: Why is it not? Does view space to world space transform matrix change per vertex?

src/terrain_3d_material.h
lfxu: Done

src/terrain_3d_material.cpp
lfxu: Done

src/shaders/main.glsl
```diff
@@ -124,10 +128,15 @@ void vertex() {
 
 		// Get final vertex location and save it
 		VERTEX.y = get_height(UV2);
-		v_vertex = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
```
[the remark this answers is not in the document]
lfxu: Done

TokisanGames:
Ok, found a few more errors.

* Decal wasn't scaling. I fixed it.
* New slope sculpting doesn't pick the correct heights. At vertex spacing 1, use the slope sculpting brush, pick a peak, pick a plane, it makes a ramp. Repeat and the ramp gets stronger. 
  * At larger spacing, the higher peak decreases in height each iteration. It should stay the same. 
  * At smaller spacing, the higher end increases in height.
  * The flatten brush height picker seems fine.
* Detecting the mouse cursor has an odd hiccup at some camera angles, when scaling is above 1. Set to 2 or more, use the color map to paint red, pick a flat area and start making vertical streaks and you'll see it jump.


https://github.com/TokisanGames/Terrain3D/assets/632766/5a80ae42-1338-4dbb-8b86-2f429f8a1626

src/shaders/main.glsl
TokisanGames: I wasn't really familiar with it. I see it disables interpolation, not truncates to integer, which is what it sounds like. So why not also make v_vertex and v_vertex_dist flat? All of the varyings we're using currently are there to provide fixed vertex data.

lfxu:
These don't seem introduced by this PR.
1. The slope tool always decreases a little bit. This is on `main`.
2. Drawing has a gap at distance, in a round shape, due to a branch in `editor.gd` https://github.com/Toki
sanGames/Terrain3D/blob/b95f7f81646ccb98f9611bef9ce7da62d3057c41/project/addons/terrain_3d/editor/editor.gd#L126.

See the reproductions on `main`.
[draw.webm](https://github.com/TokisanGames/Terrain3D/assets/16812130/8c35bf45-7c99-478a-b2ae-94acc0d6c8a1)
[slope.webm](https://github.com/TokisanGames/Terrain3D/assets/16812130/32973c34-df48-4f94-8d19-7a7ca1fd7d94)

These issues should be addressed elsewhere.

A real issue though is the slope decals are at wrong locations. I fixed them.

TokisanGames:
The slope sculpting was still broken due to mouse_global_position.y also being scaled when it shouldn't have been. That's now fixed.

I also made all of the varyings flat.

The mouse cursor jump can be addressed in #121.

This is ready to go. Great work. Thank you!

lfxu:
Thanks!

Issue #96, opened 2023-07-01
Implement Slope Sculpting Tool 


TokisanGames:
This issue is for sculpting a slope. This should be a tool. For painting textures by slope, see #282

Brush Slope does nothing and was hidden in 283767e1162e558f4f32bb1037014a08507406e8, merged in #91.

Implement it.

tcoxon:
It's looking like I'll need this tool for my project. Do you have a particular idea in mind for how it would work? If our needs align, I could consider implementing & contributing it.

Something like the flatten tool, where you pick a starting height, then click and drag, with the difference being the further you drag from your starting point, the lower the height gets, according to an angle setting?

TokisanGames:
I think there are two implementations of `slope`. I'll take PRs for either or both.

1) Slope when applying textures and color, painting only on flat surfaces or only on slopes.

2) Slope height brush: Select starting point, Select ending point, and it automatically creates a grade between them in the general shape and size of the current brush.

Starting/ending elevations can be predefined with the other brushes. 

I've found when brushing heights that I need a slope to connect between a higher area and a lower area. I don't think I've ever needed an arbitrary length brush. Have you found differently? Where you want to start making a grade without a destination in mind?

tcoxon:
> Have you found differently? Where you want to start making a grade without a destination in mind?

We're only just in the process of initially blocking out terrain. In our case, the destination doesn't matter too much because almost all space is unused. And it can be useful to guarantee a particular angle to the slope to ensure navigability. But I _can_ see that if you've already got a lot of stuff on your terrain that you want to keep, it makes sense to need a particular end point.

Do you see slope height brush as being:
1. Drag from starting point to ending point. Then when you release the mouse button, it modifies the terrain with a slope between them in a straight line from start to end.
2. Or click starting point, click ending point, then click and drag out a potentially wavy line between the two, modifying terrain as the mouse moves?

TokisanGames:
What I envision is a simple: Click start, click end, straight slope is created. No drag. 

However, if you want to do # 2, click start, click end, then it becomes drawable with height auto adjusting depending on proximity to start or ending points, also using opacity. That's an interesting option. 

Why don't you do both? Have a drawable checkbox. If it's not checked we get the instant straight slope. If checked it is drawable.

TokisanGames:
Try the height brush with low opacity and you'll see it fills in more gradually than when it is 100%.

niceandgoodonline:
[...]

PR #297, opened 2024-01-09
Implement slope sculpting tool

tcoxon:
This PR implements the slope sculpting feature discussed in #96.

[Screencast from 2024-01-09 17-18-00-edit.webm](https://github.com/TokisanGames/Terrain3D/assets/165720/17aeb754-4a04-4d16-a0d4-a7744690a56d)

Quick word on naming first: I've used both 'Gradient' and 'Slope' here, but to mean different things. A slope is a gradient applied to the height map, whereas a gradient operation could theoretically apply to other maps like color and roughness (if implemented).

UI is shown in the above video. Click two points, then if drawable is unchecked there's a little pause as it builds a straight-line slope. If drawable is unchecked you can click and drag to make wiggly slopes. If you selected the wrong point, you can use the dropper buttons in the tool settings bar to pick a different one.

Figuring out how to implement the non-drawable gradient was pretty difficult because the flow is significantly different from other tools that are already present. I've created an OperationBuilder abstraction to avoid turning `editor.gd`, `ui.gd`, and `toolbar_settings.gd` into spaghetti. Gradient operations are the only ones that use this currently.

The gradient operation in `terrain_3d_editor.cpp` uses the direction of movement of the brush as well as its position. This to keep winding slopes from having one edge higher than another. The following screenshots demonstrate the difference this makes:

Screenshot from 2024-01-09 18-34-16

Screenshot from 2024-01-09 18-32-14


Saul2022:
tested and it works well, though it freezing when painting a slop when drawable   is off. Specifically after you pinned the first pin , when you try to position the second one the engine freezes, which doesn't happen when drawable is on.

TokisanGames (changes requested):
Thanks for this PR. Excellent work. 

Some notes:

* The insta-slope is obviously slow. 10-50x slower than just using the drawable brush, so I looked into why. In `gradient_operation_builder.gd:apply_operation` the movement `dir` of 1 is unnecessarily small, especially on large brushes. I found that the movement can be 1/3rd the brush size for optimal speed without gaps. So a 60m brush can move 20 each iteration. And a 9m brush, 3m brush, 1. 

  This makes the tool much more usable. It's more than 10x faster. It makes long slopes across many regions possible. And now opacity makes more of a difference. At 100% it has a similar appearance as when movement is 1, but at 10%, it has only a subtle effect as it should.

* If 2 points are selected with drawable on, then drawable is turned off, the 2 points are still selected. On the next click it draws another slope between the two points. I think it should clear the points when drawable is turned off.

* Pressing Escape, switching to another tool, or deselecting Terrain3D should cancel picking. Currently:
  * Escape deselects Terrain3D (as it does for all nodes) and whenever Terrain3D is reselected, it's still picking
  * If you pick, then click another tool, say flatten, the picker decal is still there. If you click the terrain, the pick happens and flattening doesn't until another click.

tcoxon:
Thanks for the great feedback. I've implemented your suggestions.

TokisanGames:
Excellent, thank you. I increased the frequency from .333 to .2 for now.

PR #304, opened 2024-01-22
Make sure occlusion meshes contain holes where there are holes in terrain

### LINKED ISSUE #301: Occluder generation misses some holes
### Description

@tcoxon The occluder generation at lod4 misses the hole on the left in the demo. When the player runs through the tunnel, he disappears for a moment when the capsule is on one side of the occluder and the camera is on the other.

image

tcoxon:
Fixes #301.

Previously, the occlusion mesh baker would only leave a hole where one or more of the vertices of a triangle was directly over a hole. This meant if there was a hole that fit inside a triangle, or just didn't touch any vertices, the baker would fill the hole. This is possible because occlusion meshes are generally baked at a lower LOD.

With the changes in this PR, the occlusion baker checks every nearby position it samples the heights of for holes, and creates a hole if any of them have holes.

This doesn't change the ArrayMesh baker. The left hole in the example shown in #301 still gets filled by the ArrayMesh baker, but this is (IMO) an acceptable loss of detail that doesn't cause correctness issues.


TokisanGames:
Very good, thank you. Fixed the issue in the demo.

PR #305, opened 2024-01-22
Move terrain_edited signal to Terrain3DStorage.maps_edited

tcoxon:
As discussed in #270. I've done this so that Terrain3DEditor can be made not a singleton.


TokisanGames:
Thank you. Changed /2 to /2.0f, rebased and fixed a conflict w/ slope sculpt in start_operation

Issue #145, opened 2023-07-21
Add detiling option to the shader


TokisanGames:
The current rotation used for detiling is very unattractive for strong patterns like rock, and it is questionable for weak patterns like grass/mud. One user mentioned [rotation messes up normals](https://discord.com/channels/691957978680786944/1130291534802202735/1130577162563502150).

I think we should differentiate between detiling and rotation and make these changes:

* (This Issue) Include a better detiling algorithm, such as [one of these from IQ](https://iquilezles.org/articles/texturerepetition/), or one from [hterrain](https://github.com/Zylann/godot_heightmap_plugin/blob/a481edb415d29eb01f7b0f220bde301e9052dcfb/addons/zylann.hterrain/shaders/multisplat16.gdshader#L181). Either are superior to the current implementation.

* (Separate issue) Include paintable rotation in our control map so we can manually paint rotated textures like cobblestone paths and orient them properly. 

See proposed control map https://github.com/outobugi/Terrain3D/discussions/64#discussioncomment-6448308
Related to #97
This should also address #69

outobugi:
We probably can change the de-tiling but it will be a lot slower than our current one. Our method utilizes the index blending and the de-tiling is basically free. Most de-tiling methods use noisy patterns and ours is a grid. We would need to check if neighboring indices are same, skip them and do the de-tiling. See updated #64

Xtarsia:
![image](https://github.com/TokisanGames/Terrain3D/assets/69606701/b41da4a3-7aaf-4e4c-a2c4-71c1e2d4313b)

https://github.com/TokisanGames/Terrain3D/assets/69606701/9574e60e-09ca-4ac4-bd9e-fd203318f62e

Along side Per Texture, UV Scale & UV Rotation;
 a UV Rotation Blend value would be very handy. (some textures need hardly any at all, making de-tiling almost free for those cases)

Currently as implemented the performance hit to enable de-tiling is between 5-15% increase in sample reads.
The limit if setting blend to absolute maxium is 100% (double) but anything past 20% is kind of redundant.

and note 1 optimisation as im typing this: can disable when blend value is at maxium)

(some minor optimisations / refactoring to fit with the rest of the base shader aside)

JekSun97:
[...]

PR #317, opened 2024-02-05
Rework anti-tiling

### LINKED ISSUE #201: Rotation inconsistent in/out side of regions
Rotation doesn't look bad outside of regions. However it looks like garbage within regions. In these pics, there's one region in the top left. Figure out the difference, make it consistent, and make it look like it does outside of regions.

I suspect it is applied twice within regions.

Rock texture

image

Checkered texture

image

outobugi:
This implements a new anti-tiling method. The UV is moved around a point (UV+0.5). The result is jittered/dithered and less chaotic than the old broken rotation. Obviously no anti-tiling is made for every texture. 

I removed '_noise3_' and used the textures alpha to blend the domain transitions. This is because the noise blending looked messy when blending two or more of the same texture together.

I also removed normal map rotation as it is redundant because nothing is rotated anymore.

Closes(?) #145
Fixes #201

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -38,7 +38,6 @@ uniform vec4 _texture_color_array[32];
 uniform int _background_mode = 1;  // NONE = 0, FLAT = 1, NOISE = 2
 
 // Public uniforms
-
 uniform bool height_blending = true;
 uniform float blend_sharpness : hint_range(0, 1) = 0.87;
 //INSERT: AUTO_SHADER_UNIFORMS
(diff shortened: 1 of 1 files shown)


TokisanGames:
This is an improvement over the old uv_rotation. And it may help some users with a cheap, psuedo detiling on some textures, so it can close #201.

However, it's not an adequate detiling option so #145 can stay open and be renamed.

v0.8.2 vs this PR
https://github.com/TokisanGames/Terrain3D/assets/632766/e08fbe3d-ad68-40db-8c53-8593ec25cbd7

> I removed 'noise3' and used the textures alpha to blend the domain transitions. This is because the noise blending looked messy when blending two or more of the same texture together.

I'm conflicted about this. noise3 is used to add noise to the blending edge of textures. I see that it does strengthen the normals in the textures that are rotated. However removing the noise makes the blending pattern between textures much blocker and far more difficult to blend naturally.

Blending between textures is far more important to me than the rotation feature.

However this does consider the height on the blend more so than before, so some textures like rock through grass do blend better. But other textures very easily turn into squares and are much more difficult to blend without these squares.

Without noise3

image

With noise3

image

Perhaps we can keep noise3 and average it into the height with a user settable factor.

TokisanGames:
Here is the blend with this PR and no noise3

image

With noise3

image

Here is both averaged

image


Here is both added. So far, I think this is the best option. Painting is good and the height of rocks is improved over the main branch. And the rotated textures still has most of its height.

image


I think added might be the best option.

TokisanGames:
This is an improvement, so I will add noise3 to the height (unless you want to) and merge in this PR today or tomorrow.

TokisanGames:
Thanks!

PR #323, opened 2024-02-13
Fix mesh vertex spacing

TokisanGames:
#296 Mesh scaling did not address Storage or Editor. Only the mouse position was scaled. This means get_height and other storage functions didn't work.

This PR adds on to the work done by @lfxu to ensure the API is global. That is all function calls (eg get_height) operate in absolute global positions. If looking at the height of a mountain peak with `get_height(512, 512)`, then you set vertex spacing to 2.0, you'll need to query that mountain peak with `get_height(1024, 1024)`.

### Working
* Add/remove regions
* Storage.get_height
* Storage.get_normal
* Collision
* Brushing
* Brush decal size
* Picker decals
* Importing - works fine using scaled global positions
* Navigation doesn't generate. With spacing=2, even a very small area produces no nav mesh.
* Baking array mesh and Occlusion culling off
* Slope sculpting, heights are a little off
* Edited AABB - The idea is internally it is in descaled coordinates. What it receives and gives externally are absolute global coordinates.

### DIFF
--- a/doc/api/class_terrain3d.rst
+++ b/doc/api/class_terrain3d.rst
@@ -315,6 +315,12 @@ The correlated size of the terrain meshes. Lod0 has ``4*mesh_size + 2`` quads pe
 
 The distance between vertices. Godot units are typically considered to be meters. This scales the terrain on X and Z axes.
 
+This variable changes the global position of landscape features. A mountain peak might be at (512, 512), but with a vertex spacing of 2.0 it is now located at (1024, 1024).
+
+All Terrain3D functions with a global_position expect an absolute global value. If you would normally use :ref:`Terrain3DStorage.import_images<class_Terrain3DStorage_method_import_images>` to import an image in the region at (-1024, -1024), with a mesh_vertex_spacing of 2, you'll need to import that image at (-2048, -2048) to place it in the same region.
+
+To scale heights, export the height map and reimport it with a new height scale.
+
 .. rst-class:: classref-item-separator
 
 ----
--- a/doc/api/class_terrain3dstorage.rst
+++ b/doc/api/class_terrain3dstorage.rst
@@ -417,7 +417,7 @@ And :ref:`get_region_offset<class_Terrain3DStorage_method_get_region_offset>` wh
 - void **set_region_size** **(** :ref:`RegionSize<enum_Terrain3DStorage_RegionSize>` value **)**
 - :ref:`RegionSize<enum_Terrain3DStorage_RegionSize>` **get_region_size** **(** **)**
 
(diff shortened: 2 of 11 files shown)


tcoxon:
[This commit](https://github.com/tcoxon/Terrain3D/commits/fix-vertex-spacing-baker/) fixes some of the issues with ArrayMesh, Occluder3D and NavigationMesh baking. Basically, a few places scale the wrong thing, giving the output the appearance of having been scaled twice.

Unfortunately there still seems to be some kind of issue specifically with navmesh baking that I don't have time to look into at the moment. It could be another navmesh-related engine issue, because baking other kinds of meshes shows no issues here.
image

I've only briefly looked at the effect of vertex spacing on slope sculpting. It appears to be "spiky" whenever vertex spacing is below or above 1.

TokisanGames:
Thank you, I have added the fix for the arraymesh/occlusion/navigation generation. I also simplified the scaling code.

I generated a section and it seems fine. Note that with 3x mesh scaling, the total size of the navigation mesh increases 3x. There is a finite limit on the size of what can be generated, so I painted a much smaller navigable section and it generated fine.

 I also fixed the last row of collision.

TokisanGames:
I fixed the dithered brush and slope sculpting heights. The only remaining things are: AABB, and reviewing the code.

TokisanGames:
@tcoxon Edited aabb is wrong in the main tree. Flattening the demo mountain peak w/ vertex scaling:
* @ 1 Terrain3DStorage::add_edited_area: [P: (513.7451, 336.2854, 613.7875), S: (50, 26.30222, 50)]
* @ 2 Terrain3DStorage::add_edited_area: [P: (521.157, 339.005, 623.036), S: (50, 26.07883, 50)]
* @ 0.5 Terrain3DStorage::add_edited_area: [P: (509.7927, 335.5546, 604.967), S: (50, 26.65845, 50)]

This pr:
* @ 1 Terrain3DStorage::add_edited_area: [P: (511.0797, 333.2357, 614.1357), S: (50, 31.5379, 50)]
* @ 2 Terrain3DStorage::add_edited_area: [P: (1058.114, 355.972, 1258.456), S: (50, 9.005615, 50)]
* @ 0.5 Terrain3DStorage::add_edited_area: [P: (241.0488, 284.7164, 292.0633), S: (50, 80.21317, 50)]

That scaling looks right to me. I think this PR is now done.

TokisanGames:
@tcoxon I'm going to merge this in so we can get more people using and testing. If there's a specific problem with edited_area we can follow up in another issue. Hopefully before I release 0.9.1, but I want to release it soon. Maybe today.

PR #325, opened 2024-02-19
Maps edited signal

tcoxon:
First commit fixes the scale of the AABB parameter on the storage `maps_edited` signal. Due to a mistake it was getting scaled by mesh_vertex_spacing twice.

Second commit adds the `terrain_prop_group.gd` script which makes use of this signal, and may also be of utility to users of Terrain3D. This script only runs in the editor. Its purpose is to move its children vertically up/down whenever they're moved or the terrain is edited, to try to preserve their vertical heights above the surface of the terrain. If a child is directly touching the terrain for instance, it will always snap to the surface of the terrain. This is very useful when modifying the heightmap of terrain in a scene you've already manually added props to!

To set up TerrainPropGroup, add it to the scene like this, with some models as children:

image

Then the nodes will 'snap' whenever moved, and automatically adjust up/down with the height of the terrain:

[Screencast from 2024-02-19 18-12-09.webm](https://github.com/TokisanGames/Terrain3D/assets/165720/a7e72b47-a337-4223-b699-8358ac573e75)


project/addons/terrain_3d/utils/transform_changed_signaller.gd
```diff
@@ -0,0 +1,18 @@
+@tool
+extends Node3D
+
+signal transform_changed
+
+
+func _ready() -> void:
+	assert(Engine.is_editor_hint())
+	if not Engine.is_editor_hint():
```
TokisanGames, line 9: Aren't these two redundant?

project/addons/terrain_3d/utils/terrain_prop_group.gd
```diff
@@ -0,0 +1,180 @@
...
+	if not (p_node is Node3D) or p_node.has_node(CHILD_HELPER_PATH):
+		return
+	
+	assert(p_node.get_parent() == self)
+	
+	var helper: TransformChangedSignaller = TransformChangedSignaller.new()
+	helper.name = CHILD_HELPER_NAME
+	p_node.add_child(helper, true, INTERNAL_MODE_BACK)
+	helper.transform_changed.connect(_on_child_transform_changed.bind(p_node))
+	assert(p_node.has_node(CHILD_HELPER_PATH))
```
TokisanGames, line 72: What will the user do with this and the above asserts on failure? How about pushing an error with direction and aborting?

project/addons/terrain_3d/utils/terrain_prop_group.gd
```diff
@@ -0,0 +1,180 @@
+@tool
+extends Node3D
+class_name TerrainPropGroup
```
TokisanGames, line 3: How about `Terrain3DProps` or `Terrain3DObjects`? Terrain3D is the brand, and group is maybe redundant and it's not a Godot `group(tm)`

project/addons/terrain_3d/editor/editor.gd
```diff
@@ -78,6 +86,11 @@ func _edit(p_object: Object) -> void:
 	else:
 		nav_region = null
 	
+	if p_object is TerrainPropGroup:
+		p_object.editor_setup(self)
+	elif p_object is Node3D and p_object.get_parent() is TerrainPropGroup:
+		p_object.get_parent().editor_setup(self)
```
TokisanGames, line 92: The check here and in _handles won't allow for nested objects. Imagine a node with a grove of trees around.
* A structure like: Node/Tree won't move anything.
* A structure like: Node3D/Tree won't the tree if the ground under it is sculpted. But the Node3D will move itself and all children if the ground under the Node3D position is changed.

Though the intention might be the user will make one TerrainPropGroup for each type they want: rocks, trees, etc, I think it's reasonable to assume some will make one TPG and make node groups underneath and will expect it to work properly. And I think the expectation is that the absolute position of the individual items would change.

So I think this should walk up the tree until a node is a TerrainPropGroup, or the edited_scene_root. (find_parent only works by name).

project/addons/terrain_3d/utils/terrain_prop_group.gd
```diff
@@ -0,0 +1,180 @@
...
+	_editor_interface = p_plugin.get_editor_interface()
+	_undo_redo = p_plugin.get_undo_redo()
+	
+	for child in get_children():
+		_on_child_entered_tree(child)
+	
+	child_entered_tree.connect(_on_child_entered_tree)
+	child_exiting_tree.connect(_on_child_exiting_tree)
+	
+	set_process(true)
```
TokisanGames, line 45: _process isn't used, so what is this and in _ready for?

project/addons/terrain_3d/utils/terrain_prop_group.gd
```diff
@@ -0,0 +1,180 @@
...
+
+func _on_child_transform_changed(p_node: Node3D) -> void:
+	if _ignore_transform_change:
+		return
+	
+	var lmb_down := Input.is_mouse_button_pressed(MOUSE_BUTTON_LEFT)
+	if lmb_down and (_is_node_selected(p_node) or _is_node_selected(self)):
+		# The user may be moving the node using gizmos.
+		# We should wait until they're done before updating otherwise gizmos + this node conflict.
+		return
```
TokisanGames, line 104: This function could support live tracking of heights with some modification. This lmb_down could be limited to saving the undo/redo, while checking heights in realtime.

TokisanGames:
Thanks for this. It looks very useful.

* `transform_changed_signaller` -> `transform_changed_notifier` is consistent with Godot's use of `notify` and `notifier` in the rendering server, `VisibleOnScreenNotifier3D`, and various functions and variables. 
  * Search help for signaller vs notif / notifier.
  * There are some signaling (note one L) names for use with WebRTC, but that's an entirely different use for the word `signal`.

### Nodes

* Moving a child node out of the Terrain3D adjacent TerrainPropGroup produced this error:
`Node not found: "../Terrain3D" (relative to "/root/@EditorNode@17141/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@62/@VBoxContainer@63/@PanelContainer@110/MainScreen/@CanvasItemEditor@9462/@VSplitContainer@9281/@HSplitContainer@9283/@HSplitContainer@9285/@Control@9286/@SubViewportContainer@9287/@SubViewport@9288/Demo/TerrainPropGroup/MeshInstance3D5").`

### Positions

* Changing the position of an object in the inspector does not update the height, only with the gizmos.


* Moving a tracked object (that has been moved before) out of defined regions, might result in a nan position and eventually crash. Or,
  * If an object hasn't moved yet, it may allow you to move it out fine. Then if you move it back in to defined regions, the object no longer tracks the terrain.


* If a tracked object is on a region, then that region is removed, or
  * If a tracked object is on a void region and that region is added,
  * the position of that object is nan, the editor slows, and the console starts dumping errors:
`ERROR: Condition "!v.is_finite()" is true.
   at: instance_set_transform (servers/rendering/renderer_scene_cull.cpp:922)`

TokisanGames (changes requested):
Thanks for this. It looks very useful.

* `transform_changed_signaller` -> `transform_changed_notifier` is consistent with Godot's use of `notify` and `notifier` in the rendering server, `VisibleOnScreenNotifier3D`, and various functions and variables. 
  * Search help for signaller vs notif / notifier.
  * There are some signaling (note one L) names for use with WebRTC, but that's an entirely different use for the word `signal`.

### Nodes

* Moving a child node out of the Terrain3D adjacent TerrainPropGroup produced this error:
`Node not found: "../Terrain3D" (relative to "/root/@EditorNode@17141/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@62/@VBoxContainer@63/@PanelContainer@110/MainScreen/@CanvasItemEditor@9462/@VSplitContainer@9281/@HSplitContainer@9283/@HSplitContainer@9285/@Control@9286/@SubViewportContainer@9287/@SubViewport@9288/Demo/TerrainPropGroup/MeshInstance3D5").`

### Positions

* Changing the position of an object in the inspector does not update the height, only with the gizmos.


* Moving a tracked object (that has been moved before) out of defined regions, might result in a nan position and eventually crash. Or,
  * If an object hasn't moved yet, it may allow you to move it out fine. Then if you move it back in to defined regions, the object no longer tracks the terrain.


* If a tracked object is on a region, then that region is removed, or
  * If a tracked object is on a void region and that region is added,
  * the position of that object is nan, the editor slows, and the console starts dumping errors:
`ERROR: Condition "!v.is_finite()" is true.
   at: instance_set_transform (servers/rendering/renderer_scene_cull.cpp:922)`

project/addons/terrain_3d/utils/transform_changed_signaller.gd
tcoxon: Yes, the if statement is redundant. This node is not meant to be used at runtime, so I will just leave the assert as a guard against this assumption being invalidated by future changes.

project/addons/terrain_3d/utils/terrain_prop_group.gd
tcoxon: Thanks, I was hoping you'd have some suggestions. Both sound good to me - I'll go with `Terrain3DObjects`.

project/addons/terrain_3d/utils/terrain_prop_group.gd
tcoxon: These asserts are meant to check assumptions that the code makes. They're not user errors, but rather guards against bugs being introduced in future. Have you found a way to trigger assert failure?

project/addons/terrain_3d/utils/terrain_prop_group.gd
```diff
@@ -0,0 +1,180 @@
...
+			terrain = terrains[0]
+		_terrain_id = terrain.get_instance_id() if terrain else 0
+	
+	if terrain and terrain.storage and not terrain.storage.maps_edited.is_connected(_on_maps_edited):
+		terrain.storage.maps_edited.connect(_on_maps_edited)
+	
+	return terrain
+
+
+func _on_child_entered_tree(p_node: Node) -> void:
```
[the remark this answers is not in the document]
tcoxon: Ooh, great catch there. I will definitely fix that.

project/addons/terrain_3d/editor/editor.gd
tcoxon: The issue is if you're placing a large object made out of smaller nodes, like for instance a complex building, into a `Terrain3DObjects` node, you don't really want the walls of the building to move out of alignment with each other. In that case, what a user wants is for the building to act as a single object, move all of it as one piece, and if it's necessary, adjust the terrain afterwards once they can see where it pokes through the floor or leaves gaps under exterior walls.

IMO it's better to use multiple `Terrain3DObjects` nodes. There isn't really any additional cost to doing so. What do you think?

project/addons/terrain_3d/utils/terrain_prop_group.gd
tcoxon: You're right. It's a holdover from an early version of this script. I'll remove it. Thanks!

project/addons/terrain_3d/utils/terrain_prop_group.gd
tcoxon: I have experimented with this before, but in practice it's difficult to get the edge cases right. I found that while I could make live height tracking work for _the objects under_ `Terrain3DObjects`, I couldn't do it for the `Terrain3DObjects` node _itself_. Doing so would break all the offsets and positions upon undo & redo, due to the order the transform changes get processed.

Implementing live tracking for the nodes under Terrain3DObjects, but not Terrain3DObjects itself, would leave it with inconsistent behavior.

If you definitely want this I can spend a bit more time on it. Otherwise, since it's so complicated and low impact, I think it makes sense to leave this for future investigation?

tcoxon:
> Moving a child node out of the Terrain3D adjacent TerrainPropGroup produced this error:

This is [Godot issue #80589](https://github.com/godotengine/godot/issues/80589). It happens whenever you reparent a MeshInstance3D node with the skeleton property set.

> Changing the position of an object in the inspector does not update the height, only with the gizmos.

Not sure what you mean here. I've edited the X, Y and Z coordinates in the inspector and the height updates correctly.

> Moving a tracked object (that has been moved before) out of defined regions, might result in a nan position and eventually crash. Or,
> If a tracked object is on a region, then that region is removed, or

Thanks! I'll fix this in the script by treating NaNs as 0 height, so that you can still move nodes outside of regions.

project/addons/terrain_3d/editor/editor.gd
TokisanGames: I see, yes we have houses made like that. And we have groves of trees under a node, placed via script.

What if instead of a Terrain3DObjects node, the user can assign a terrain3d_object group to any object. An actual Godot Group(tm). This script could fetch them all via `get_nodes_in_group()`. Then they can live anywhere in the tree. Then I would assign the root node of the house, and the individual trees to that group, and they'd stay organized where they were, and move with the terrain.

TokisanGames:
Right now edited_area is emitted for all tools. So all objects positions are adjusted for texture/color painting, and soon foliage. Maybe the signal should send maptype and operate only on heightmap.

project/addons/terrain_3d/utils/terrain_prop_group.gd
TokisanGames: Future is fine

TokisanGames:
As discussed on discord regarding updating children of a Terrain3DObjects node vs a Group(tm), the former is better because objects can be filtered by location. If there are a million objects, the AABB of the Terrain3DObjects nodes allows us to cut out huge amounts of objects that are not in the edited area. But notifying a group means all objects everywhere get and must process the signal. Though not implemented yet, positional filtering could be done.

The updates to this PR work well.

I still think there's another optimization improvement available, by emitting the map type so objects are not moved when non-sculpting operations occur.

Nice work as always, @tcoxon . Thanks!

PR #338, opened 2024-03-15
Create Asset Dock

### LINKED ISSUE #132: Make TextureList panel it's own dockable panel
In 4.1 panels are undockable. If you undock the inspector and redock it, TextureList doesn't follow and remains stuck in it's own half panel state. It should just be it's own panel.

https://docs.godotengine.org/en/4.1/tutorials/plugins/editor/making_plugins.html#doc-making-plugins

TokisanGames:
Fixes #132 

* Renames Texture dock to Asset dock
* Makes it a real dock that can be moved around or pulled into a separate window
* Converts it to a scene for better editability
* Fixes crashes in Terrain3DEditor
* Toggles between side dock and lower editor
* Hides dock when Terrain3D not selected
* Can pin dock to prevent above when used as a floating dock
* Asset texture scale limits or slider
* Panel retains position when switching scenes w/ two Terrain3Ds. Updates to material list of selected terrain3D.

image
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ export_presets.cfg
 # Editors
 .vscode/
 .vs/
+project/addons/explore-editor-theme/
--- a/project/addons/terrain_3d/editor/components/asset_dock.gd
+++ b/project/addons/terrain_3d/editor/components/asset_dock.gd
@@ -1,6 +1,7 @@
+@tool
 extends PanelContainer
 
-
+signal placement_changed(index: int)
 signal resource_changed(resource: Resource, index: int)
 signal resource_inspected(resource: Resource)
 signal resource_selected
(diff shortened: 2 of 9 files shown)


Saul2022:
Tested  it a bit and it look´s  pretty good already the dock colocation seem´s to work well, except when putting in to the bottom one. While it works and it add it to the bottom docks it too big and ca´t be dissable unless you select other node other than terrain.

It seem´s it may be a screen size issue, though this is the default sethings, nothing was changed.
 
editor_screenshot_2024-03-15T121513

TokisanGames:
Wait at least until it is not a Draft PR. 
The bottom editor can be closed, just click the blue Terrain3D in the bottom bar, same as any bottom editor.

PR #340, opened 2024-03-24
Add foliage instancer

### LINKED ISSUE #43: [Tracker] Improve foliage instancing
## Planned Improvements

* [ ] #932 
* [ ] Collision generation
* [ ] Apply transforms in mesh asset scene file
* [ ] Multi select and paint (checkbox in the lower left, or highlight)
* [ ] Change the Eye icon to MMI visibility instead of enable/disable for faster changes

------
* [ ] Drag and drop reorganize in asset dock
* [ ] Retain current height offset while sculpting like Terrain3d_objects does
* [ ] Foliage transform tool, a brush with expand/shrink scale, raise/lower, paint instance color.
* [ ] Create & destroy small MMI segments dynamically. 
* [ ] Improve generated grass: Meshless, vertex count, pointy, bend, gradient, wind
* [ ] AO/sample ground - wait for #245 

Optimizations
* Reusing RIDs
* Pre-Allocating MultiMeshes w/ maximum amount of memory
* Dynamically "create and destroy" (enable/disable/reuse preallocated buffers) MMI/MMs as the camera moves
* Updating the RS Buffer more directly with RS->multimesh_set_buffer(rid, buffer)
* Use RS->multimesh_set_visible_instances(rid, count) to limit the instances visible and the amount of the buffer used if using a preallocated maximum. Empty cells can be 0

### Biome projects
Whichever project is completed and integrated into this repo will be the most widely used.

* https://github.com/TokisanGames/Terrain3D/discussions/656
* https://github.com/jgillich/foliage3d
* https://github.com/caphindsight/Foliage3D

## Done
First pass of an MMI implementation was merged in #340, improved in #402 and individual commits

Redesign of the data structure was laid out in #524 and implemented in #523. 

* [x] Drag and drop into asset dock for meshes 99b64d768a5b5e715e92f9ccf714379a738546ec
* [x] Automatically adjust height of foliage on sculpting #424
* [x] Replace editor.cpp if instancer enable w/ add/subtract??
* [x] Option to remove all mesh types
* [x] Smaller MMI chunks 32-64/region
* [x] Store transforms in a data structure instead of an MM and generate as needed.
* [x] Store transforms in local space instead of global
* [x] Handle other meshes in scene file? - No, MMIs are for one mesh
* [x] LODs in scene files #604 
* [x] Use root node if MeshInstance3D #604 
* [x] #469 
* [x] Option to instance on all collision, not just terrain (eg grass on rocks) https://github.com/TokisanGames/Terrain3D/pull/732
* [x] Highlight MMI instances on the ground for the asset selected in the dock (temp swap override material), and/or add a picker #736 
* [x] Count used instanced #738 
* [x] Pick instance on ground to select in asset dock #734 
* [x] #788 

------------------
## Original

My idea for a paintable instancer is:
* Painting occurs on a splat/control map that can be saved in storage, but never goes in VRAM.
* MMI chunks are generated as the camera moves
* MMI chunks are regenerated if contained in the edited aabb of any terrain modifications

#324 and #278 need to be completed before this.

Reasoning:
* A control map based particle shader means placement around vertices with noise offsets. MMI nodes can be placed exactly where painted.
* Erasing instances is easy when painting on a control map. It's difficult to find and erase instances with an MMI, but regenerating a whole chunk while editing isn't a big deal.
* Particle shaders might be 2-4% faster, but will require yet another control map in VRAM. Fine if it's placement based on ground texture, but not if paintable. We're already consuming 4GB VRAM in OOTA and need to conserve.
* Particles can be frustum and distance culled. MMIs cannot, but with small enough chunks should be fine.
* MMIs seem to operate with Godot's autolod system (from what I see in SGT code). Haven't tested it, and don't know about particles.

 
--------------
## Other Notes

The current [control map format](https://terrain3d.readthedocs.io/en/latest/docs/controlmap_format.html) has 11 bits free, and most likely need 9 of them for paintable uv scale, slope, and rotation. 2 more could possibly be added to the base & overlay, or used for non-destructive layers.

Thus we probably need a second control map that could allocate foliage particles, water, and other items. However this other control map may not need to be in vram. It could be used to populate an MMI or a buffer filled with positions for a vertex shader.

Surprisingly, Godot's autolod apparently works with MMIs, according to SimpleGrassTextured, so this should be tested with MMIs and particle shaders.

### Other solutions
* [Hterrain](https://github.com/Zylann/godot_heightmap_plugin/) approach uses splatmaps to store [detail layers](https://github.com/Zylann/godot_heightmap_plugin/blob/master/addons/zylann.hterrain/hterrain_detail_layer.gd), then uses that data to generate MMI chunks

* [SimpleGrassTextured](https://github.com/IcterusGames/SimpleGrassTextured/tree/main) appears to be an MMI painter. See https://github.com/IcterusGames/SimpleGrassTextured/blob/main/addons/simplegrasstextured/grass.gd. 

* What do other solutions use? [MTerrain](https://github.com/mohsenph69/Godot-MTerrain-plugin), UE, Unity?

### My ideas
* A paintable control map Image that does not get stored in vram. This is used to generate MMI chunks. Chunks are regenerated upon sculpting updates within the editable AABB.
* Use a vram control map that feeds directly into a particle shader. Offer 2 particle types per vertex and a (paintable?) density. Several instances of the specified particles are placed around the vertex based on sampled noise and density.
* Collision could be generated on enabled particles, especially with the chunk based generator.  
* Compute based grass https://ueshita.hatenablog.jp/entry/2023/12/06/220451
* Either particle or MMI
* Should be paintable w/ multiple levels eg. paint grass, sunflowers, rocks
* Collision is not expected, but could be generated with our chunk system
* Maybe an option for the user to provide either a textured mesh, or just a 2D texture that gets applied to a generated quad

https://godotshaders.com/shader/wandering-clipmap-stylized-grass/
https://godotshaders.com/shader/wandering-clipmap-foliage-particle-process/

* Ghost of Tsushima style Grass
https://www.youtube.com/watch?v=bp7REZBV4P4&t=2s
https://www.youtube.com/watch?v=Ibe1JBF5i5Y
https://www.gdcvault.com/play/1027033/Advanced-Graphics-Summit-Procedural-Grass
Ansraer: "another neat trick is to use SDF for foliage alpha. That way you can gradually increase leaf size as objects get further away from the camera to hide the lower leaf density when it switched to a lower poly lod. If you are smart about it you could write a shader that only samples the alpha channel in the pre-pass, saving you from having to sample it in the main color pass. (due to the GPU's built in early discard feature)"
https://discord.com/channels/691957978680786944/1065519581013229578/1384197000802861057

* Consider rendering foliage without meshes
https://www.youtube.com/watch?v=R-bjXOEQyX8
https://github.com/godotengine/godot/issues/19473
https://github.com/godotengine/godot/issues/83446

* GL Grass
https://github.com/SomePineaple/GLGrass
https://www.youtube.com/watch?v=2h5NX9tIdno

* https://gpuopen.com/learn/mesh_shaders/mesh_shaders-procedural_grass_rendering/
* https://www.youtube.com/watch?v=jw00MbIJcrk

* Ghost of tsushima style grass in Godot
https://github.com/2Retr0/GodotGrass

TokisanGames:
Partially addresses #43
Partially addresses #368

This PR extends the asset dock to include mesh assets.
A PackedScene can be dropped in, meaning a .glb, .fbx asset, or a .tscn. It will search the scene for all MeshInstance3Ds and currently selects the first one found.

### General Changes Done
* ~~`Opacity` renamed to `strength`, and >100 is allowed for all tools~~ (merged)
* ~~Tool settings are now shown instead of hidden, including brush selector~~ (merged)
* ~~CTRL Key negative input for `Enable`~~ (merged)
* ~~Undo/redo redesign system~~ (merged)
* Redesign flexible asset dock that can move to all positions and save settings in EditorSettings
* Rename Terrain3DTexture -> Terrain3DTextureAsset
* Rename Terrain3DTextureList -> Terrain3DAssets

### Foliage Instancer Done
* Quadmesh populates where painted into an MMI, with correct height and normal
* Restrict foliage brush to ring
* Instancing evenly distributed within circle indicator, fix square placement
* Remove MM instances (with CTRL, or `Enable` unchecked)
* Create an Instancer resource class that can be saved
* Create UI for mesh assets and textures
* Create instanceable mesh object
* Create MMIs in chunks
* Move data to Storage
* Create instanced object settings for:
  * Height offset (adjust origin)
  * Quad mesh w/ texture and/or material
  * Custom mesh
* Tool settings toggle for applying ground normal, height offset
* Check default asset dock position w/ default settings, place bottom right
* Change generated mesh card option
* Undo/redo for MMIs
* Document

### Pushed to future PRs
* Drag and drop into asset dock for meshes
* Adjust heights on sculpting
* Improve generated grass: enum, Meshless, vertex count, pointy, bend, gradient, wind
* AO/sample ground?
* Create & destroy small MMI segments dynamically. Waiting on #278
* Smaller MMI chunks 32-64/region
* Collision generation
* Apply transforms in scene file
* LODs, except what is provided by MMI and autolod
* Handle other meshes in scene file? - probably not

image


Saul2022:
Looking great as far as i tested(ik it pretty early , but couldn't wait to check  it out) even with shadows seem's like performance increase a bit as it hiding the  terrain mesh, though something i notice is that using transform  in terrain  changes the grass location

TokisanGames:
Alright, I have disabled transforms.

TokisanGames:
I did a major overhaul on the asset dock and need testers. I will likely merge this commit into the main tree once it's polished, before this foliage PR is finished.

What should work and needs testing:
* Move asset dock to bottom or any sidebar position
* Resize icons
* Texture selection/adding/removing/editing settings should still work like normal
* Switching between scenes w/ seperate Terrain3D instances open the texture list should update
* Save the layout settings and reload them on restart - Note you must save your project before this will work, as this configuration is saved in Project Settings
* Popout to a window w/ following behavior:
  * Pin that window so its always on top or not
  * Saving the project (CTRL+S) while the window is focused
  * Window Focus follows the mouse between godot and asset dock so that the mouse cursor and painting works fine in the viewport

Using these options do work, but they don't fire any signals that I can detect without hacks
image

But I made my own buttons which should be used for moving position or popping out
image


Other notes:
* Mesh/Texture buttons do nothing, but will act like tabs
* Needs to be cleaned up and prints a lot of troubleshooting to the console.

This binary can be tested:
https://github.com/TokisanGames/Terrain3D/actions/runs/8587341009

Saul2022:
Testing:

1 Moving the dock on every sidebar or bottom seem´s to work well through when the terrain 3d node is selected and the asset dock is at the bottom  it not possible to see the other tabs, meanwhile when you diselect the node and  click on the tab it works well without overlapping the other docks. ( althought that is fixed when you make the editor fullscreen.

editor_screenshot_2024-04-17T184339
editor_screenshot_2024-04-17T184550
editor_screenshot_2024-04-17T184651


2 Resizing  works but when you make it way too big  it not possible to  edit the corresponding texture as you can´t acces the three dots or delete. 

editor_screenshot_2024-04-17T184847


3 Works as usual

4 I am not sure what  you mean, but as far as i tried having two scenes with the same textures/ material list save to save, for example i add one new material to a scene, then save and then on the other scene the textures remain as positioned and the same as the one i recently saved.

5 Don´t have layout.

6not sure if it fine or made something wrong but sseems that when pinned it stays on screen within other tabs and when saving it stays on the same position it was put before saving, though when minimizing the godot  app the screen stays in other programs unless you minimize it too, so maybe when minimizing godot you could make it so the pinned dock is also minimized, same when you open it again. 

editor_screenshot_2024-04-17T190735

TokisanGames:
Thanks for testing @Saul2022 .

> 1. when the terrain 3d node is selected and the asset dock is at the bottom  it not possible to see the other tabs, 

This when Godot is in a window, or maybe a narrow height screen resolution? This is due to the minimum height of the panel. The other bottom tabs do the same thing. Especially click Audio, which has a very tall minimum height. The bar drops way below.


> 2 Resizing  works but when you make it way too big  it not possible to  edit the corresponding texture as you can´t acces the three dots or delete. 

You can right-click to edit or middle click to clear/delete.

> 4 I am not sure what  you mean, 

Two scenes open: Demo and imported. Each has a copy of Terrain3D, with different storage and asset dock. Switching back and forth between the scenes, the asset dock should populate with their respective textures, at least if Terrain3D is selected.

> 5 Don´t have layout.

Layout is the dock position, size, pinned. This is saved to project settings so should be retained when you restart godot, as long as it doesn't crash.

> 6 not sure if it fine or made something wrong but sseems that when pinned it stays on screen within other tabs

The pin means always on top. It should stay on screen.

>  and when saving it stays on the same position it was put before saving, 

The dock should stay at the position it was in before you saved. I don't understand. You move the doc to bottom right, then save it, it should stay at the bottom right. 

> so maybe when minimizing godot you could make it so the pinned dock is also minimized, same when you open it again. 

That's a good idea. I don't know if I have access to detect that, though.

TokisanGames:
Updates:
* Reduced maximum size of tiles so the delete/edit buttons don't slide off the sidebar
* Windowed asset dock now hides when Godot is minimized
* Window focusing much more subtle and effective

Bottom bar disappearing I'll ignore for now as I think it's a godot issue, as they all do it.

darrylryan:
I see no reason why you couldn't also paint trees and bushes using this same system. The only differences would be:

1. The spacing would need to be calculated from the AABB of the mesh
2. The density would need to be capped much lower for large meshes.
3. We couldn't draw complex meshes into the far distance and far away chunks couldn't just "disappear" for items large enough to be seen from miles away or it would look weird. A good solution would be to use this: https://github.com/zhangjt93/godot-imposter to generate an  impostor for each mesh and then set the MM to the real tree for closest chunk and all the further away chunks could just be impostors (which in effect work much like the current grass system).

TokisanGames:
Update:
* The asset dock can now load packed scenes for meshes. It will automatically locate the first MeshInstance.
* There's an option for generated grass, which is currently just the basic green quad.
* Multiple MMIs are created or destroyed when painting, based on location and mesh id. Currently this breaks down by region, but will be smaller segments later. This is saveable.

image

Saul2022:
Tested on latest action and seems like when deleting an mesh instance, engine crashes inmediatly.

Also modifying the new properties like color, height etc,seems to not work with neither the generated blade nor a custom mesh.

Furthermore when you try replace the terrain plugin folder on an existing project that had it already(which in this case was the one used to test last weeks grass project) it made the scene corrupted for some reason.

TokisanGames:
> Tested on latest action and seems like when deleting an mesh instance, engine crashes inmediatly.

Ok, I see, thanks. You mean removing a MeshAsset from the asset dock, not removing an instance with CTRL.

> Also modifying the new properties like color, height etc,seems to not work with neither the generated blade nor a custom mesh.

Only the UI elements for the tool settings are there. The functionality hasn't been implemented yet.

> Furthermore when you try replace the terrain plugin folder on an existing project that had it already(which in this case was the one used to test last weeks grass project) it made the scene corrupted for some reason.

Not "corrupted", just unparseable because you saved a resource with a different name inside. Easily fixable by editing the file to the new resource name.

Whatever you tested on last week is incompatible. Assuming you're using the demo, reset the project back to the main branch. If using your own project, that's dangerous. Reset your terrain3D resource files.

> And grass keeps dissapearing whenever i switch between scenes ( in this case a custom billboarder scene)

Ok, I see. 

Thanks for the reports.

Saul2022:
> > Tested on latest action and seems like when deleting an mesh instance, engine crashes inmediatly.
> 
> Ok, I see, thanks. You mean removing a MeshAsset from the asset dock, not removing an instance with CTRL.
> 
The first yea

TokisanGames:
Update:

* Fixed meshes disappearing when changing scenes
* Fixed crash on deleting MeshAsset from dock
* Countless improvements
* Implemented all paintable instancer settings:
  * height offset (per mesh and painted), scale, rotation w/ random options
  * Vertex color, randomized hue shift and darken
  * Materials must have `vertex_color_use_as_albedo` enabled for this to work.
  * `Random hue shift` shifts the color specified in vertex color. If saturation is 0, the hue shift won't be visible.


image

image

image


* Data from older versions of this PR are invalid, revert terrain_storage.res to main

-------
I need help with one thing in this bit of code:

https://github.com/TokisanGames/Terrain3D/blob/f593c53bac268f3e7555b166f6870e3009f3f96e/src/terrain_3d_instancer.cpp#L228-L260
 
* align_to_normal: works, object grows perpendicular to terrain face
* rotate_y_only: works, object rotates around it's local y, even if align_to_normal is enabled
* height_offset: works for both above options and offsets the mesh along its normal

What doesn't work is if rotate_y_only is disabled, regardless of align_to_normal or not, I want the base to stick in the ground and the top to point in any direction.

You can see the tops do go in any direction, but some of the bases float above the ground.

image

TokisanGames:
Fixed rotation, refined other settings, and changed the generated mesh to a default card that can be textured.

This PR can start being tested with [this build](https://github.com/TokisanGames/Terrain3D/actions/runs/9301135793). Backup your storage and texture list resources, and scene.

The last feature to add is undo/redo, which will come soon.

TokisanGames:
Undo/redo now works with the instancer. This PR is feature complete. We need testers so we can merge it in a few days. Test this build. Backup your scene/storage/asset resources.

https://github.com/TokisanGames/Terrain3D/actions/runs/9330335676

JekSun97:
@TokisanGames everything works fine, except that the player cannot collide with static objects, although there is collision, or is that intended?

I also noticed a strange thing, I created this cube, slightly elongated on the sides

but in the docking station, and when it is placed, it becomes not elongated, but an ordinary cube

TokisanGames:
Thanks for testing. As stated, no collision generation in this PR.

The PR takes the mesh from the first meshinstance in the file. Did you make the boxmesh long in its settings? Or did you scale it? Do the former.

JekSun97:
Everything is fine, I just didn’t immediately understand that only the size of the mesh itself is taken into account, and not its scale in Transform

TokisanGames:
Read the bottom of the top post for plans. All data, no. Meshes, static collision, possibly transforms.
This isn't a scene placer. It's a multimeshinstance painter. To place scenes use scatter or AssetPlacer.

Saul2022:
Testing a bit, and as far as i can check undo redo, seems to not worl( doing Crtl z and nothing happens with the meshinstance, neither it does on history tab).

Then when i tried using a tree ( that composed of two meshes, the foliage and the trunk itself)as an instance, it only took into account the foliage, without the poor trunk .

editor_screenshot_2024-06-05T120252

Then when i joined those in blender using the append function, and importing to godot as a glb it rotation worked just fine in blender and using godot base transform, but when appilying as an terrain3d instance, this happens.

editor_screenshot_2024-06-05T115557

TokisanGames:
Thanks for testing. 4.3 is definitely not supported, please test with 4.2.2. Your undo tab should fill up with every operation, including instancing. 

It currently uses only the first mesh instance it finds in the scene file.

It looks like through your importing and exporting you roasted the transform of the tree. I bet in your scene it's currently rotated, and the default transform is laying flat. Reset the transforms in blender and import again, ensuring all nodes in your scene has an empty transform.

Saul2022:
> Thanks for testing. 4.3 is definitely not supported, please test with 4.2.2. Your undo tab should fill up with every operation, including instancing.
> 
ok fine though 4.3 as far as it goes really looks like working (except on that damm 4.3 beta 1) though ye at the end downloading a new version of 4.2 worked, though for 4.3 beta 1 everything with parses broke(not plugin fault).

TokisanGames:
> ok fine though 4.3 as far as it goes really looks like working 

Except your undo/redo doesn't work, and many things in the new version, and now it "upgraded" and broke your scenes so you can't go back easily. That's why we won't touch it until the RCs.

I'm going to merge this in. Thanks to the couple of testers. More improvements will come as in subsequent phases but this should be usable now for more testers.

Saul2022:
> Except your undo/redo doesn't work, and many things in the new version, and now it "upgraded" and broke your scenes so you can't go back easily. That's why we won't touch it until the RCs

Well as for scenes breaking it was me being a complete blind(disability dumb stuff)  not seeing that i installed 4.1 beta instead of 4.3 beta1 😅, atleast i had had copies alr so i didn't loose anything.

> I'm going to merge this in. Thanks to the couple of testers. 

Nice

TokisanGames:
Merged. Track additional improvements in #43

TokisanGames:
> Except your undo/redo doesn't work

I have no problems using 4.3 beta 1 with a nightly build with this PR merged in. Mouse cursor, Undo/redo works fine.

Issue #318, opened 2024-02-05
Misleading naming in shader


outobugi:
### Description

There is some confusing naming going on in the shader. Partly my fault.

- `get_region_uv()` -> returns texture coordinates, not UV
- `uv_center` in `get_material()` is not related to material UV and is texture coordinates referring to domain's/control map pixel's center position

Rename `get_region_uv` to `get_region_coords` and `uv_center` to `domain_center` or `cmap_pixel_center` for absolute clarity.

TokisanGames:
@Xtarsia Can you offer your thoughts please?

Xtarsia:
get_region_uv() could indeed be more aptly named, perhaps get_region_texel_coord()

uv_center change makes sense too, perhaps index_coord instead.

Xtarsia:
get_region_uv() -> get_index_coord()
get_region_uv2() -> get_index_uv()

uv_center -> index

Issue #284, opened 2023-12-20
Colormap artifact in mipmaps along region boundaries


TokisanGames:
### Terrain3D version

v0.9.0

### System information

Godot v4.1.3.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 Laptop GPU (NVIDIA; 31.0.15.4633) - 12th Gen Intel(R) Core(TM) i9-12900H (20 Threads)

### Is the issue reproducable in the demo?

Yes

### Issue description

Mipmaps were generated on the colormap array in 7e4f9ebe985236ba0a45d13581847193b46a4eb1 to avoid distant noise, which was very obvious when using a full colormap such as in GIS applications w/ satellite photos (farther below). However, this has introduced another artifact on the region boundary. Enabling `renormalize` on `generate_mipmaps()` has no affect.

`ALBEDO=colormap*vec3(.1)`

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/08a44db7-d5a1-46eb-8dc4-c969f8464593)

Regular shader

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/c13f44ae-841b-4579-a119-514cee418d0e)

Though mipmaps should be generated on the whole map at once, it appears this tiling is not a problem and the individual mipmaps blend together fine. It's actually surprising that they blend at all, given how different from each other the lower lods look. 

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/7ec706aa-3fc7-4b18-b25c-aaa937bb05ab)

The only artifact is the small grid line that shows up when up close on mipmap0 shown at the start, and also far away shown here.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/ebc66d92-0ab0-4e7a-a178-2ed3d5fb2dea)

Upon further testing, I've found that indeed it is not the mipmap lods, it is actually the system that determines which LOD to use that is the problem.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/e3ef8053-ec66-49d1-806f-0be274ef81ef)

These lines are the borders of the individual textures in the arrays and show up when using `texture()`.

I have found I can clamp the lod based on distance.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/6cbc6e9c-0623-48ff-b21d-ce79c27074fd)

[...]

Issue #185, opened 2023-08-16
Normal artifact along region boundaries


TokisanGames:
This artifact appears on regions borders. Along a region border make multiple heights at 40, 20, 0, -40 then start smoothing them together and it will appear. The thickness of this artifact appears to be 2 vertices, and is offset. So on X it goes from 0.5 to -1.5.

The artifact is in the normals. It occurs in the fragment() normal calculation. It has to do with using texture() with filter_linear, as it linearly blends the edge of the height map in it's normal calculation. If we switch to filter_nearest or texelfetch() it does resolve, however the normals are blocky. 

Possible solutions:
* use texelfetch but bilinearly interpolate between the lookups to smooth them, rather than letting the card do it
  * We can't do it only on the vertices (looked up in vertex()) because they spread out on lower lods, but
  * We can use the same point grid and bilinear interpolation that the material lookups use.
* generate normal map and pass it in to the shader
  * perhaps this can be done upon mouse release for the section that was just modified, then blit in to the normal map
  * when GPU painting works, could be generated by the compute shader on the fly, but will it have the same issue?

![image](https://github.com/outobugi/Terrain3D/assets/632766/5be61fe4-c4e7-4d67-b4a3-b036e497ec8b)

![image](https://github.com/outobugi/Terrain3D/assets/632766/5eee37b1-d408-49a6-afa0-276aa2dd3537)

Testing notes:
- offset collision half a texel instead of shader - no it's the shader that's off due to filtering/interpolation
- vertex, calculate 9 heights around, say in varying, then using textureGather
- set map to nearest & use texel lookup
- fragment linear interpolate normals (or heights) from vertex neighbors, lerped based on position
- use texel lookup per pixel, but lerp it
- if on edge of region use texel otherwise texture?

I tried clamping to the region so it wouldn't sample across boundaries and just average, that reduced the size but didn't help
```
	vec2 region_min = floor(uv/_region_size); //1023/1024->0, 4095/4096->0
	vec2 region_max = region_min + vec2(_region_size);
	float left = get_height(clamp(uv + vec2(-_region_texel_size, 0), region_min, region_max));
	float right = get_height(clamp(uv + vec2(_region_texel_size, 0), region_min, region_max));
	float back = get_height(clamp(uv + vec2(0, -_region_texel_size), region_min, region_max));
	float front = get_height(clamp(uv + vec2(0, _region_texel_size), region_min, region_max));
```

TokisanGames:
[...]

PR #348, opened 2024-04-13
UV2 texel centered in vertex func, but not un-centered in get_region_uv2 when it's converted to layer offset.

FishOfTheNorthStar:
UV2 is texel-centered in Vertex Func, so it should be un-centered in get_region_uv2 to keep the last row and column within the correct layer of the texture array for get_height / get_normal functions.  
The vertex func adds the offset in to keep the height samples centered.  Does anything else actually use the centered texel? It might be cleaner to just add the center in for the vertex function but otherwise leave it basically unaltered, so it can be mod'd later without trying to figure out first if its the pixel centered version or not.

### DIFF
--- a/project/addons/terrain_3d/extras/minimum.gdshader
+++ b/project/addons/terrain_3d/extras/minimum.gdshader
@@ -1,4 +1,4 @@
-// This shader is the minimum needed to allow the terrain to function.
+// This shader is the minimum needed to allow the terrain to function, without any texturing.
 
 shader_type spatial;
 render_mode blend_mix,depth_draw_opaque,cull_back,diffuse_burley,specular_schlick_ggx;
@@ -12,8 +12,19 @@ uniform int _region_map_size = 16;
 uniform int _region_map[256];
 uniform vec2 _region_offsets[256];
 uniform sampler2DArray _height_maps : repeat_disable;
+uniform usampler2DArray _control_maps : repeat_disable;
+uniform sampler2DArray _color_maps : source_color, filter_linear_mipmap_anisotropic, repeat_disable;
+uniform sampler2DArray _texture_array_albedo : source_color, filter_linear_mipmap_anisotropic, repeat_enable;
+uniform sampler2DArray _texture_array_normal : hint_normal, filter_linear_mipmap_anisotropic, repeat_enable;
 
-varying vec3 v_vertex;	// World coordinate vertex location
+uniform float _texture_uv_scale_array[32];
+uniform float _texture_uv_rotation_array[32];
+uniform vec4 _texture_color_array[32];
+uniform uint _background_mode = 1u;  // NONE = 0, FLAT = 1, NOISE = 2
+uniform uint _mouse_layer = 0x80000000u; // Layer 32
+
+varying flat vec2 v_uv_offset;
+varying flat vec2 v_uv2_offset;
 
 ////////////////////////
 // Vertex
@@ -34,12 +45,17 @@ ivec3 get_region_uv(vec2 uv) {
 // XY: (0 to 1) coordinates within a region
 // Z: layer index used for texturearrays, -1 if not in a region
 vec3 get_region_uv2(vec2 uv) {
-	ivec2 pos = ivec2(floor(uv)) + (_region_map_size / 2);
+	// Vertex function added half a texel to UV2, to center the UV's.  vertex(), fragment() and get_height()
+	// call this with reclaimed versions of UV2, so to keep the last row/column within the correct
+	// window, take back the half pixel before the floor(). 
+	ivec2 pos = ivec2(floor(uv - vec2(_region_texel_size * 0.5))) + (_region_map_size / 2);
 	int bounds = int(pos.x>=0 && pos.x<_region_map_size && pos.y>=0 && pos.y<_region_map_size);
 	int layer_index = _region_map[ pos.y * _region_map_size + pos.x ] * bounds - 1;
+	// The return value is still texel-centered.
 	return vec3(uv - _region_offsets[layer_index], float(layer_index));
 }
 
+// 1 lookup
 float get_height(vec2 uv) {
 	highp float height = 0.0;
 	vec3 region = get_region_uv2(uv);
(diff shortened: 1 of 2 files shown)


TokisanGames:
Thanks. I'm thinking about it and need to allocate the time to test it.

Note that any or all of the variable names, function names, or comments might be wrong due to a misunderstanding of what we actually wrote or copied from elsewhere. So, I appreciate efforts to ensure accuracy and clarity.

I'd like your take on these related issues, also referenced farther below:

Long standing normal artifact w/ pending workaround. Perhaps this PR helps. #185 

Slightly related. Renaming get_region_uv might lead to renaming get_region_uv2. #318

There was a colormap artifact Clayjohn helped me resolve. See his answer at the bottom. #284

UV is transformed to work around precision loss on mobile #296 

-----
> The vertex func adds the offset in to keep the height samples centered. Does anything else actually use the centered texel? 

Looking through the shader, here's how UV/UV2 are used: 
* `_uv` and UV, int 0-1024+ range, (which Roope says should be texel uv, though I have it labeled as worldspace):
  * A UV [transformation was applied in vertex and unapplied in fragment #296](https://github.com/TokisanGames/Terrain3D/pull/296/files#diff-ddd98455d1d9f905df13f0b57cc27f6eb11f8ad776c54d742d114cccfbcdab83) by @lfxu for vertex spacing and mobile precision.
  * Determine if vertex is a hole (v_region f/ control map lookup in vertex()), unused in fragment
  * 4 adjacent vertices (index00-11) for control map lookups to be bilerped in fragment
  * Used with noise scale because I thought it was world space, but if it's texel space :/
* `_uv2` and UV2, float 0-1+ range:
  * centered on texel with 0.5 offset in vertex
  * UV transform applied at the end of vertex and unapplied in fragment
  * get_height() - needs texel center offset so it aligns w/ collision, IIRC
  * get_normal() - and there's a normal artifact on region borders #185
  * colormap lookup to determine if the fragment is in a region - and there was an artifact from my original implementation #284




cc: @Xtarsia, @lfxu, @outobugi

Xtarsia:
regarding this and #185, without any workaround applied, the normal artifact at region boarders diverges by half a vertex, but still persists. the pending workaround still works.

the currently applied partial workaround looks worse

FishOfTheNorthStar:
What i've done to correct the normal issue is basically this:
when iterating across the verts within vertex(), imagine we have a loop for (x=0->region_width-1) {ie, 1023, or mod(VERTEX.xy, region_width, etc), now within that loop there's two sets of very similar co-ordinates in use.  The first is basically how it does it currently, convert texel index into normalized range,  and sample the height from that.  But the second UV set used in that loop takes the normalized X/Y and multiplies it by (1.0-region_texel_size), so that second sets range is more like 0->0.99something.  From that second set, sample normals.  Note: I calculate all normals within vertex function but your current version is still doing that per pixel, i suspect to help the panels be more seamless between resolution jumps.  So some of this explanation might sound weird.  Also, before sampling normals, I offset the 'upper left' (first tap) by half a pixel towards 0,0, because I found the half pixel error of the first edge could actually be salvaged as usuable in this way.  That leaves the last pixel edge as unusable, thus the adjustment from 0->1 to 0->0.999whatever.

So the end result of this is spreading the error pixel row/column across the entire width, and in this way it basically completely disappears.  I can't detect it anyways, in Grey view.

I've been thinking about a more durable fix and I think i have a plan that would 100% mitigate the issue forever, and considerably improve performance at the same time, but I suspect you won't like it:

Encode an extra 1 pixel margin along each edge of all control maps / height maps.  So row 0 and row 1023 (and same columns), those are just repeats of row 1 and row 1022, basically.  Then when in use, normalize texel offsets to lie within 0+texel_size -> 1.0-texel_size. 

Why would this help?  Well for one, you can just stop checking which region neighbor pixels are within.  They'll always 100% of the time be in the same region.  But also, because of this you can use textureGather for most things involving control maps and height maps, which is far more performant in my tests so far.  It's effectively one sample to get four, with no 'calculate the neighbor UVs' step at all.  Also, the normal issue at edges will be a thing of the past and it will never return.  And from within the shader, the actual changes involved to support this are fairly minimal.  I think the real problem with implementing this would be within the editor and mapping changed pixels to correct region texels, and I haven't looked at that code at all yet.  Also, of course, migrating existing datasets over to new format... oof.  Whatever, skipping that part for now.

In my last commit, the normal issue is gone, but be warned I changed the code with reckless abandon, that did introduce other artifacts and that is why i have not brought that code over here for review yet.  But you can have a look if it helps this TL;DR make more sense.

I've made more changes locally i didn't push yet but will soon, and they help perfomance quite a bit.  Probably the single biggest change is moving all the control map lookups into the vertex shader and storing the results inside a varying uint4 which get_material then uses and skips all of it's previous i00, i01, etc etc index calcs.  Gone.  At the same time, it also stores the region index of each of those samples (since they might be on another panel), into another varying int vec.  Also the random angle calcs, cached into a varying vec4 (one entry per neighbor, and one for self), so none of that happens in the fragment shader anymore, and that seems to help quite a bit.

Message getting too long.  More later.  Good luck and thanks again to everyone involved.

TokisanGames:
Regarding this PR:

I see no visual difference, except making the region normal artifact worse, which we have some solutions for. Why do we need it? Correctness is fine, but is it the right solution? More below.

> UV2 is texel-centered in Vertex Func, so it should be un-centered in get_region_uv2 to keep the last row and column within the correct layer of the texture array for get_height / get_normal functions.

I don't fully understand the issue. For last row and column, say 1023/1024, UV2 = .999, .999
```
Old: 
ivec2 pos = ivec2(floor(uv)) + (_region_map_size / 2); 
ivec2 pos = floor(vec2(.999)) + 8
ivec2 pos = (0, 0) + 8

New: 
ivec2 pos = ivec2(floor(uv - vec2(_region_texel_size * 0.5))) + (_region_map_size / 2);
ivec2 pos = floor(vec2(.999)) - vec2(0.00048828125)) + 8
ivec2 pos = (0, 0) + 8

```

What am I missing?


> The vertex func adds the offset in to keep the height samples centered. Does anything else actually use the centered texel? 

As listed above, only get_height/normal need the center AFAIK. So why this solution? Why not remove the offset from vertex() and put it only in get_height instead?

> It might be cleaner to just add the center in for the vertex function but otherwise leave it basically unaltered, so it can be mod'd later without trying to figure out first if its the pixel centered version or not.

I'm sorry, I don't understand this.

FishOfTheNorthStar:
Hi, I was trying to think of a clean way to explain the problem (as I think it is) within these text comments.   Here's my best attempt:

here's a "line of pixels", with the same letter twice in a row to represent each pixel.  Two copies of the letter because in the example we'll be shifting them a half pixel over, and then mod'ing them against the width, to see where they end up.

original is like:
AABBCCDDEEFFGGHH etc.  Let's say the image with is 4, and we mod that row against '4 wide', we might end up with:
AABBCCDDAABBCCDD etc.

But now take the same, and add a half pixel space, we get

?AABBCCDDEEFFGGHH  etc.  the ? represents pixel from undefined source.

Modded against the same hypothetical '4 pixel wide image', we get:

?AABBCCD?AABBCCD?AABBC.... etc

So I think maybe that shows the problem?  the first half of the first row/column, after being offset by half a pixel, no longer have a source available to supply the first half of their respective row/column's pixels.


BTW totally agree about arcane variable names, that "vgh" one is a side effect of running out of 'maximum string length' space.  I was running into situations where adding bits of code to main.glsl would no longer compile because the file was getting too long to include the way it does currently.  I've since shuffled some  things around to open up space in main.glsl to avoid that.

TokisanGames:
Thanks!

PR #353, opened 2024-04-18
Update main.glsl low LOD vertex normals

### LINKED ISSUE #185: Normal artifact along region boundaries

Xtarsia:
branching for vertex normals to fix #185 caused performance loss. so make that branch as big as possible without visual loss, whilst also gaining performance!

moved a couple things in the vertex function, and removed redundant matrix multiply.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,3 +16,5 @@ Contributors
 * [@directedchaossoftware](https://github.com/directedchaossoftware)
 * Jacob Coughenour [@jacobcoughenour](https://github.com/jacobcoughenour)
 * [@stakira](https://github.com/stakira) / [@lfxu](https://github.com/lfxu)
+* [@XTarsia](https://github.com/XTarsia)
+* Matt [@FishOfTheNorthStar](https://github.com/FishOfTheNorthStar)
--- a/doc/docs/shader_design.md
+++ b/doc/docs/shader_design.md
@@ -24,7 +24,7 @@ Optionally, world noise is inserted here, which generates fractal brownian noise
 
 `get_height()` returns the value of the heightmap at the given location. If world noise is enabled, it is blended into the height here.
 
-Finally `vertex()` sets the UV and UV2 coordinates, and the height of the mesh vertex. Elsewhere the CPU creates flat meshe components and a mountainous collision mesh. Here is where the flat meshe vertices have their heights set to match the collision mesh with `VERTEX.y = get_height(UV2)`.
+Finally `vertex()` sets the UV and UV2 coordinates, and the height of the mesh vertex. Elsewhere the CPU creates flat mesh components and a mountainous collision mesh. Here is where the flat mesh vertices have their heights set to match the collision mesh with `VERTEX.y = get_height(UV2)`.
 
 ## Fragment()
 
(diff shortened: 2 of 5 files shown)


Xtarsia:
@TokisanGames This should be ready now...

Performance with this vs without is generally better when on the ground (+0% to +20%), slightly worse when airborne, and it is significantly faster in all cases when world_noise is on?

The default range of 128 is fairly conservative, pushing higher makes meshLOD transitions more visible.

past 500, can gain quite large FPS boosts, whilst still maintaining usual detail on mountains in the distance etc.

and of course, no more normal artifacts :)

Xtarsia:
Distance should be based on 2D plane only - solves performance drop when flying.

Note: in testing the transition is least noticable when the distance is a multiple of mesh size.

The demo has mesh size set to 64, so 128 value is imperceptible.

The default setting for a new project is 48 however, where 96 or 144 are better values.

TokisanGames:
I do see some artifacts. In grey scale mode, when I set world noise blend near <0.75 and far 1. It's even worse when the camera is positioned near it.

image

At near 0.8 the world noise has enough margin so it doesn't blend at all, nor produce an artifact.

image

Overall, it's a great improvement, but perhaps you want to look at the blending algorithm in get_height?

Also I think vertex_normals_margin can be hard coded to 8-16. It doesn't provide any artistic control, so it will be one less slider in the list.

Xtarsia:
This also fixes an issue with incorect height reads for the world noise, despite the world noise per pixel normals being correct.

PR #365, opened 2024-05-01
UV rotation & scale painting

### LINKED ISSUE #145: Add detiling option to the shader

### LINKED ISSUE #322: Improve texture scale & rotation
### Description

* Rename Texture/`UV Rotation` to `Detiling`
* Add debug views for UV Scale/rotation
* Texture rotation doesn't properly rotate normals. It rotates them like an image rather than a 3D object. This video shows how https://www.youtube.com/watch?v=iXqCszt3LbE

---
Other items completed or tracked elsewhere

Brushing with tools should be able to change only certain features such as:
* ~~Paint UV scale / rotation, but don't change texture~~ Implemented in #365
* ~~Limit painting of color, texture, foliage to texture mask that are already present on the terrain(eg. inspector layers)~~ Duplicate #192

Xtarsia:
de-tiling improvements as well. ( fixes #145 )

Currently
~~5 bit angle, 11.25 degrees, 360.~~
4 bit angle, 22.5 degrees, 360.
3 bit scale, 20 scale step, 

Toggles for scale/uv on/off in both Roller and Spray.
Movement based directional painting mode for angle.
~~Seperate brush for just UV painting, with both toggles as well.~~

Feedback appreciated!

~~(will do Doc update later depending on feedback)~~
some Docs added.

------
Admin edit: Partially fixes #322

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -221,6 +221,13 @@ vec4 height_blend(vec4 a_value, float a_height, vec4 b_value, float b_height, fl
 	}
 }
 
+vec2 uv_detiling(vec2 uv, vec2 uv_center, int mat_id){
+	vec2 rotation_center = floor(uv_center) + 0.5;
+	float random_angle_1 = (random(rotation_center) - 0.5) * 2.0 * TAU; // -180deg to 180deg
+	vec2 outUV = rotate_around(uv, rotation_center, random_angle_1 * _texture_uv_rotation_array[mat_id]);
+	return outUV;
+}
+
 // 2-4 lookups
 void get_material(vec2 base_uv, uint control, ivec3 iuv_center, vec3 normal, out Material out_mat) {
 	out_mat = Material(vec4(0.), vec4(0.), 0, 0, 0.0);
@@ -229,16 +236,32 @@ void get_material(vec2 base_uv, uint control, ivec3 iuv_center, vec3 normal, out
 
 //INSERT: AUTO_SHADER_TEXTURE_ID
 //INSERT: TEXTURE_ID
-	float random_value = random(uv_center);
-	float random_angle = (random_value - 0.5) * TAU; // -180deg to 180deg
-	base_uv *= .5; // Allow larger numbers on uv scale array - move to C++
-	vec2 uv1 = base_uv * _texture_uv_scale_array[out_mat.base];
-	vec2 matUV = rotate_around(uv1, uv1 - 0.5, random_angle * _texture_uv_rotation_array[out_mat.base]);
+	
+	// Control map scale & rotation, apply to both base and center uv.
+	// To correctly rotate & scale around each control map pixel we must 
+	// translate uv center from "control map space" to "uv space".
+	uv_center += _region_offsets[region] * _region_size;
+	// Define and apply base scale from control map value as array index. 0.5 as baseline.
+	float[8] scale_array = { 0.5, 0.4, 0.3, 0.2, 0.1, 0.8, 0.7, 0.6};
+	float control_scale = scale_array[(control >>7u & 0x7u)];
+	base_uv *= control_scale;
+	uv_center *=  control_scale;
+	// calculate baseline derivatives
+	vec2 ddx = dFdxCoarse(base_uv);
+	vec2 ddy = dFdyCoarse(base_uv);
+	// Apply global uv rotation from control map.
+	float uv_rotation = float(control >>10u & 0xFu) / 16. * TAU;
+	base_uv = rotate_around(base_uv, vec2(0), uv_rotation);
+	uv_center = rotate_around(uv_center, vec2(0), uv_rotation);
 
+	vec2 matUV = base_uv;
 	vec4 albedo_ht = vec4(0.);
 	vec4 normal_rg = vec4(0.5f, 0.5f, 1.0f, 1.0f);
 	vec4 albedo_far = vec4(0.);
 	vec4 normal_far = vec4(0.5f, 0.5f, 1.0f, 1.0f);
+	float mat_scale = _texture_uv_scale_array[out_mat.base];
+	vec2 ddx1 = ddx;
+	vec2 ddy1 = ddy;
 	
 //INSERT: UNI_SCALING_BASE
 //INSERT: DUAL_SCALING_BASE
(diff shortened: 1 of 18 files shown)


project/addons/terrain_3d/editor/components/toolbar.gd
```diff
@@ -47,6 +47,7 @@ func _ready() -> void:
 	add_child(HSeparator.new())
 	add_tool_button(Terrain3DEditor.COLOR, Terrain3DEditor.REPLACE, "Paint Color", load(ICON_COLOR), tool_group)
 	add_tool_button(Terrain3DEditor.ROUGHNESS, Terrain3DEditor.REPLACE, "Paint Wetness", load(ICON_WETNESS), tool_group)
+	add_tool_button(Terrain3DEditor.UV, Terrain3DEditor.REPLACE, "UV Rotation & Scale", load(ICON_REGION_ADD), tool_group)
```
TokisanGames, line 50: Let's remove the UV tool and instead use the `enable` setting for the Paint/Spray brush to enable painting texture, so each has Enable Texture, Enable Angle, Enable Scale. Then it can operate exactly the same as the UV tool just by turning off texture.

Don't worry about the verbose labels. In another PR we'll clean up the labels so it's more concise like this:

Brush, Size, [x] Texture, [x] Angle 0-180, [x] Scale 0-10

For now, yours can look like this:
Brush, Size, [x] Enable, [x] Enable Angle Angle 0-180, [x] Enable Scale Scale 0-10

FYI in my branch, the foliage brush has a lot of UI submenus with additional options and adjustments coming.

TokisanGames: However, you could put in another option for filter by selected texture. That way it will only paint/scale/rotate if the underlying texture matches the currently selected `texture_id`. Maybe call it `[x] Selected Only`.

That's been a requested feature for the color map as well, which might be out of scope but fine if you're interested.

src/shaders/main.glsl
```diff
@@ -221,19 +221,32 @@ vec4 height_blend(vec4 a_value, float a_height, vec4 b_value, float b_height, fl
 	}
 }
 
+vec2 rota_scale_uv(vec2 uv, vec2 uv_pivot, float scale, int mat_id, float control_angle){
```
TokisanGames, line 224: Does "rota" mean anything in particular? I'd prefer to spell out rotate.

src/terrain_3d_editor.cpp
```diff
@@ -290,13 +299,17 @@ void Terrain3DEditor::_operate_map(Vector3 p_global_position, real_t p_camera_di
...
+					uint32_t uvrotation = get_uvrotation(src.r);
+					uint32_t uvscale = get_uvscale(src.r);
 					bool hole = is_hole(src.r);
 					bool navigation = is_nav(src.r);
 					bool autoshader = is_auto(src.r);
 
 					real_t alpha_clip = (brush_alpha > 0.1f) ? 1.f : 0.f;
 					uint32_t dest_id = uint32_t(Math::lerp(base_id, texture_id, alpha_clip));
 
+					std::array<uint32_t, 8> scale_align = { 5, 6, 7, 0, 1, 2, 3, 4 };
```
TokisanGames, line 311: Please write a comment as to what these numbers mean.

TokisanGames (changes requested):
First of all, thanks for making this a reality. It's mostly working well and will be fantastic once completed.

* Why 5 bits instead of 4 bits? That leaves us 3 bits instead of 4 remaining, which might be used for paintable slope. Witcher3 uses 3 bits for slope. However, that last 1-2 bits might be usable for enabling a water or a non-destructive layer. 
A clock only has 12 segments. 3 bits gives us 16. Do we really need 32?

image


* I first looked at rotating with the checkerbox pattern. It seems there are some subtle artifacts. Top at 0, Lower at 90.

Godot_v4 2 2-stable_win64_7kOkTbcyL0

Here is 180 where the circle outline is now wreathed in the artifact.

image
Other angles like 45 are strange. I would expect to see the whole pattern rotated at 45 degrees. Doing this with rock might look fine, but with cobblestone it will look like a broken pattern. 

image

If it's impossible to get what I (and users will expect) when reading "angle", we might have to change the name of angle and rotation to something that doesn't sound like a pure rotation. Also, if we're not going to get a pure rotation, then giving rotation 5 bits doesn't make any sense. Since it's going to just be rotation-like changes, 16 angles should be just as usable as 32 angles.

* Note I have some changes in `tool_settings.gd` coming this week, so one of us will need to merge with the other depending on which comes first.

https://github.com/TokisanGames/Terrain3D/pull/340/files#diff-d23e039311b38cc86aa4c8470bad9f826e1bd79cea0a4b9caf5b3c65fb56cde5

* Using the per texture uv_rotation as a detiling option is looking great for the first time ever! I've always hated that option, but now it's looking great.

project/addons/terrain_3d/editor/components/tool_settings.gd
```diff
@@ -16,6 +16,7 @@ enum SettingType {
 	COLOR_SELECT,
 	PICKER,
 	POINT_PICKER,
+	SEPERATOR,
```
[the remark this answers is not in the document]
TokisanGames: Actually, I have a better solution coming in https://github.com/TokisanGames/Terrain3D/pull/367 that will be merged soon, so you can remove all of the separator stuff.

TokisanGames:
Oh, another thing that we'll really need is adding rotation and scale pickers like we have for height/color/roughness.

src/shaders/main.glsl
Xtarsia: completley reworked, its "uv_detiling()" now.

project/addons/terrain_3d/editor/components/toolbar.gd
Xtarsia: Removed the UV tool, and built as suggested!

the filter might have to wait, but it is a good idea.

Xtarsia:
hopefully ive not missed / broken anything. some very minor artifacts remain for edge cases, but I dont think resorting to texturegrad/additional sampling is worth it.

I found 4 bits to be very noticable when trying to paint a curved path, 5bits was enough.

Picker is in too!

Xtarsia:
`weight = sqrt(weight);`

removing this improves the worse edge case artifacts quite a lot
with:
image
without:
image

smoothstep present for both (a general improvement regardless)

Xtarsia:
Final commit has no artifacts at all, implementing textureGrad with only a single call to each dFdx() and dFdy() functions in coarse mode.

overall enableing full rotation and detiling isnt free, but not too far from it.

shame that the bilinear blend wasnt enough to cover the derivative errors at mid range.

There may be some decent perf gains yet to be had, but thats out of scope for this.

src/shaders/main.glsl
```diff
@@ -229,16 +236,32 @@ void get_material(vec2 base_uv, uint control, ivec3 iuv_center, vec3 normal, out
...
-	vec2 uv1 = base_uv * _texture_uv_scale_array[out_mat.base];
-	vec2 matUV = rotate_around(uv1, uv1 - 0.5, random_angle * _texture_uv_rotation_array[out_mat.base]);
-
+	
+	// Control map scale & rotation, apply to both base and center uv.
+	// To correctly rotate & scale around each control map pixel we must 
+	// translate uv center from "control map space" to "uv space".
+	uv_center += _region_offsets[region] * _region_size;
+	// Define and apply base scale from control map value as array index. 0.5 as baseline.
+	float[8] scale_array = {0.5,0.4,0.3,0.2,0.1,0.8,0.7,0.6};
```
TokisanGames, line 245: Is this index 0-4 for positive values, index -1 to -3 for scale reduction?

Xtarsia: Yes.

TokisanGames:
Really amazing work.

* I reorganized the UI.

image

* I added a separate enabled_texture for paint/spray so it could have a unique label, until we can expand the tool settings to allow something like that.
 
* I don't think you have the clang hooks installed. When I changed the C++, clang complained and fixed code you modified:

image

* I also applied this style to your shader code, placing spaces around operators.

* Oh wow, the painted scale and angle affect the autoshader. Cool.

* We need some debug views for visualizing the scale and autoshader by color. We can do a separate PR so I can merge this. Maybe also finalizing #322 and limiting all options based on existing texture.

Issue #81, opened 2023-06-02
[TRACKER] Terrain Tools Menu UI


TokisanGames:
Now that the bakers are added to the Tools menu, we can track other tool issues here. Future tools include:

- [ ]  Import/export

- [ ] Move regions

- [ ] Image processing (de/normalize, transform heights, resize maps, swap texture ids by changing the control map pixels)

- [ ] Restructure channel packer code to remove find_child
  
- [ ]  Region viewer (heightmap / normal / color / control thumbnail view of each region)

- [ ] An information panel (# regions, total space used, map resolution, texture resolution, memory used, texture resolution, filenames, etc) (shown: GTerrain, precursor to Terrain3D)

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/34389008-37dd-4403-905a-05d07cffa767)

- [ ]  A terrain generator (shown: Zylann's HTerrain) #101 

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/1ead3c18-7d82-4a9c-93fc-9da195c4f57e)

Finished:
- [x] Arraymesh baker
- [x] Occlusion Culling baker
- [x] [Navigation baker](https://github.com/TokisanGames/Terrain3D/pull/253)
- [X] Channel packer #125 
- [x] Channel packer improvements #308
- [x]  A comprehensive tools UI panel that houses all of these tools UIs. No separate dialog boxes. #697

PR #387, opened 2024-05-26
import_images improvements

### LINKED ISSUE #130: import_images should only overwrite the maps provided
`import_images` currently expects a full set of height/color/control maps though they can be null, and it overwrites what is there. 

Change it to only overwrite what is provided. If the images are null, then don't touch current data. If the images are valid but empty that should clear current data.

Xtarsia:
Fix vertex_spacing issues, keep unmodified data.

closes #130


Xtarsia:
of course I realise after doing this pull request that perhaps the intention is for the inputed position in the importer to be in world space rather than than "map space". Tho personally I found that counter intuitive.

TokisanGames:
Yes world space. Map/region space might make more sense. This PR is to fix a bug, but we can explore improved methods for importing later. I want to add an import tool that gives some visuazliation #81 

Thanks for the fix. Looks good.

PR #403, opened 2024-06-28
Fix out-of-bounds access in Terrain3DAssets.

az-raven:
Previously, `Terrain3DAssets::set_texture` could perform an out-of-bounds access on line 373, potentially leading to a segmentation violation which would crash the editor. This notably happened when adding new textures to a resource of that type using the asset dock.

This occurred because even if `_texture_list.is_empty()` evaluates to `false`, `p_id` could still be outside the bounds of the `_texture_list` array. For example, when adding a texture to a Terrain3DAssets resource which already had items present, this code would cause a SIGSEGV on my machine, because p_id was equal to the length of the array rather than an index to an item within it.


TokisanGames:
Great, thanks!

PR #407, opened 2024-07-02
Terrain3DObjects fixes

### LINKED ISSUE #390: Moving Terrain3DObjects in tree breaks it
### Description

Add a Terrain3DObjects node, possibly as a child of Terrain3D, move it elsewhere and its children objects stop updating. No way to reconnect it except destroy and recreating it.

It should at the least have a reconnect button in the inspector, but better is to reconnect automatically. _enter_tree,  tree_entered, NOTIFICATION_PATH_RENAMED, or the current parent's NOTIFICATION_CHILD_ORDER_CHANGED might be fired during reparenting.

cc: @tcoxon

### LINKED ISSUE #391: Terrain3DObjects emits gizmo errors & buttons don't work
### Description

Selecting a Terrain3DObjects or child node keeps the Terrain3D toolbar open, which is fine.
* Selecting Terrain3D results in the first error. Then selecting another node, results in the second error.
```
 editor/plugins/node_3d_editor_gizmos.cpp:761 - Condition "valid" is true.
 editor/plugins/node_3d_editor_gizmos.cpp:782 - Condition "!valid" is true.
```
* Clicking a Terrain3D tool should reselect the Terrain3D node so it can be used. (which can be done with [plugin.select_terrain()](https://github.com/TokisanGames/Terrain3D/blob/61bebcf1148cee580fe76d586fc41226dcec1356/project/addons/terrain_3d/editor.gd#L288) from my PR #340)

cc: @tcoxon

tcoxon:
This PR fixes #390 and fixes #391.

- Gizmo errors are fixed by clearing gizmos when the Terrain3D node is deselected.
- Clicking a tool button re-selects the Terrain3D node.
- Terrain3DObjects reconnects when reparented.

The latter may look more complicated than it really is. I took this moment as an opportunity to clean up the script by making use of the new EditorInterface singleton that Godot gained in 4.2. The functional parts are:

1. Using `call_deferred` in `_on_child_entered_tree` to set up signals after Godot has updated the node's transform following reparenting.
2. Some code that moved from `_edit` to `_handles` to fix cut-and-pasting of Terrain3DObjects.

### DIFF
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -85,7 +85,7 @@ func _edit(p_object: Object) -> void:
 			terrain.storage_changed.connect(_load_storage)
 		_load_storage()
 	else:
-		terrain = null
+		_clear()
 
 	if is_instance_valid(_last_terrain) and _last_terrain.is_inside_tree():
 		if p_object is NavigationRegion3D:
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -54,11 +54,17 @@ func _exit_tree() -> void:
 func _handles(p_object: Object) -> bool:
 	if p_object is Terrain3D:
 		return true
-	if is_instance_valid(_last_terrain) and _last_terrain.is_inside_tree():
-		if p_object is NavigationRegion3D:
-			return true
-		if p_object is Terrain3DObjects or (p_object is Node3D and p_object.get_parent() is Terrain3DObjects):
-			return true
+	
+	# Terrain3DObjects requires access to EditorUndoRedoManager. The only way to make sure it
+	# always has it, is to pass it in here. _edit is NOT called if the node is cut and pasted.
+	if p_object is Terrain3DObjects:
+		p_object.editor_setup(self)
+	elif p_object is Node3D and p_object.get_parent() is Terrain3DObjects:
+		p_object.get_parent().editor_setup(self)
+	
+	if is_instance_valid(_last_terrain) and _last_terrain.is_inside_tree() and p_object is NavigationRegion3D:
+		return true
+	
 	return false
 
 
@@ -93,12 +99,7 @@ func _edit(p_object: Object) -> void:
 		else:
 			nav_region = null
 
-		if p_object is Terrain3DObjects:
-			p_object.editor_setup(self)
-		elif p_object is Node3D and p_object.get_parent() is Terrain3DObjects:
-			p_object.get_parent().editor_setup(self)
-	
-		
+
 func _make_visible(p_visible: bool, p_redraw: bool = false) -> void:
 	visible = p_visible
 	ui.set_visible(visible)
--- a/project/addons/terrain_3d/utils/terrain_3d_objects.gd
+++ b/project/addons/terrain_3d/utils/terrain_3d_objects.gd
@@ -7,15 +7,25 @@ const TransformChangedNotifier: Script = preload("res://addons/terrain_3d/utils/
 const CHILD_HELPER_NAME: StringName = &"TransformChangedSignaller"
 const CHILD_HELPER_PATH: NodePath = ^"TransformChangedSignaller"
 
-var _editor_interface = null
 var _undo_redo = null
 var _terrain_id: int
 var _offsets: Dictionary # Object ID -> Vector3(X, Y offset relative to terrain height, Z)
 var _ignore_transform_change: bool = false
 
 
+func _enter_tree() -> void:
+	if not Engine.is_editor_hint():
+		return
+	
+	for child in get_children():
+		_on_child_entered_tree(child)
+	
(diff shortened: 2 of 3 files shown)


project/addons/terrain_3d/src/ui.gd
```diff
@@ -102,6 +102,7 @@ func set_menu_visibility(p_list: Control, p_visible: bool) -> void:
 	
 
 func _on_tool_changed(p_tool: Terrain3DEditor.Tool, p_operation: Terrain3DEditor.Operation) -> void:
+	plugin.select_terrain()
```
TokisanGames, line 105: This is unnecessary now. UI is only visible if the editor plugin is visible, which only happens now if Terrain3D is selected. Before the toolbar showed up when a Terrain3DObject was selected, and this would have been useful then. However your change to editor.gd:_handles no longer makes the toolbar visible on Terrain3DObjects.

TokisanGames (changes requested):
Thanks for this update. Here's what I see when testing w/ 4.2.2.

Re: Gizmo errors
* I no longer see them when selecting

Re: Terrain3DObjects

Seems to work, but prints errors

* Terrain3D selected, add Terrain3DObjects node as a child (**I think most users are going to do this, even if by mistake, as it's a natural workflow**)
* Click player, for the location.
* Add sphere mesh
* Reparent the sphere from the player to the terrain3d objects.
It works, but I get this **message**:
```
ERROR: Node not found: "../../../Player" (relative to "/root/@EditorNode@17122/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@61/@VBoxContainer@62/@PanelContainer@109/MainScreen/@CanvasItemEditor@9461/@VSplitContainer@9280/@HSplitContainer@9282/@HSplitContainer@9284/@Control@9285/@SubViewportContainer@9286/@SubViewport@9287/Demo/Player/MeshInstance3D").
```
* Reparent Terrain3DObjects to the root node, nothing on player, and it **repeats the same message**.
* Reparent the spheremesh to the root, then back into Objects, and **the message repeats 2x**
* Delete spheremesh and make a new one attached to the root
* Reparent to Terrain3DObjects, works
* Drag to the root node and get **this message**:
```
ERROR: Node not found: "../Terrain3DObjects" (relative to "/root/@EditorNode@17122/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@61/@VBoxContainer@62/@PanelContainer@109/MainScreen/@CanvasItemEditor@9461/@VSplitContainer@9280/@HSplitContainer@9282/@HSplitContainer@9284/@Control@9285/@SubViewportContainer@9286/@SubViewport@9287/Demo/Terrain3DObjects/MeshInstance3D").
```

* At one point, I had it erroring, even if I moved the spheremesh to the root, player, root and other nodes without touching Terrain3DObjects
----
* I added some print statements and found that resetting and returning to the earlier step of moving a meshinstance from player to terrain3dobjects continually calls _on_child_transform_changed, long after the change has occurred. Indeed transform_changed_notifier is constantly emitting every frame

-----
* I made a Terrain3DObjects and a MeshInstance3D attached to the root. Moving the mesh around had no issue.
* I moved Objects as a child of Terrain3D, then moved the mesh into it and it started constantly emitting the signal.
* Moving Objects out from under Terrain3D and the signal stopped. Also it doesn't produce the error. Seems the reparent errors and signal spam occur only if Terrain3DObjects is a child of Terrain3D, which as I mentioned is likely going to be common.

tcoxon:
```
ERROR: Node not found: "../../../Player" (relative to "/root/@EditorNode@17122/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@61/@VBoxContainer@62/@PanelContainer@109/MainScreen/@CanvasItemEditor@9461/@VSplitContainer@9280/@HSplitContainer@9282/@HSplitContainer@9284/@Control@9285/@SubViewportContainer@9286/@SubViewport@9287/Demo/Player/MeshInstance3D").
```

These errors aren't anything to do with Terrain3D or Terrain3DObjects. Reparenting meshinstances can break the "skeleton" NodePath. It's an engine/editor bug, although I can't find the issue ID to reference atm. You can get the same errors if you use just a standard Node3D without the Terrain3DObjects script.

> Seems the reparent errors and signal spam occur only if Terrain3DObjects is a child of Terrain3D, which as I mentioned is likely going to be common.

I had a look. It seems like Terrain3D is setting its transform every frame, which propagates down to its child nodes. If you reparent Terrain3DObjects to anything else you can see that the transform_changed signal only emits when the node moves. I think it's a bug - I'll add a fix to my PR.

TokisanGames (approved):

PR #408, opened 2024-07-02
Fix baker script transforming nav source geometry incorrectly

### LINKED ISSUE #392: Navmesh generation gets confused when NavigationRegion3D isn't on origin
### Terrain3D version

fd5adb0

### System information

Godot v4.2.2.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.5222) - AMD Ryzen 7 3700X 8-Core Processor (16 Threads)

Not applicable

### Issue description

I was splitting my navigation up into multiple NavigationRegion3Ds, and noticed that they seemed to bake the navmesh wrong for Terrain3D, while it worked fine for normal physics objects. Each region was moved away from the origin as a node, which seemed to cause them to sample the terrain from another area.

When moving the regions back to the scene's origin, and using only the AABB filter region to set their boundaries, the generated navmesh lined up with the actual terrain.

The actual position of the NavigationRegion3D node is probably not correctly applied to the geometry that the Terrain3D provides to the navmesh baker.

(Sorry for the low-effort issue, I already got past this on my project so I'm just noting it for posterity)

### Logs

_No response_

tcoxon:
Fixes #392.

The issue was the baker de-transforming source geometry vertices after they've been generated. Source geometry is generated from the terrain heightmap (masked to the navigable areas), and used as the input to the navmesh baking process.

As well as being translated, the nav region can now be rotated/scaled and still bake correctly:

image

### DIFF
--- a/project/addons/terrain_3d/src/baker.gd
+++ b/project/addons/terrain_3d/src/baker.gd
@@ -193,7 +193,7 @@ func _bake_nav_region_nav_mesh(p_nav_region: NavigationRegion3D) -> void:
 		aabb = p_nav_region.global_transform * aabb
 		var faces: PackedVector3Array = terrain.generate_nav_mesh_source_geometry(aabb)
 		if not faces.is_empty():
-			source_geometry_data.add_faces(faces, p_nav_region.global_transform.inverse())
+			source_geometry_data.add_faces(faces, Transform3D.IDENTITY)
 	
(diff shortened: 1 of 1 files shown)


TokisanGames:
Great, works well. Thanks!

Issue #275, opened 2023-12-13
Add a way to raise height at the bottom of a cliff without also raising it at the top


tcoxon:
### Description

Currently, if you have a steep cliff in your terrain and use the raise tool to increase height at the bottom of it, the terrain at the top is also raised. This creates a sharp 'lip' at the top:

[Screencast from 2023-12-13 17-29-57.webm](https://github.com/TokisanGames/Terrain3D/assets/165720/8d0e0c4d-726c-45cb-a2b5-5be3c6b28ddc)

![image](https://github.com/TokisanGames/Terrain3D/assets/165720/339d0314-56bf-48d6-8cfa-1b30d555e0c6)

Obviously these examples are very exaggerated, but the 'lip' is noticeable in smoother terrain. The opposite also happens when the lower terrain tool is used at the top of a cliff.

Lips can be fixed by using the flatten tool, but it's a bit annoying to have to do another pass over them when making small adjustments at the bottom of a cliff.

It's not really incorrect behavior, so I've reported this as a feature request. I think my intuitive expectation would be for the raise/lower tool brushes to have a vertical extent, in addition to the extents along the X and Z axes. Not sure exactly how that would be determined though.

### Are you willing to help create this feature?

Yes

TokisanGames:
There isn't a way currently. Those steep vertical inclines are not recommended to have in your game. 
* The vertical wall stretches the vertices which stretches the texture and the collision mesh. 
* The player may be able to pop through the cliff wall when it's like that. 
* The meshes often flicked when you have perfectly vertical 90 degree meshes. Not sure why, it probably has to do with the normal calculation. 

The way I would handle this is to smooth the cliff which will condense the vertices and make the cliff wall more of an 80-85 degree angle. I'd also include bumps and height variation so it's not completely smooth. Then I'd flatten the top again if I wanted more of a sharp edge. 

3D projection can fix the texture stretching. #65 

I'm all for improving the tools as they're all clunky. If you want to add this, fine. But the other problems remain.

tcoxon:
This is a better demonstration that doesn't rely on steep cliffs:

https://github.com/TokisanGames/Terrain3D/assets/165720/f0f3b5b4-f0d0-41e9-8518-cb78b545e898

In the middle, where the two brush strokes intersect is higher than either side. Maybe this is sometimes what you want, but I often find I need a way to clip the operation to only the vertices lower than the point clicked.

[...]

Issue #123, opened 2023-07-17
Quick smooth shortcut.


ettiSurreal:
When a key is held down (Shift should be familiar to Blender users), the current tool is temporarily switched to the smooth brush, and goes back to the previous tool when let go.
Additionally, whenever https://github.com/outobugi/Terrain3D/issues/102 is implemented, holding down Shift and CTRL at the same time, it should switch to the erosion brush, as the inverse of smoothing is sharpening, and erosion would be the closest to it.

Calinou:
See my suggestion here: https://github.com/outobugi/Terrain3D/issues/122#issuecomment-1641019108

TokisanGames:
Let's consolidate this ticket into #122

Issue #122, opened 2023-07-17
Consolidate Tools & Hotkeys


ettiSurreal:
Raise and Lower are combined into one tool, with a toggle somewhere into UI to change between addition and subtraction.
Additionally, a modifier key should be implemented to temporarily switch to the opposite mode (it goes back to what it was before when it's let go). CTRL should be familiar for Blender sculpt mode users.

Chevifier:
Better Idea hold control to use the alternate tool.

TokisanGames:
In the UI, separate raise/lower buttons is the same as a combined button plus a raise/lower toggle.

I'm not opposed to defaulting to raise and CTRL for the opposite for raise/lower. The same for expand/reduce. And the same for smooth/erode.

TokisanGames:
Currently, CTRL works as negative for Holes, Navigation, Autoshader and Foliage.

@tcoxon is working on adding a lift floor / flatten peak option that will likely go with Alt in #275 .

Here's what I'm thinking we move towards:
| Positive | Negative |
|---|---|
| <kbd>Click</kbd>: Raise | <kbd>Ctrl + Click</kbd>: Lower
| <kbd>Alt + Click</kbd>: Expand | <kbd>Ctrl + Alt + Click</kbd>: Reduce
| <kbd>Shift + Alt + Click</kbd>: Lift floors | <kbd>Shift + Ctrl + Alt + Click</kbd>: Flatten peaks
| <kbd>Shift + Click</kbd>: Smooth | <kbd>Ctrl + Shift + Click</kbd>: Erode (To be implemented)

The alternate mode buttons are hidden, and are replaced when the keys are pressed.

Raise/Expand strengths need refinement. They're fine when far out, but up close they're too strong. I think their strength needs to adjust based on distance. Raise should be lessened *0.1 to match Expand, and/or adjusted based on distance to the camera.

For the rare tablet user (we have at least one), we can detect if we're on an android/ios/(touch enabled?) system, and enable modifier toggle buttons on the toolbar at the bottom of the stack to enable CTRL, SHIFT, ALT. The toolbar because they modify it, and we'll have a lot more room after hiding the negatives.

TokisanGames:
With #409 and #418, we have expanded support for CTRL to invert all operations it can:
* Regions
* Raise/lower
* Expand/reduce sculpting
* Paint color
* Paint wetness
* Autoshader
* Holes
* Navigation
* Foliage

[...]

PR #409, opened 2024-07-03
Add 'Raise Only Lower Terrain' setting to Raise Terrain tool #275

tcoxon:
This implements the feature requested in #275. It allows you to touch up the bottoms of cliffs without affecting the cliff itself, blending into the existing terrain:

https://github.com/TokisanGames/Terrain3D/assets/165720/68ffe6f9-1208-4929-b4e1-391ced71b6ce

With the setting off, the result of the operation would instead be this, where the top of the hill is pushed up, which matches Terrain3D's existing behavior:

https://github.com/TokisanGames/Terrain3D/assets/165720/5823d239-5b40-47c0-afc9-5639c2a7c0cc

I'd like to ask for feedback, especially on the naming/position of the new toggle. Would "Blend Into Existing Terrain" be a clearer name, or is that less specific? Or should we hide this behind a modifier key rather than a visible toggle?

**Limitations:**

While it works perfectly for touching up small areas, when used over a large area, it tends to produce flatter-looking hilltops than the default.

Screenshot from 2024-07-03 16-17-12

The reason is that Terrain3D brush operations overlap and this pushes up the area under the middle of the brush stroke. Whereas there isn't as much overlap when the new setting is on:

raise_only_lower

A "solution" for this would probably be pretty complex, but I'm not sure that it's necessary. What do you think?

### DIFF
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -22,11 +22,6 @@ var visible: bool
 var current_region_position: Vector2
 var mouse_global_position: Vector3 = Vector3.ZERO
 
-# Track negative input (CTRL)
-var _negative_input: bool = false
-# Track state prior to pressing CTRL: -1 not tracked, 0 false, 1 true
-var _prev_enable_state: int = -1
-
 
 func _enter_tree() -> void:
 	editor = Terrain3DEditor.new()
@@ -124,26 +119,12 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> 
 	if not is_terrain_valid():
 		return AFTER_GUI_INPUT_PASS
 	
-	## Track negative input (CTRL)
-	if p_event is InputEventKey and not p_event.echo and p_event.keycode == KEY_CTRL:
-		if p_event.is_pressed():
-			_negative_input = true
-			_prev_enable_state = int(ui.toolbar_settings.get_setting("enable"))
-			ui.toolbar_settings.set_setting("enable", false)
-		else:
-			_negative_input = false
-			ui.toolbar_settings.set_setting("enable", bool(_prev_enable_state))
-			_prev_enable_state = -1
+	ui.update_modifiers()
 	
 	## Handle mouse movement
 	if p_event is InputEventMouseMotion:
 		if Input.is_mouse_button_pressed(MOUSE_BUTTON_RIGHT):
 			return AFTER_GUI_INPUT_PASS
-
-		if _prev_enable_state >= 0 and not Input.is_key_pressed(KEY_CTRL):
-			_negative_input = false
-			ui.toolbar_settings.set_setting("enable", bool(_prev_enable_state))
-			_prev_enable_state = -1
 
 		## Setup for active camera & viewport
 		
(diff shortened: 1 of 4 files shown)


TokisanGames:
**Names:** 
  * For lifting I like: lift_floor, lift_only, (meh: lift/fill troughs/valleys)
  * For reducing gain I like: flatten_peaks, lower_peaks, (meh: reduce_peaks, reduce_gain)

**Usage:** 
I think it's too odd to use as full brush or tool setting that will be on all the time. At higher strengths it's overly aggressive and confusing to use. But when used at low strength ~3-10% it is intuitive.

Next, it should be usable on expand/reduce. So I think it will work best as a modifier, let's say ALT. I think it should decrease brush strength by *0.1 so it lives between 1-10% while in use. I hesitate to also make it a tool setting, as that will encourage people to use it (at all strengths).

I've already built some minor infrastructure in editor.gd to track CTRL for removing holes, navigation, foliage instances. What I'd like to move towards is:
* CTRL switches to the negative form for all tools. Pairs of buttons are combined. Only the positive buttons are shown. If CTRL is pressed the negative buttons are shown. #122 
* SHIFT to temporarily switch to smoothing #123. CTRL+SHIFT as erode, though it can just smooth for now.
* ALT could lift floors, while CTRL+ALT flatten peaks. Though I do like [Calinou's suggestion](https://github.com/TokisanGames/Terrain3D/issues/122#issuecomment-1641019108) of switching between Raise/Expand.
* Raise/Expand strengths need refinement. They're fine when far out, but up close they're too strong. I think their strength needs to adjust based on distance. Raise should be lessened *0.1 to match Expand. Though we don't want to double up on this *0.1, and the one above.

You could work on as much of that as you like, but I'd say at least make it a modifier, reduce strength when used, add it to expand/reduce, and remove from tool settings.

TokisanGames:
@tcoxon I saw some pushes but assume it's not ready for another review without a comment here or on discord. No rush, thanks.

tcoxon:
Ah right. It's ready for review again. Next time I'll leave a comment.

TokisanGames (approved):
Very good improvements. Thanks!

Btw, if you don't mind, could you delete your old PR branches that have been merged. I think you have a link above on this merged PR itself. But in your repo there are like 10 completed ones.

t4paN:
Sorry if this isn't the correct place to make this comment, but as a Maya navigation user in Godot, the key combo "alt+left click" for raising lower terrain completely overrides my orbit rotate functionality. I find it very hard to orbit rotate with just right click, is there a way I can switch these shortcuts?

TokisanGames:
What key is free? Link a document to the mouse and keybindings.

t4paN:
Sorry, I don't know how to find the specific keybindings, but I am using the "Maya navigation scheme" in Editor Settings > Editors > 3D > Navigation. It uses the Alt key as the orbit, pan and zoom modifiers and this gets overridden when sculpting terrain. Is there a way to just switch "alt+LMB" to "drag RMB"? I don't even need to use "raise only lower terrain" anyway, I can do this process by hand using flatten and smooth I believe.

TokisanGames:
For now comment out these lines
https://github.com/tcoxon/Terrain3D/blob/d616eec395be9fb13c773a472ffd52526e8e717a/project/addons/terrain_3d/src/ui.gd#L403-L408

Follow #441

t4paN:
Currently using terrain3d-0.9.2-beta, the ui.gd script in that folder ends at line 386. No worries, I followed #441 and can work around this issue until a GUI solution becomes available. Thank you for taking the time to look into this!

PR #411, opened 2024-07-08
Implement const correctness & pass all objects by reference

### LINKED ISSUE #410: Clean up Pass by reference & const
### Description

Some functions pass various Godot objects by reference, others not. Here's a sizeof comparison, showing which objects are worth passing by reference. By reference passes a uint64_t 8 byte pointer. 

Type, sizeof(type)
* Transform3D: 48
* AABB: 24
* Variant: 24
* Color: 16
* Vector3: 12
* Vector2, Vector2i: 8
* Ref, String, Array, Dict, RID: 8

There are also many uses of `const Ref<Terrain3DStorage> &` in parameters, when the ref is already a reference pointer.

Tasks:
* Standardize on everything >4 bytes passing by const & for 32-bit systems
* Add const to parameters where missing and appropriate

Started on a PR but then accidentally pushed it into main, so WIP at 24008a6137c38cd00156f3efbda9626413eae5a9

TokisanGames:
Fixes #410

Implements:
* const correctness on function declarations
* Pass everything > 4 bytes by reference: arrays, RIDs, Vector3, Transform3D, String, etc.

Reviewers, here's what to audit:
1. Functions that guarantee they won't change the class state are marked const (typically getters). Static & standalone functions like in Terrain3DUtil cannot be marked const like this.
```C++
-	RID get_albedo_array_rid() { return _generated_albedo_textures.get_rid(); }
+	RID get_albedo_array_rid() const { return _generated_albedo_textures.get_rid(); }
```

2. Every function parameter is const, unless the function intentionally modifies it (eg Terrain3D::_generate_triangles)
```C++
-void Terrain3DInstancer::_destroy_mmi_by_region_id(int p_region_id, int p_mesh_id) {
+void Terrain3DInstancer::_destroy_mmi_by_region_id(const int p_region_id, const int p_mesh_id)
```

3. All objects are now passed by reference or by pointer
```C++
-	void _operate_region(Vector3 p_global_position);
+	void _operate_region(const Vector3 &p_global_position);
```

4. Local function variables have been added where the function previously modified the parameter

```C++
Terrain3DEditor::_rotate_uv
-	p_uv = (p_uv - rotation_offset).rotated(p_angle) + rotation_offset;
+	Vector2 uv = (p_uv - rotation_offset).rotated(p_angle) + rotation_offset;
```

5. Ignore functions in `protected:` blocks, which derive from the engine definitions.

Also I jumped the gun on committing, so 24008a6137c38cd00156f3efbda9626413eae5a9 should be reviewed as well. Alternatively you can look at [this link](https://github.com/TokisanGames/Terrain3D/compare/c23683f577ac68db3177733c3aea7f9183491942...7bedc57a3875888074e23e71de15e9ae8fd495b9) to join that commit with the ones in this PR.

Todo: 
* update_multimesh changes can be more optimal

### DIFF
--- a/src/terrain_3d_assets.cpp
+++ b/src/terrain_3d_assets.cpp
@@ -13,7 +13,7 @@
 // Private Functions
 ///////////////////////////
 
-void Terrain3DAssets::_swap_ids(AssetType p_type, int p_src_id, int p_dst_id) {
+void Terrain3DAssets::_swap_ids(const AssetType p_type, const int p_src_id, const int p_dst_id) {
 	LOG(INFO, "Swapping asset id: ", p_src_id, " and id: ", p_dst_id);
 	Array list;
 	switch (p_type) {
@@ -32,17 +32,17 @@ void Terrain3DAssets::_swap_ids(AssetType p_type, int p_src_id, int p_dst_id) {
 		return;
 	}
 	Ref<Terrain3DAssetResource> res_a = list[p_src_id];
-	p_dst_id = CLAMP(p_dst_id, 0, list.size() - 1);
-	if (p_dst_id == p_src_id) {
+	int dst_id = CLAMP(p_dst_id, 0, list.size() - 1);
+	if (dst_id == p_src_id) {
 		// Res_a new id was likely out of range, reset it
 		res_a->_id = p_src_id;
 		return;
 	}
 
-	Ref<Terrain3DAssetResource> res_b = list[p_dst_id];
-	res_a->_id = p_dst_id;
+	Ref<Terrain3DAssetResource> res_b = list[dst_id];
+	res_a->_id = dst_id;
 	res_b->_id = p_src_id;
-	list[p_dst_id] = res_a;
+	list[dst_id] = res_a;
 	list[p_src_id] = res_b;
 
 	switch (p_type) {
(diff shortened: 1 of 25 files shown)


src/terrain_3d_assets.cpp
```diff
@@ -432,7 +432,7 @@ void Terrain3DAssets::set_mesh_list(const TypedArray<Terrain3DMeshAsset> &p_mesh
 
 // p_id = -1 for all meshes
 // Adapted from godot\editor\plugins\editor_preview_plugins.cpp:EditorMeshPreviewPlugin
-void Terrain3DAssets::create_mesh_thumbnails(int p_id, Vector2i p_size) const {
+void Terrain3DAssets::create_mesh_thumbnails(int p_id, const Vector2i &p_size) const {
```
TokisanGames, line 435: p_id

src/terrain_3d_assets.h
```diff
@@ -66,19 +66,19 @@ class Terrain3DAssets : public Resource {
...
+	PackedFloat32Array get_texture_detiles() const { return _texture_detiles; }
 	void update_texture_list();
 
 	void set_mesh_asset(int p_id, const Ref<Terrain3DMeshAsset> &p_mesh_asset);
 	Ref<Terrain3DMeshAsset> get_mesh_asset(int p_id) const;
 	void set_mesh_list(const TypedArray<Terrain3DMeshAsset> &p_mesh_list);
 	TypedArray<Terrain3DMeshAsset> get_mesh_list() const { return _mesh_list; }
 	int get_mesh_count() const { return _mesh_list.size(); }
-	void create_mesh_thumbnails(int p_id = -1, Vector2i p_size = Vector2i(128, 128)) const;
+	void create_mesh_thumbnails(int p_id = -1, const Vector2i &p_size = Vector2i(128, 128)) const;
```
TokisanGames, line 81: p_id

PR #418, opened 2024-07-12
Reduce buttons, improve CTRL operations & docs

TokisanGames:
This PR builds on #409 and expands support for CTRL to invert operations. It now works on:
* Regions
* Raise/lower
* Expand/reduce sculpting
* Paint color
* Paint wetness
* Autoshader
* Holes
* Navigation
* Foliage

Next, negative buttons are hidden (lower, reduce, remove region) and they appear when holding CTRL

Systems w/ touch displays should have an `Invert` setting on the toolbar appear to take the place of CTRL

Updated docs and added a page for the UI for CTRL and ALT operations.

Found and fixed some UI bugs such as toolbars open on NavigationRegion3D. Now toolbar is shown only when Terrain3D is selected, Menu shows then and on NavigationRegion3D. The only caveat is when a scene is opened and the user clicks NavigationRegion3D it won't show the menu until Terrain3D is clicked first, then it will.


Saul2022:
Testing on mobile and most option are working pretty well as far as i tested( add/ remove region, raise /lower, expand/divide,  Autoshader, holes, navigation  and foliage ) And it includes undo/redo so even better 😄. Tho i would say for mobile instead of saying remove lock , it could be called invert , as it does the opposite..

As far as color and wetness it not possible to test because they don't appear at the  toolbar of the bottom 

Screenshot_20240712_170425
Screenshot_20240712_170420


bout the terrain not working that well in sculpting, it caused by 2 issues

1 The performance when sculpting  as it cpu and no partial texture is way worst than on a igpu pc ( different architectures and conditions) gpu paint will prob fix this for both as an qualcom 740 is like way faster.

2 The process is being interrupted by godot  held finger rotation   , as it  seems that when you hold  press the screen it enables the camera rotation, avoiding the process that's  been taking place as it takes  more priority from  the editor , i think for that it might be good to talk with m4gr3d as i think he 's the one that made the editor for android or simply dissabling that when sculpting( wjile still being able torotate with the right joystick or the gimble thing( idk it name, the thing at the top right of the screen. )

TokisanGames:
Thanks for testing.Color erases by painting white, wetness with 0.5.

Saul2022:
> Thanks for testing.Color erases by painting white, wetness with 0.5.

Nice, but why not just make them like the other options with remove lock?

Saul2022:
> Thanks for testing.Color erases by painting white, wetness with 0.5.

Color works well, though wetness it doesnt seem to want to mark on 0.5, i tried applying it on roughness , but not working, only 1% or 0.

TokisanGames:
Sorry, Roughness at 0% (0.5 value internally) clears it. Set to -30% or so to see wet, 0% to reset.

Saul2022:
> Sorry, Roughness at 0% (0.5 value internally) clears it. Set to -30% or so to see wet, 0% to reset.

Fine, it works, well done!

TokisanGames:
Changed remove lock to invert
Added a bunch of documentation and fixed several UI bugs.
Touch screen isn't something I intend to fully support or spend much time on, at least at this time. This was just a way to not break it and get the option out of the cramped settings for all users.

PR #420, opened 2024-07-15
Fixes a 'parse error' in Terrain3DObjects that occurs when running an exported build

tcoxon:
Even though the script doesn't do anything at runtime, it still exists, and its use of the EditorInterface class causes a load failure in Godot builds that don't have editor classes.

https://github.com/godotengine/godot/issues/91713

The fix accesses it via `Engine.get_singleton(&"EditorInterface")` instead. Since this never runs except in the editor, it won't cause a compile error when EditorInterface doesn't exist.


TokisanGames:
Thanks!

PR #432, opened 2024-07-26
Expose physics RID

Ryan-000:
It's quite useful when you need to interact with the terrain from PhysicsDirectSpaceState3D.

### DIFF
--- a/src/terrain_3d.h
+++ b/src/terrain_3d.h
@@ -154,6 +154,7 @@ public:
 	uint32_t get_collision_mask() const { return _collision_mask; };
 	void set_collision_priority(const real_t p_priority);
 	real_t get_collision_priority() const { return _collision_priority; }
+	RID get_collision_rid() const;
 
 	// Terrain methods
 	void snap(const Vector3 &p_cam_pos);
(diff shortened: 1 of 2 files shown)


src/terrain_3d.h
```diff
@@ -154,6 +154,7 @@ class Terrain3D : public Node3D {
 	uint32_t get_collision_mask() const { return _collision_mask; };
 	void set_collision_priority(const real_t p_priority);
 	real_t get_collision_priority() const { return _collision_priority; }
+	RID get_rid() const { return _static_body; };
```
TokisanGames, line 157: The function call will look like `terrain->get_rid()`, but this is not the RID of Terrain3D. Call it `get_collision_rid()`.

If debug collision is enabled, and _debug_static_body != nullptr, you need to return `_debug_static_body->get_rid()` or `RID()` if nothing. 

Bear in mind #278 will redesign collision and the API might change in the future.

TokisanGames:
Thanks for the change. I have updated this for you so we can merge it in. 

You made your change on the `main` branch instead of another as instructed in the [Godot pull request workflow](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html). I had to force push into your `main` branch, which has probably broken it on your system. That means you might not be able to pull changes. If so, you'll have to delete your local `main` branch and download it again, or do another clone. In the future, always make another branch to make changes in.

PR #437, opened 2024-07-29
Hide the navigation layer in the editor when another node is selected

Dekker3D:
In the editor, the terrain would keep its purple overlay even when deselected (and all the other UI elements for the terrain were hidden). This felt unintuitive to me, because if I don't have the terrain selected, I can't draw on the navigation layer anyway, and I'm probably not trying to do so either. So after a quick discussion on Discord, I wrote up a fix for this.

The currently selected tool and operation are stored whenever they're changed, and when the UI is hidden, the tool is set to TOOL_MAX. It is restored to its previous selection when the UI is shown again.

### DIFF
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -91,6 +91,7 @@ func _edit(p_object: Object) -> void:
 		region_gizmo.set_node_3d(terrain)
 		terrain.add_gizmo(region_gizmo)
 		terrain.set_plugin(self)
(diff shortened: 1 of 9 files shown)


TokisanGames:
This works for when they deselect Terrain3D, but if they run the game, it's still enabled. That's half of people's complaints.

To hide the navigation overlay properly it should:
* In editor
  * show if terrain & navigation selected
  * show if navigation debug view is enabled
* In game
  * show if navigation debug view is enabled

I have extended this PR to address these.

PR #444, opened 2024-08-03
Improve brush defaults and settings, Add SHIFT f/ smooth

### LINKED ISSUE #122: Consolidate Tools & Hotkeys

TokisanGames:
Makes the following changes:

**General**
* Brush size default 50m to 20m
* Brush strength default 10% to 33%

**Raise/Lower**
* Raise/Lower brush reduced to 1/10th strength
* ~~Expand only goes up (instead of down below zero), Reduce only goes down (instead of up below zero)~~
* Expand/reduce removed. They are multiplicative, so at 0 do nothing, at large heights are super strong. Raise/lower is additive and the same regardless of heights. There's now little difference from Raise/lower, and it's an inconsistent brush.
* Lift/flatten mode via ALT now has colored decals

**Smooth**
* Ordered Smooth above Slope/Height buttons
* Smooth enabled on SHIFT
* Smooth brush 2x stronger, lower cap added (maximum is not stronger)

**Height**
* Flatten brush renamed to Height brush
* Height brush default height 50m to 20m
* Height / Slope strength 1/2th

**Texture/Rough**
* Spray brush strength 2x, upper clamp 10x
* Wetness defaults to -65%

Fixes #122

### DIFF
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -130,7 +130,8 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> 
 	if not is_terrain_valid():
 		return AFTER_GUI_INPUT_PASS
 	
-	ui.update_modifiers()
+	if p_event is InputEventKey:
+		ui.update_modifiers()
 	
 	## Handle mouse movement
 	if p_event is InputEventMouseMotion:
--- a/project/addons/terrain_3d/src/tool_settings.gd
+++ b/project/addons/terrain_3d/src/tool_settings.gd
@@ -57,10 +57,10 @@ func _ready() -> void:
 	add_setting({ "name":"instructions", "label":"Click the terrain to add a region. CTRL+Click to remove. Or select another tool on the left.",
 		"type":SettingType.LABEL, "list":main_list, "flags":NO_LABEL })
 
-	add_setting({ "name":"size", "type":SettingType.SLIDER, "list":main_list, "default":50, "unit":"m",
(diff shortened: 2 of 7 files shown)


TokisanGames:
Shift now temporarily switches to smooth. Also adjusted decal opacity and slope/height strength so they are more consistent between tools.

Saul2022:
Not able to test pr , but considering this might enable automically smooth when holding shift, it could be fine to remove the icon for less cluttering in the toolbar. For mobile just make the smooth button on raise/lower or in height.

TokisanGames:
Hmm, blender has a smooth tool, in addition to the hot key, even on desktop. Maybe for now I'll leave it. At least until everyone is familiar with the shortcuts.

TokisanGames:
@Ryan-000 are you using the current main branch or this PR build? Lower is now hidden. Raise/lower are one tool, toggled by ctrl. Reduce/expand so the same, but are going away today.

TokisanGames:
I've now removed expand/reduce, and the editor settings pushed to another PR. 
This is ready to merge.

> I'm just talking about the main branch (I'm currently using 0.9.2-dev), I haven't tried the PR build or the latest release yet.

There is a nightly build from the main branch, 0.9.3-dev which has Raise/Lower already combined.
There is 0.9.2 release which has them separate.

PR #447, opened 2024-08-04
Allow region size changes

### LINKED ISSUE #77: Support more region sizes
Edit:
Changing region_size broke existing data. And no other size even worked, so I disabled them all except 1024 in https://github.com/outobugi/Terrain3D/commit/ebf9df002d4dfe44165191f57b7b0b1899af8bd4.

Now let's get all the other sizes working again and look at much larger spaces like up to 2048 regions (45 square) and up to 2k or 4k region sizes. See https://github.com/outobugi/Terrain3D/issues/77#issuecomment-1641554083

-------
Changing region size should not change the data at all. It should only change the size of the regions, represented as white boxes in the images below.

Instead it gives all sorts of artifacts:
* The data on the region map loses order, some sections appear split and move around
* Some data gets scaled laterally, some is flattened vertically
* Terrain normals are not recalculated afterwards
* The total size changes, rather than only the regions changing with the same data. 

It may be as simple as running some update functions after changing region size.

4k x 2k @ 1024 region size

image

Change to 2048 region size. The total size is 2x as large. Instead only the white boxes should be bigger. That "mountainous" section in the top left between the split peaks is actually flat. But the normals make it look like there are heights there.

image

Changing to 512 region size and everything got smaller with less resolution. The data should remain the same, and the only difference we should see is the white boxes getting smaller.

image

### LINKED ISSUE #487: Extra regions after setting up navigation
### Description

Setup navigation then save. In the demo it reports the following. Only the first three regions are correct.
```
Terrain3DNavigation: Finished baking 1 NavigationMesh.
Terrain3DRegion#2026:save: Writing region (0, -1) to res://demo/data/terrain3d_00-01.res
Terrain3DRegion#1406:save: Writing region (0, -2) to res://demo/data/terrain3d_00-02.res
Terrain3DRegion#0786:save: Writing region (0, 0) to res://demo/data/terrain3d_00_00.res
ERROR: Terrain3DData#8583:save_region: No region found at: (1, -1)
   at: push_error (core/variant/variant_utility.cpp:1091)
ERROR: Terrain3DData#8583:save_region: No region found at: (1, 0)
   at: push_error (core/variant/variant_utility.cpp:1091)
ERROR: Terrain3DData#8583:save_region: No region found at: (1, -2)
   at: push_error (core/variant/variant_utility.cpp:1091)
ERROR: Terrain3DData#8583:save_region: No region found at: (0, 1)
   at: push_error (core/variant/variant_utility.cpp:1091)
ERROR: Terrain3DData#8583:save_region: No region found at: (1, 1)
   at: push_error (core/variant/variant_utility.cpp:1091)
```

Dekker3D:
# Update by TokisanGames

Fixes #77
Fixes #487 

- [x] Enable changing region sizes, reslicing data (by Dekker3D)
- [x] Change do_for_regions to work with Callables for all language support
- [x] Region size is set automatically from the first loaded file
- [x] Add a UI for changing region size
- [x] Bug: occasionally reports regions are missing upon saving. They are null. See https://github.com/TokisanGames/Terrain3D/pull/447#issuecomment-2353416445
- [x] Bug: Some sequences produce regions that are not blank but have data copied from other regions See https://github.com/TokisanGames/Terrain3D/pull/447#issuecomment-2350957671
- [x] Bug: Heights and AABBs are messed up after slicing.
- [x] Bug: Occasionally prints a get_dependencies error See https://github.com/TokisanGames/Terrain3D/pull/447#issuecomment-2350957671
- [x] Bug: Cannot go from 2 adjacent squares to the next size higher or that square disappears. Must have kitty-corner squares to fill in the blanks.
- [x] Foliage needs to be sliced and moved as well. Currently it is deleted.
- [x] Bug: Cannot import upgrade files then change region size or terrain disappears. Must save first, then reload the scene before region size can be changed. 

----
# Original PR

This should fix #77, and I haven't been able to spot any remaining issues regarding different region sizes. Regions are resized when you change the size, so if you have smaller regions that don't fully fill up a larger region when you change to a larger size, the other parts of that region will get default values.

The code for resizing is a little ugly (much uglier when sizing up, rather than down) and probably not very well optimized, but it works, and people are unlikely to change region sizes so often that it's a show-stopper.

### DIFF
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -68,6 +68,100 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
 	return region_arr;
 }
 
+void Terrain3DData::do_for_regions(Rect2i p_bounds, std::function<void(Terrain3DRegion *, Rect2i, Rect2i, Rect2i)> p_callback, bool p_do_empty_regions) {
+	Rect2i index_bounds;
+	index_bounds.position = Point2i((Point2(p_bounds.position) / _region_size).floor()); // rounded down
+	index_bounds.set_end(Point2i((Point2(p_bounds.get_end()) / _region_size).ceil())); // rounded up in a dumb way
+	Point2i current_region_loc;
+	for (int y = index_bounds.position.y; y < index_bounds.get_end().y; y++) {
+		current_region_loc.y = y;
+		for (int x = index_bounds.position.x; x < index_bounds.get_end().x; x++) {
+			current_region_loc.x = x;
+			Ref<Terrain3DRegion> region = get_region(current_region_loc);
+			if (region.is_valid() || p_do_empty_regions) {
+				Rect2i world_bounds = p_bounds.intersection(Rect2i(current_region_loc * _region_size, _region_sizev));
+				Rect2i area_bounds(world_bounds.position - p_bounds.position, world_bounds.size);
+				Rect2i region_local_bounds(world_bounds.position - (region->get_location() * _region_sizev), area_bounds.size);
+				p_callback(region.ptr(), world_bounds, area_bounds, region_local_bounds);
+			}
+		}
+	}
+}
+
+void Terrain3DData::set_region_size(int p_new_region_size) {
+	if (p_new_region_size == _region_size)
+		return;
+
+	// List new regions that we need
+	std::set<Point2i> new_region_points;
+	std::list<Ref<Terrain3DRegion>> old_regions;
+
+	Array locs = _regions.keys();
+	int region_id = 0;
+	for (int i = 0; i < locs.size(); i++) {
+		Ref<Terrain3DRegion> region = _regions[locs[i]];
+		if (region.is_valid() && !region->is_deleted()) {
+			Rect2i index_bounds;
+			Point2i region_position = region->get_location() * _region_size;
+			index_bounds.position = region_position / p_new_region_size;
+			index_bounds.set_end(Point2i((Point2(region_position + _region_sizev) / p_new_region_size).ceil()));
+
+			for (int y = index_bounds.position.y; y < index_bounds.get_end().y; y++) {
+				for (int x = index_bounds.position.x; x < index_bounds.get_end().x; x++) {
+					new_region_points.insert(Point2i(x, y));
+				}
+			}
+			old_regions.push_back(region);
+		}
+	}
+
+	// Make new regions
+	std::list<Ref<Terrain3DRegion>> new_regions;
+	for (Point2i p : new_region_points) {
+		Ref<Terrain3DRegion> region;
+		region.instantiate();
+		region->set_location(p);
+		region->set_region_size(p_new_region_size);
+		region->set_modified(true);
+		region->sanitize_maps();
+
+		// Fill data
+		Rect2i bounds;
+		bounds.position = p * p_new_region_size;
+		bounds.size = Vector2i(p_new_region_size, p_new_region_size);
+
+		Ref<Image> height = region->get_map(Terrain3DRegion::TYPE_HEIGHT);
+		Ref<Image> control = region->get_map(Terrain3DRegion::TYPE_CONTROL);
+		Ref<Image> color = region->get_map(Terrain3DRegion::TYPE_COLOR);
+
+		do_for_regions(
+				bounds, [&height, &control, &color](Terrain3DRegion *region, Rect2i world_bounds, Rect2i area_bounds, Rect2i region_local_bounds) {
+					height->blit_rect(region->get_height_map(), region_local_bounds, area_bounds.position);
+					control->blit_rect(region->get_control_map(), region_local_bounds, area_bounds.position);
+					color->blit_rect(region->get_color_map(), region_local_bounds, area_bounds.position);
+				},
+				false);
+
+		height->clear_mipmaps();
+		control->clear_mipmaps();
+		color->clear_mipmaps();
+		new_regions.push_back(region);
+	}
+
+	for (Ref<Terrain3DRegion> r : old_regions) {
+		remove_region(r, false);
+	}
+
+	_region_size = p_new_region_size;
+	_region_sizev = Vector2i(p_new_region_size, p_new_region_size);
+
+	for (Ref<Terrain3DRegion> r : new_regions) {
+		add_region(r, false);
+	}
+
+	force_update_maps();
+}
+
 void Terrain3DData::set_region_modified(const Vector2i &p_region_loc, const bool p_modified) {
 	Ref<Terrain3DRegion> region = _regions[p_region_loc];
 	if (region.is_null()) {
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -73,10 +73,8 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
 // The callable receives: source Terrain3DRegion, source Rect2i, dest Rect2i, (bindings)
 // Used with change_region_size, dest Terrain3DRegion is bound as the 4th parameter
 void Terrain3DData::do_for_regions(const Rect2i &p_area, const Callable &p_callback) {
-	Rect2i location_bounds;
-	location_bounds.position = Point2i((Point2(p_area.position) / real_t(_region_size)).floor()); // rounded down
-	location_bounds.set_end(Point2i((Point2(p_area.get_end()) / real_t(_region_size)).ceil())); // rounded up in a dumb way
-	LOG(INFO, "Processing area: ", p_area, " -> ", location_bounds);
+	Rect2i location_bounds(V2I_DIVIDE_FLOOR(p_area.position, _region_size), V2I_DIVIDE_CEIL(p_area.size, _region_size));
+	LOG(DEBUG, "Processing global area: ", p_area, " -> ", location_bounds);
 	Point2i current_region_loc;
 	for (int y = location_bounds.position.y; y < location_bounds.get_end().y; y++) {
 		current_region_loc.y = y;
@@ -111,14 +109,11 @@ void Terrain3DData::change_region_size(int p_new_size) {
 	// Get current region corners expressed in new region_size coordinates
 	Dictionary new_region_points;
 	Array locs = _regions.keys();
-	int region_id = 0;
 	for (int i = 0; i < locs.size(); i++) {
 		Ref<Terrain3DRegion> region = get_region(locs[i]);
 		if (region.is_valid() && !region->is_deleted()) {
 			Point2i region_position = region->get_location() * _region_size;
-			Rect2i location_bounds;
-			location_bounds.position = region_position / p_new_size;
-			location_bounds.set_end(Point2i((Point2(region_position + _region_sizev) / p_new_size).ceil()));
+			Rect2i location_bounds(V2I_DIVIDE_FLOOR(region_position, p_new_size), V2I_DIVIDE_CEIL(_region_sizev, p_new_size));
 			for (int y = location_bounds.position.y; y < location_bounds.get_end().y; y++) {
 				for (int x = location_bounds.position.x; x < location_bounds.get_end().x; x++) {
 					new_region_points[Point2i(x, y)] = 1;
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -68,25 +68,30 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
 	return region_arr;
 }
 
-void Terrain3DData::do_for_regions(Rect2i p_bounds, std::function<void(Terrain3DRegion *, Rect2i, Rect2i, Rect2i)> p_callback, bool p_do_empty_regions) {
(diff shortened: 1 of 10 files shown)


TokisanGames:
In my new PR, Terrain3DStorage has been renamed to Terrain3DData and changed from a savable resource to an Object. It won't save any data itself. All of the saved data are in Terrain3DRegions or the scene file. I'm moving the (global) region_size setting to Terrain3D. 

This PR reslices when changing the setting. Fine with a few regions, but dangerous with 256-2048 regions.

What about undo/redo? If it's an inspector setting, Godot will undo/redo that setting which will reslice each time. Also not a great user experience.

The regions track their own region size based on the size of the first map loaded and validate subsequent loads against it. What if you're moving region files from one project to another? You need to set the right global region size setting before loading them or you'll get weird/broken results. Maybe when loading regions, it automatically sets global region size based on the first region and complains about any regions that are different. Once loaded the user can manually reslice if desired.

I'm thinking about rather than giving them an enum to choose and reslicing all data each time they change it, instead:
* Region_size is read-only in the inspector
* Add a menu option where they can resize regions
* Add an API function to resize regions, called by the above. Then users will expect a process. We won't need an undo, they can just reslice.

In my PR all changes are done in memory (add/remove regions, sculpting, painting) and are not saved to disk until the scene is saved. Reslicing should work the same. Looks like this PR also does that.

Looks like this is a minor addition to what is already here, but my new region structure will require more updating.

TokisanGames:
My PRs are merged. Please rebase and update this so we can merge it in.

src/terrain_3d_data.h
```diff
@@ -92,6 +93,8 @@ class Terrain3DData : public Object {
 	Dictionary get_regions_all() const { return _regions; }
 	PackedInt32Array get_region_map() const { return _region_map; }
 	static int get_region_map_index(const Vector2i &p_region_loc);
+	void do_for_regions(Rect2i bounds, std::function<void(Terrain3DRegion *, Rect2i, Rect2i, Rect2i)> callback, bool do_empty_regions);
```
TokisanGames, line 96: If std::function were a Callable or a CallableCustom instead, this function would work for gdextension and all languages. And you can still get a native function pointer to it. Why not do that so it's universally accessible?

src/terrain_3d_data.cpp
```diff
@@ -68,6 +68,100 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
...
+
+		height->clear_mipmaps();
+		control->clear_mipmaps();
+		color->clear_mipmaps();
+		new_regions.push_back(region);
+	}
+
+	for (Ref<Terrain3DRegion> r : old_regions) {
+		remove_region(r, false);
+	}
```
TokisanGames, line 153: remove_region doesn't remove from _regions, it only marks those regions as deleted.
When add_region comes in, it is adding regions in a populated dictionary, but the indexed keys are two different scales.
Saving after this operation reports missing regions because _regions has been corrupted.
Rather than remove_region, _regions should just be erased, and new regions added. The region map will be rebuilt in update_maps.

src/terrain_3d_data.h
Dekker3D: I considered that, but couldn't find out what Godot's Callable looks like in C++. Didn't search long, I'll admit. Was focused on just getting this to work in time. It does sound like a good idea, assuming it doesn't lose too much performance (for other uses of this function).

TokisanGames: You have to look at the Godot source for most code. Godot-cpp is only a wrapper.
Look at callable_method_pointer.h.

callable_mp can't make a function pointer out of C++ lambdas currently; only gdscript lambdas. So we can either keep it only working with C++ lambdas. Or change the lambda to an inline function so we can use Callables and expose it for all languages. I'm leaning towards the latter.

src/terrain_3d_data.cpp
```diff
@@ -68,6 +68,100 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
...
 
+void Terrain3DData::do_for_regions(Rect2i p_bounds, std::function<void(Terrain3DRegion *, Rect2i, Rect2i, Rect2i)> p_callback, bool p_do_empty_regions) {
+	Rect2i index_bounds;
+	index_bounds.position = Point2i((Point2(p_bounds.position) / _region_size).floor()); // rounded down
+	index_bounds.set_end(Point2i((Point2(p_bounds.get_end()) / _region_size).ceil())); // rounded up in a dumb way
+	Point2i current_region_loc;
+	for (int y = index_bounds.position.y; y < index_bounds.get_end().y; y++) {
+		current_region_loc.y = y;
+		for (int x = index_bounds.position.x; x < index_bounds.get_end().x; x++) {
+			current_region_loc.x = x;
```
TokisanGames, line 79: All of this receives integer points in p_bounds, then converts to float to get the region location. Then it iterates through the included region locations. There's no reason or benefit to limit input to integers, or even calculate here. Instead, it can receive real world global coordinates and use get_region_location() to convert to location, including descaling.

Also below will crash on line 84 (region->get...) if region is null and p_do_empty_regions is true.

src/terrain_3d_data.cpp
```diff
@@ -68,6 +68,100 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
...
+		Rect2i bounds;
+		bounds.position = p * p_new_region_size;
+		bounds.size = Vector2i(p_new_region_size, p_new_region_size);
+
+		Ref<Image> height = region->get_map(Terrain3DRegion::TYPE_HEIGHT);
+		Ref<Image> control = region->get_map(Terrain3DRegion::TYPE_CONTROL);
+		Ref<Image> color = region->get_map(Terrain3DRegion::TYPE_COLOR);
+
+		do_for_regions(
+				bounds, [&height, &control, &color](Terrain3DRegion *region, Rect2i world_bounds, Rect2i area_bounds, Rect2i region_local_bounds) {
```
TokisanGames, line 138: world_bounds isn't used here.

src/terrain_3d_data.cpp
```diff
@@ -68,6 +68,100 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
...
+					height->blit_rect(region->get_height_map(), region_local_bounds, area_bounds.position);
+					control->blit_rect(region->get_control_map(), region_local_bounds, area_bounds.position);
+					color->blit_rect(region->get_color_map(), region_local_bounds, area_bounds.position);
+				},
+				false);
+
+		height->clear_mipmaps();
+		control->clear_mipmaps();
+		color->clear_mipmaps();
```
TokisanGames, line 147: height and control don't use mipmaps. Only color. and you can force it to regenerate by adding a parameter to force_update_maps.

TokisanGames:
I think dd729a42ba07181a888aff22e145decec1a6b2a9 is much more versatile, allowing us to provide copy/paste functionality to the editor in the future, and exposing do_for_regions to all languages. Now we can do this in gdscript.

```gdscript
func _ready() -> void:
	print("Readying do for regions")
	data.do_for_regions(Rect2i(0,-3072, 2048, 2048), 
		func(region:Terrain3DRegion, sarea:Rect2i, darea:Rect2i): 
			print("GDScript: ", region, sarea, darea)
	)

GDScript: [Wrapped:0][P: (0, 1024), S: (1024, 1024)][P: (0, 0), S: (1024, 1024)]
GDScript: [Wrapped:0][P: (1024, 1024), S: (1024, 1024)][P: (0, 0), S: (1024, 1024)]
```

src/terrain_3d_data.cpp
TokisanGames: removed

src/terrain_3d_data.cpp
TokisanGames: removed

Dekker3D: I don't see a commit where you removed it, but world_bounds seems useful for other purposes, and the point of do_for_regions is to be versatile.

src/terrain_3d_data.cpp
Dekker3D: Ah, I didn't test at all with p_do_empty_regions set to true. It's just a thing that LayerProcGen supports and it seemed sensible to do the same here. Like when you want to fill an area with a default colour, if no terrain exists there.

The main reason this version is limited to integers, is because it'll slow everything down by a lot if you need to consider interpolation. For one, interpolation can require you to use more than one region at the same time. I haven't really come up with a clean way to do that yet.

src/terrain_3d_data.cpp
TokisanGames: I mean your lambda, not do_for_regions, received a world_bounds parameter, then did nothing with it. Since world bounds can or should be accessible through other means, and the function here doesn't use it, it can be cut. The other parameters are not accessible anywhere else so are appropriate. 

Removed in dd729a4

src/terrain_3d_data.cpp
TokisanGames: I'm going to drop do_empty_regions. You're calling the region and have nothing to call if it's empty. We can add more functionality later based on need. Perhaps it can add empty regions, then call them, but right now there's no use for it.

TokisanGames:
This is working well with Callable and reslicing is very fast.

However, there are three outstanding bugs that I haven't tracked down yet. They are present in your original and my current. My note about making a new _regions database might be accurate and help these. This is what I'm working on next.

1. Sometimes when increasing/decreasing/adding/removing regions then saving, it reports that several regions cannot be found. In the code it means `_regions[location].is_null() == true`. These locations are ones not visible/present. I haven't found a reliable sequence that triggers this, but it only occurs on saving. We could just remove the null entries, but I'd like to know why.

2. Starting at the 3x1024 demo data, change to 512. Delete four regions in a corner (say +x, +z). Change to 2048. One region already has data.
    - Alternatively start at 1024, change to 512, change to 1024, change to 2048. Here the same corner region looks like above, but 4 other sections have extra data.

{6BDBB2CD-20F5-4B38-A946-AA053C8AE5C5}

3. Sometimes when saving it prints "Writing region (0, 0) to res://demo/data/terrain3d_00_00.res", then prints after all saving:
```
ERROR: Method/function failed.
at: get_dependencies (scene/resources/resource_format_text.cpp:1771)
```
This is the engine source
https://github.com/godotengine/godot/blob/4.2/scene/resources/resource_format_text.cpp#L1771

It might be attempting to `get_path()` on an invalid region or something.

TokisanGames:
Exposing copy_paste allows this to work in gdscript. It currently copies the contents of region maps from 0 to region_size. In the future this should be expanded to work with a rect2i in global coordinates, then loop through regions within the area
```
	data.copy_paste(data.get_region(Vector2i(0,-2)), Rect2i(0,0,512,512),  data.get_region(Vector2i(0,0)), Vector2i(0,0))
	data.force_update_maps()
```

TokisanGames:
Alright, I found a sequence that causes `No region found errors` in both the original PR and my changes.
1. Load the default demo data, 3 x 1024.
region_locations: [(0, -1), (0, 0), (0, -2)]
_regions: { (0, -1): [Wrapped:0], (0, 0): [Wrapped:0], (0, -2): [Wrapped:0] }

2. Convert to 2048
region_locations: [(0, -1), (0, 0)]
_regions: { (0, -1): [Wrapped:0], (0, 0): [Wrapped:0], (0, -2): [Wrapped:0], (1, -2): <null>, (1, -1): <null>, (1, 0): <null>, (0, 1): <null>, (1, 1): <null> }

3. Save
Reports
```
ERROR: Terrain3DData#9529:save_region: No region found at: (1, -2)
ERROR: Terrain3DData#9529:save_region: No region found at: (1, -1)
ERROR: Terrain3DData#9529:save_region: No region found at: (1, 0)
ERROR: Terrain3DData#9529:save_region: No region found at: (0, 1)
ERROR: Terrain3DData#9529:save_region: No region found at: (1, 1)
   at: push_error (core/variant/variant_utility.cpp:1091)
```
region_locations: [(0, -1), (0, 0)]
_regions: { (0, -1): [Wrapped:0], (0, 0): [Wrapped:0], (1, -2): <null>, (1, -1): <null>, (1, 0): <null>, (0, 1): <null>, (1, 1): <null> }

Something is inserting ghost regions into _regions.

TokisanGames:
Attempting to put this in Out of the Ashes, I realized this doesn't reslice any of the foliage data. It just erases it. I have updated the top post with a tracker of tasks to do.

TokisanGames:
Updates:
* Identified the get_dependencies error as a godot bug unrelated to Terrain3D.
* Fixed a bug all throughout Terrain3DData that created null regions, which are detected on save
* Fixed some sequences creating new region space with duplicate data from random regions instead of blank/flat
* Fixed issues expanding a single or 2 adjacent regions to a larger size (disappeared), and having it properly fill out to the edges of the region it is in instead of moving to another, (-1, -1 to 0, 0 moved to 0, 0 to 1, 1).
* It's now working with vertex_scaling

I'm working on moving foliage properly

TokisanGames:
This PR is ready for review and testing. 

There's just one last bug I'm tracking down: region sizes can't be changed right after upgrading. You must upgrade to the new storage, save, and reload the scene before region sizes can be changed.

PR #450, opened 2024-08-09
Fix crash when selecting after pressing Escape

### LINKED ISSUE #449: Godot crashes after starting to paint mesh then pressing Escape and selecting anything in scene
### Terrain3D version

8a45d61a331fcd8e9c6d8afbc25e278480c3460e

### System information

Godot Engine v4.3.rc3.official.03afb92ef - Windows 11 Home - Vulkan 1.3.278 - Forward+ - NVIDIA GeForce RTX 2070 with Max-Q Design

Yes

### Issue description

Tested on:
- Terrain3D
    - 0.9.2 (from AssetLib)
    - 0.9.3-dev [nightly build](https://github.com/TokisanGames/Terrain3D/actions/runs/10237771326)
- Godot
    - 4.3-rc2
    - 4.3-rc3

To reproduce:
- Create a terrain and select it
- Go into the `Terrain3D tab > Meshes`
- With the red circle decal showing in the viewport press `Escape`
    - This will throw few errors because during the `ui.gd:264 await get_tree().create_timer(.05).timeout` the `editor.gd:118 _clear` function is called which makes the `plugin.terrain` become null (for which the check was done before the await)

After this trying to select anything in the scene with the blue selection rectangle will crash Godot on mouse release.

However, if the terrain node is selected and then deseleted in the hierarchy, there is no crash.

By searching for `get_deepest_editable_node` I found [this issue](https://github.com/TokisanGames/Terrain3D/issues/395) but the fix for that was  merged into Godot in [4.3-rc2 commits](https://github.com/godotengine/godot/compare/e343dbbcc1030f04dc5833f1c19d267a17332ca9...3978628c6cc1227250fc6ed45c8d854d24c30c30)
> Fix crash when drawing a selection box on an empty scene with certain plugins
> @TokisanGames TokisanGames committed on Jun 20 1780f2840765358b52d8f9bd28a488fd3ef89842

### Logs

```bash
SCRIPT ERROR: Invalid call. Nonexistent function 'get_camera' in base 'Nil'.
    at: update_decal (res://addons/terrain_3d/src/ui.gd:269)
SCRIPT ERROR: Invalid call. Nonexistent function 'get_camera' in base 'Nil'.
    at: update_decal (res://addons/terrain_3d/src/ui.gd:269)
SCRIPT ERROR: Invalid call. Nonexistent function 'get_camera' in base 'Nil'.
    at: update_decal (res://addons/terrain_3d/src/ui.gd:269)
ERROR: Condition "!is_ancestor_of(p_start_node)" is true. Returning: p_start_node
    at: get_deepest_editable_node (scene/main/node.cpp:2525)
WARNING: Terrain3D#3844:_notification: NOTIFICATION_CRASH
    at: push_warning (core/variant/variant_utility.cpp:1112)

================================================================
Dumping the backtrace. Please include this when reporting the bug to the project developer.
[1] error(-1): no debug info in PE/COFF executable
...
[26] error(-1): no debug info in PE/COFF executable
-- END OF BACKTRACE --
================================================================
```

Radivarig:
A proper fix would be to find what part of `_on_tool_changed` undoes the state that makes the crash happen, but for now this fixes the crash and those post await errors.

### DIFF
--- a/project/addons/terrain_3d/src/ui.gd
+++ b/project/addons/terrain_3d/src/ui.gd
@@ -264,6 +264,10 @@ func update_decal() -> void:
 		await get_tree().create_timer(.05).timeout 
 		decal.visible = true
 
+	if not plugin.terrain:
+		_on_tool_changed(last_tool, last_operation)
+		return
+
 	decal.size = Vector3.ONE * brush_data["size"]
 	if brush_data["align_to_view"]:
 		var cam: Camera3D = plugin.terrain.get_camera();


TokisanGames:
Thanks for this.
The engine bug that caused the error and crash were resolved by https://github.com/godotengine/godot/pull/95420 which will close #449.

Decal visibility was reworked to remove the coroutine in 529c7e9 which resolves both issues now.

PR #451, opened 2024-08-11
Optimizations

Dekker3D:
Admin edit
* Optimizes Terrain3DData, eg. set/get_pixel(), get_height() and more by replacing creation of Refs with pointer usage
* Meshing will be optimized in or after #622 
* Checks more pointers before using
* Consolidates nullptr checks using implicit boolean conversion
* Fixes some improper usage 

Results
```
10M get_pixel(): 2836ms => 1713ms, +39.6%
1M get_height() interpolated: 1571ms => 982ms, +37.5%
bake_mesh 3x1024: 40765ms => 34750ms, +14.8%
```

-------
This PR should cut down navigation mesh baking time by about half. Terrain3D::_generate_triangle_pair uses fewer control map lookups (-6 when p_require_nav is false, -2 if true), and Terrain3DStorage::get_pixel uses a modified private version of get_map_region that returns an Image* and skips converting it to a godot::Ref, reducing the time spent on get_pixel by about 1/3rd. This is based on the assumption that nobody will put anything other than godot::Image in the height/control/color map arrays, which should be true because they're TypedArrays and they're not directly accessible from outside the extension.

The modified version of get_map_region introduces some code duplication, and it might be good to use this modified version in other places where a Ref really isn't needed, so this PR isn't as clean as it perhaps could've been. I'm personally unlikely to even use it in my project, as it should be much faster to just directly sample the physics shapes I'm making, which can also be generated in a thread, unlike Terrain3D.

### DIFF
--- a/src/terrain_3d_util.cpp
+++ b/src/terrain_3d_util.cpp
@@ -2,6 +2,7 @@
 
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/file_access.hpp>
+#include <godot_cpp/classes/time.hpp>
 
 #include "logger.h"
 #include "terrain_3d_util.h"
@@ -420,6 +421,40 @@ Ref<Image> Terrain3DUtil::luminance_to_height(const Ref<Image> &p_src_rgb) {
 		}
 	}
 	return dst;
+}
+
+void Terrain3DUtil::benchmark(Terrain3D *p_terrain) {
+	if (!p_terrain) {
+		return;
+	}
+	Terrain3DData *data = p_terrain->get_data();
+	if (!data) {
+		return;
+	}
+	uint64_t start_time;
+	Vector3 vec;
+	for (int i = 0; i < 3; i++) {
+		start_time = Time::get_singleton()->get_ticks_msec();
+		for (uint32_t j = 0; j < 10000000; j++) {
+			data->get_pixel(TYPE_HEIGHT, vec);
+		}
+		LOG(MESG, "get_pixel() 10M: ", Time::get_singleton()->get_ticks_msec() - start_time, "ms");
+	}
+
+	vec = Vector3(0.5f, 0.f, 0.5f);
+	for (int i = 0; i < 3; i++) {
+		start_time = Time::get_singleton()->get_ticks_msec();
+		for (uint32_t j = 0; j < 1000000; j++) {
+			data->get_height(vec);
+		}
+		LOG(MESG, "get_height() 1M interpolated: ", Time::get_singleton()->get_ticks_msec() - start_time, "ms");
+	}
+
+	for (int i = 0; i < 2; i++) {
+		start_time = Time::get_singleton()->get_ticks_msec();
+		p_terrain->bake_mesh(0);
+		LOG(MESG, "Bake ArrayMesh: ", Time::get_singleton()->get_ticks_msec() - start_time, "ms");
+	}
 }
 
 ///////////////////////////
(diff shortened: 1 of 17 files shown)


src/terrain_3d_storage.cpp
```diff
@@ -355,6 +355,36 @@ void Terrain3DStorage::set_map_region(const MapType p_map_type, const int p_regi
...
+			}
+			break;
+		default:
+			LOG(ERROR, "Requested map type is invalid");
+			break;
+			return nullptr;
+	}
+}
+
 Ref<Image> Terrain3DStorage::get_map_region(const MapType p_map_type, const int p_region_index) const {
```
TokisanGames, line 388: Why not remove the code redundancy by having this function call the other?
This builds, but I didn't test it. And it could be done inline.
```c++
Image *Terrain3DStorage::_get_map_region_ptr(const MapType p_map_type, const int p_region_index) const {
	return (Image *)&(_height_maps[p_region_index]);
}

Ref<Image> Terrain3DStorage::get_map_region(const MapType p_map_type, const int p_region_index) const {
	return _get_map_region_ptr(p_map_type, p_region_index);
}
```

Dekker3D: Mainly because I am not entirely sure yet what my version does. Having one function call the other might mean having a redundant cast (not that big a deal, since it should really only be for calls from outside the extension), or messing up the reference counting somehow (pretty big deal, seems kinda unlikely, not 100% sure).

TokisanGames:
get_map_region has been dropped. You can see here in set/get_pixel how the general API focus is changing towards selecting regions by location (fixed) rather than ids (mutable) and working with Regions. So the ptr function probably needs to be moved to Terrain3DRegion.

https://github.com/TokisanGames/Terrain3D/pull/374/commits/6d77480e1cf382c89a5284c6cf160a9c34d9b560#diff-f5a6479fa5fc2f89383edd16b70ed900e9b288af17f42366f3ac1bdb5a19306a

Although we are keeping the typedarrays referenced by ID, which are updated by update_maps, and then sent to the shader. So if more optimal we could undo these changes to set/get_pixel and reference the maps directly bypassing regions and refs.

Read the comments here
https://github.com/TokisanGames/Terrain3D/blob/6d77480e1cf382c89a5284c6cf160a9c34d9b560/src/terrain_3d_storage.h#L48-L79

I'd recommend just waiting until I finish my PR before spending any more time on it, but wanted to share the information to get you thinking about it.

TokisanGames:
I'm taking this over and expanding the optimizations throughout. One already helped improve #599.

src/terrain_3d_util.cpp
```diff
@@ -422,6 +423,40 @@ Ref<Image> Terrain3DUtil::luminance_to_height(const Ref<Image> &p_src_rgb) {
...
+		LOG(MESG, "get_height() 1M interpolated: ", Time::get_singleton()->get_ticks_msec() - start_time, "ms");
+	}
+
+	for (int i = 0; i < 2; i++) {
+		start_time = Time::get_singleton()->get_ticks_msec();
+		p_terrain->bake_mesh(0);
+		LOG(MESG, "Bake ArrayMesh: ", Time::get_singleton()->get_ticks_msec() - start_time, "ms");
+	}
+}
+
```
Xtarsia, line 459: does this want limting to debug builds only?

That probably goes for other classes that have the time include as well.

Xtarsia:
Everything looks good, painting noticably faster!

I tested painting texture with 1000m brush size, before this: 4.3s, After 3.1s 🎉

src/terrain_3d_util.cpp
TokisanGames: Hmm, it's not hurting anything to leave it.
The editor uses time for tracking the pen, and rate limiting error spam.

Issue #386, opened 2024-05-26
Provide proper generated C# bindings


lw64:
### Description

Currently to access Terrain3Ds API from C# you need to go over ClassDB and other functions to convert between C# and Godot objects. This is not very convenient for cases like generated Terrain based on code.

This project here seems to have a class that can generate csharp bindings from the godot scripting API: https://github.com/DmitriySalnikov/godot_debug_draw_3d/blob/master/src/editor/generate_csharp_bindings.h

I think that could be integrated.

TokisanGames:
Automated Wrapper f/ C#
https://github.com/Delsin-Yu/CSharp-Wrapper-Generator-for-GDExtension

Sythelux:
I tried to generate C# bindings with Delsin-Yu's Generator. It will not work out of the box. So I had to customize and fix some things.

That is also why I made the pull request against the 0.9 version, and not against the master. I don't want to have broken nightly builds because of the generator.

scrolling through the original CSharp-Wrapper-Generator-for-GDExtension it has a couple of addons placed into the addons folder which apparently get all automatically generated I'm not sure if it is worth to provide terrain3D to them actually.

I'll evaluate a couple of things. I don't want to put more work onto your shoulders and to have broken C# bindings automatically generated and no way to evaluate them is an issue.

TokisanGames:
Example bindings for Terrain3D in #454

ChristopherHaws:
Not trying to be pushy or anything seeing as this is a free, open source project! I also work on open source in my free time so I know it is a ton of hard work! 💙

Having said that, I am curious why the readme claims `Can be accessed by GDScript, C#, and any language Godot supports` when this doesn't seem to be the case? As far as I can tell, I am not able to access any of Terrain3D's types from C#... Am I missing something? Is there a nuget package with bindings I need to install or something?

TokisanGames:
@ChristopherHaws There's a whole page that shows you how to access it from other languages. We have many C# users.
https://terrain3d.readthedocs.io/en/latest/docs/programming_languages.html

This issue documents that there are no bindings, and no process setup to automatically generate them. I don't use C# so am not going to maintain them manually. This issue will remain open until there's an automatic process to produce them.

TokisanGames:
#686 has been updated with the latest binding. It's far enough along and now maintainable by me so that it can go into 1.1. It needs some testing though. I also added a demo, but I need some help with it to address some errors. The PR is now building so you can download the complete artifact.

[...]

PR #454, opened 2024-08-12
added Generated and cleaned up C# bindings

Sythelux:
Generated C# Bindings

refers to #386

### DIFF
--- a/project/addons/terrain_3d_csharp/ControlExtension.cs
+++ b/project/addons/terrain_3d_csharp/ControlExtension.cs
@@ -1,0 +1,59 @@
+using System;
+
+namespace Terrain3DExtensions;
+
+public static class ControlExtension
+{
+    public static byte GetBaseTextureId(this uint control)
+        => (byte)(control >> 27 & 0x1F);
+
+    public static void SetBaseTextureId(this ref uint control, byte baseTextureId)
+        => control = (control & ~((uint)0x1F << 27)) | (uint)((baseTextureId & 0x1F) << 27);
+
+    public static byte GetOverlayTextureId(this uint control)
+        => (byte)(control >> 22 & 0x1F);
+
+    public static void SetOverlayTextureId(this ref uint control, byte overLayTextureId)
+    {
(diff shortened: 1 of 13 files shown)


TokisanGames:
Thanks for a start on this.

First of all development should only be in the main branch, not 0.9. Main gets merged into 0.9 for releases. Working on 0.9 is working backward. The place to keep broken, unfinished code is in a branch and a draft PR. It won't be merged if it's breaking. 

Next, I don't think this is placed in the right directory. We don't want another addon. It should be in project/addons/terrain_3d/csharp/, with as much snake case as possible to match GDScript and C++ files. 

Third, we need an automated system, like what I linked in #386. With each release were chasing the API and structure. These bindings are already out of date. There's a lot of C# code here. I can't maintain a system I don't use.

Finally we need instructions for building, usage, and caveats. F mor testing and for inclusion in documention.

Sythelux:
I'm aware of all of the above. This was an intentional thing and I wrote about why I did it like this in the provided issue. I will close this PR, because I wouldn't want to merge it either, but I want to provide the code for anyone interested.

Dekker3D:
> Thanks for a start on this.
> 
> First of all development should only be in the main branch, not 0.9. Main gets merged into 0.9 for releases. Working on 0.9 is working backward. The place to keep broken, unfinished code is in a branch and a draft PR. It won't be merged if it's breaking.
> 
> Next, I don't think this is placed in the right directory. We don't want another addon. It should be in project/addons/terrain_3d/csharp/, with as much snake case as possible to match GDScript and C++ files.
> 
> Third, we need an automated system, like what I linked in #386. With each release were chasing the API and structure. These bindings are already out of date. There's a lot of C# code here. I can't maintain a system I don't use.
> 
> Finally we need instructions for building, usage, and caveats. For testing and for inclusion in documentation.

I disagree with the point on snake_case. Godot uses PascalCase for the C# API, even though it uses snake_case for all GDScript and C++ things. It would be good to match the same, in a C# binding for Terrain3D.

PR #456, opened 2024-08-15
Add manual slope painting tool

### LINKED ISSUE #282: Imlement Painting by Slope tool setting
### Description

Filter painting based on slope. Eg only paint on flat surfaces, or only on a certain angle.

This should be a `tool setting`, visible on all painting options: texture, color, wetness, autoshader, navigation.

See #96 for a slope sculpting tool.

Ryan-000:
Admin edit: 
Fixes #282 
* Implements slope filtered painting, foliage
* Reduces Brush / Advanced buttons to just icons to conserve space
* Advanced button no longer pops up on hover
* Removes picker icon file and uses icon embedded in the editor
-----

It's useful when you need to paint slopes with different types of materials (because autoshader only supports 1 material).

[exampleusage.webm](https://github.com/user-attachments/assets/d0b1903c-47a6-4f67-9fdd-a8b6e33c6676)


TokisanGames:
Thank you. Please see issue #282. This should be a tool setting, similar to paintable scale and rotation, not a tool. That will allow it to provide a slider to adjust slope and affect all tools.

az-raven:
This looks really really helpful, thank you so so much for putting in the time to make this!

TokisanGames:
Thank you for the changes. I tested it today. It's not very intuitive so I need to think about it and how to improve it or document it.

Some initial notes:
* The _effect_ of the default value of 70% is off from the _effect_ of the shader auto_slope default value of 1. The numbers can be different, but the effect should be the same when using the default values. Then you could manually paint to perfectly align with the autoshader. If the numbers are the same even, better. We could adjust the shader calculation as well because 1 is meaningless. But what does 70% mean? Maybe both should represent an angle of 0-180 degrees.
* Filling in holes (with CTRL) doesn't work while slope painting is enabled. 
* Painting holes w/ slope I think can be useful if you want to make a cave entrance in a wall without touching the floor.  However there seems to be some inaccuracies in the calculation, producing this. I cannot get rid of this without disabling slope and painting regular. These inaccuracies may exist in all of the other tools as well, but I haven't noticed other than what is here:
image
* Slope 100% should be identical to off, right? So we could potentially leave it enabled all the time (removing the checkbox and skipping the get_normal code if slope ==100%). However it seems to be inaccurate. Here I used the nav brush to remove w/ slope @ 100% and it ignored the flattest areas
image
* Slope applying to foliage would also be useful. Are you leaving it for me to implement?

src/terrain_3d_editor.cpp
```diff
@@ -405,28 +420,41 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
...
-							 */
-							if (_operation == ADD) {
-								dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * roughness), brush_alpha * strength);
-							} else {
-								dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * 0.5f), brush_alpha * strength);
-							}
-							break;
-						default:
-							break;
+					bool can_colour = true;
```
TokisanGames, line 423: Please use the american spelling of color, which matches the rest of the code.

src/terrain_3d_editor.cpp
```diff
@@ -405,28 +420,41 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
...
+					if (enable_slope_paint) {
+						real_t slope = storage->get_normal(brush_global_position).y;
+						if (invert_slope_paint) {
+							can_colour = slope > slope_paint_threshold;
+						} else {
+							can_colour = slope < slope_paint_threshold;
+						}
+					}
+
+					if (can_colour){
```
TokisanGames, line 433: Please install and use [clang-format](https://terrain3d.readthedocs.io/en/latest/docs/contributing.html#install-clang-format). It should have caught and fixed the lack of space here.

project/addons/terrain_3d/src/toolbar.gd
```diff
@@ -66,7 +66,6 @@ func _ready() -> void:
 	add_tool_button({ "tool":Terrain3DEditor.COLOR,
 		"add_text":"Paint Color", "add_op":Terrain3DEditor.ADD, "add_icon":ICON_COLOR,
 		"sub_text":"Remove Color", "sub_op":Terrain3DEditor.SUBTRACT })
-
```
TokisanGames, line 69: Keep this space

src/terrain_3d_editor.cpp
```diff
@@ -297,103 +300,115 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
...
-												angle = Vector2(-_operation_movement.x, _operation_movement.z).angle();
-												// Avoid negative, align texture "up" with mouse direction.
-												angle = real_t(Math::fmod(Math::rad_to_deg(angle) + 450.f, 360.f));
+					bool can_paint = true;
+					if (enable_slope_paint) {
+						real_t slope = storage->get_normal(brush_global_position).y;
+						if (invert_slope_paint) {
+							can_paint = slope > slope_paint_threshold;
+						} else {
+							can_paint = slope < slope_paint_threshold;
```
TokisanGames, line 309: I think one or both of these need to be >=, <=.
If the ground is flat, slope will be 1.
If slope threshold is 1, you can't paint on that part of the ground because it's not greater. That's the cause of one issue I noted.

This may also be the cause of the artifact when painting holes. Perhaps those sections were near vertical?
 
Also I wonder if there is any value for expanding the range to include 0%. Please test to ensure it does nothing on flat and perfectly vertical, inverted and not. If it does do something with the <=/>=, let's include it.

src/terrain_3d_editor.cpp
```diff
@@ -608,6 +636,7 @@ void Terrain3DEditor::set_brush_data(const Dictionary &p_data) {
 	// Santize editor data
 	_brush_data["size"] = CLAMP(real_t(p_data.get("size", 10.f)), 2.f, 4096.f); // Diameter in meters
 	_brush_data["strength"] = CLAMP(real_t(p_data.get("strength", .1f)) * .01f, .01f, 100.f); // 1-10k%
+	_brush_data["slope_paint_threshold"] = CLAMP(real_t(p_data.get("slope_paint_threshold", .7f)) * .01f, .01f, 100.f);
```
TokisanGames, line 639: See the line above. This is a range of 1% to 10,000%. Change the last number to `1.f`

TokisanGames:
Oh, you definitely do not want to ever merge. That makes the commit history messy. The proper operation is to rebase using the [Godot PR Workflow](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html) linked in our contributing document. However that merge commit you just made needs to be removed first before we can rebase.

src/terrain_3d_editor.cpp
```diff
@@ -436,6 +445,74 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
...
+				Vector3 pos00 = Vector3(Math::floor(pos.x / step) * step, 0.f, Math::floor(pos.z / step) * step);
+				real_t ht00 = storage->get_pixel(Terrain3DStorage::TYPE_HEIGHT, pos00).r;
+				Vector3 pos01 = pos00 + Vector3(0.f, 0.f, step);
+				real_t ht01 = storage->get_pixel(Terrain3DStorage::TYPE_HEIGHT, pos01).r;
+				Vector3 pos10 = pos00 + Vector3(step, 0.f, 0.f);
+				real_t ht10 = storage->get_pixel(Terrain3DStorage::TYPE_HEIGHT, pos10).r;
+				Vector3 pos11 = pos00 + Vector3(step, 0.f, step);
+				real_t ht11 = storage->get_pixel(Terrain3DStorage::TYPE_HEIGHT, pos11).r;
+				return bilerp(ht00, ht01, ht10, ht11, pos00, pos11, pos);
+			}
```
Xtarsia, line 493: since we are operating on the control map, there is no need to worry about sub-pixel positions. as such we can always use the rounded position, which will be faster, and prevent the edge cases where slopes right at the threshold sometimes were hard to paint back (especially noticable with holes)

```c++
        // copied from Terrain3DStorage::get_height, but modified to work with holes
        auto get_height = [&](Vector3 pos) -> real_t {
            real_t step = _terrain->get_mesh_vertex_spacing();
            // Round to nearest vertex
            Vector3 pos_round = Vector3(
                    round_multiple(pos.x, step),
                    0.f,
                    round_multiple(pos.z, step));
            real_t height = storage->get_pixel(Terrain3DStorage::TYPE_HEIGHT, pos_round).r;
            if (isnan(height)) {
                return 0.0f;
            } else {
                return height;
            }
```

az-raven:
Hi - I just wanted to check in, is there any way I can help out with this? I'm super excited to see this as a feature, it already seems incredibly helpful.

TokisanGames:
It needs to be finished to work with the instancer. Now that I'm done with regions, I can work on this sometime before the next release.

If you want to help, you could do the legwork on the needed code. Should just be a matter of checking the slope and not calling the instancer, done all within Terrain3DEditor. It's probably 2-4 lines of code.

az-raven:
Great! If I get the chance (I have some IRL things tomorrow) I'll take a shot at it. Thanks for the update. If I get it working, should I comment with a patch or request changes?

TokisanGames:
Comment here. Only Op and I can make changes to the PR.

az-raven:
I did not get the change to work on this tonight as I got super swamped today, but I will do my best to get things working tomorrow.

az-raven:
Here's what I've gotten working:

<details>
<summary>I put the diff within this fold-out to keep things less cluttered.</summary>

```diff
diff --git a/project/addons/terrain_3d/src/ui.gd b/project/addons/terrain_3d/src/ui.gd
index f8d170a..de97f32 100644
--- a/project/addons/terrain_3d/src/ui.gd
+++ b/project/addons/terrain_3d/src/ui.gd
@@ -220,10 +220,9 @@ func _on_tool_changed(p_tool: Terrain3DEditor.Tool, p_operation: Terrain3DEditor
                        to_show.push_back("remove")
 
                Terrain3DEditor.INSTANCER:
-                       # if you want, you can add this, but you'll have to modify terrain_3d_instances.cpp to support it
-                       # eg, it would be useful to avoid placing instances on steep hills
-                       #to_show.push_back("invert_slope_paint")
-                       #to_show.push_back("minimum_slope_paint_angle_degrees")
+                       to_show.push_back("invert_slope_paint")
+                       to_show.push_back("minimum_slope_paint_angle_degrees")
+                       set_menu_visibility(tool_settings.slope_list, true)
                        to_show.push_back("size")
                        to_show.push_back("strength")
                        set_menu_visibility(tool_settings.height_list, true)
diff --git a/src/terrain_3d_editor.cpp b/src/terrain_3d_editor.cpp
index 2e61e81..db003bc 100644
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -297,7 +297,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
                                        // Shader scale array is aligned to match this.
                                        std::array<uint32_t, 8> scale_align = { 5, 6, 7, 0, 1, 2, 3, 4 };
 
-                                       if (!_can_operate_on_slope(brush_global_position)) {
+                                       if (!can_operate_on_slope(brush_global_position)) {
                                                continue;
                                        }
 
@@ -409,7 +409,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
                                        dest = Color(as_float(bits), 0.f, 0.f, 1.f);
 
                                } else if (map_type == Terrain3DStorage::TYPE_COLOR) {
-                                       if (!_can_operate_on_slope(brush_global_position)) {
+                                       if (!can_operate_on_slope(brush_global_position)) {
                                                continue;
                                        }
 
@@ -445,7 +445,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
        storage->add_edited_area(edited_area);
 }
 
-bool Terrain3DEditor::_can_operate_on_slope(const Vector3 &p_brush_global_position) {
+bool Terrain3DEditor::can_operate_on_slope(const Vector3 &p_brush_global_position) {
        const real_t minimum_slope_paint_angle_degrees = _brush_data["minimum_slope_paint_angle_degrees"];
        const bool invert_slope_paint = _brush_data["invert_slope_paint"];
 
diff --git a/src/terrain_3d_editor.h b/src/terrain_3d_editor.h
index f09caa3..438d0d1 100644
--- a/src/terrain_3d_editor.h
+++ b/src/terrain_3d_editor.h
@@ -82,7 +82,6 @@ private:
        void _region_modified(const Vector3 &p_global_position, const Vector2 &p_height_range = Vector2());
        void _operate_region(const Vector3 &p_global_position);
        void _operate_map(const Vector3 &p_global_position, const real_t p_camera_direction);
-       bool _can_operate_on_slope(const Vector3 &p_brush_global_position);
        bool _is_in_bounds(const Vector2i &p_position, const Vector2i &p_max_position) const;
        real_t _get_brush_alpha(const Vector2i &p_position) const;
        Vector2 _get_uv_position(const Vector3 &p_global_position, const int p_region_size) const;
@@ -109,6 +108,7 @@ public:
        void operate(const Vector3 &p_global_position, const real_t p_camera_direction);
        void stop_operation();
        bool is_operating() const { return _pending_undo; }
+       bool can_operate_on_slope(const Vector3 &p_brush_global_position);
 
 protected:
        static void _bind_methods();
diff --git a/src/terrain_3d_instancer.cpp b/src/terrain_3d_instancer.cpp
index 8dbcac0..6b5248a 100644
--- a/src/terrain_3d_instancer.cpp
+++ b/src/terrain_3d_instancer.cpp
@@ -222,6 +222,8 @@ void Terrain3DInstancer::add_instances(const Vector3 &p_global_position, const D
        real_t random_hue = CLAMP(real_t(p_params.get("random_hue", 0.f)) / 360.f, 0.f, 1.f); // degrees -> 0-1
        real_t random_darken = CLAMP(real_t(p_params.get("random_darken", 0.f)) * .01f, 0.f, 1.f); // 0-100%
 
+       Terrain3DEditor* editor = _terrain->get_editor();
+
        TypedArray<Transform3D> xforms;
        TypedArray<Color> colors;
        for (int i = 0; i < count; i++) {
@@ -240,6 +242,10 @@ void Terrain3DInstancer::add_instances(const Vector3 &p_global_position, const D
                        position.y = height;
                }
 
+               if (!editor->can_operate_on_slope(position)) {
+                       continue;
+               }
+
                // Orientation
                Vector3 normal = Vector3(0.f, 1.f, 0.f);
                if (align_to_normal) {


```

</details>

I made `_can_operate_on_slope` into a public method (wasn't sure if this is the right way, but it prevents a lot of code duplication) and used it to filter each instance based on its position. If you have a more preferred way to go about this, please let me know!

Super happy to help out!

TokisanGames:
@az-raven Thanks for this effort. Unfortunately it's not right. That code will crash if they're running without an editor. My direction was to do all of the work within Terrain3DEditor. 

_operate_map processes every brush pixel for painting and foliage operations. The editor current PR inserts a slope check on every brush pixel for painting. It needs to do the same for foliage, within _operate_map.

Looking at it, all we need to do is call _can_operate_on_slope here:
https://github.com/TokisanGames/Terrain3D/blob/main/src/terrain_3d_editor.cpp#L163-L170

If you want to try again great, otherwise I will fix it up sometime before the next release in a couple of weeks.

az-raven:
Ahh I'm so sorry. I had forgotten about being able to add instances at runtime. I'll do my best to revise it to work properly.

az-raven:
I may be misunderstanding something here, but it seems as though putting the call there would not work as intended. Since that section of terrain_3d_editor works based off the brush position (rather than the per-instance position) it would just filter the entire brush based off its center rather than each instance based off its position. Was that the intended behavior, or am I misinterpreting your advice?

TokisanGames:
That's probably true. Alright then I'd say `Editor::can_operate_on_slope(position)` should be moved to `Storage::is_in_slope(position, angle)` then both editor and instancer can call it, and it can access Storage directly. Instancer should be able to operate entirely without the editor. The only call is a backup function specifically for the editor's needs, that nullifies itself when not in use. However this PR is a bit out of date with the new structure and you can't update it. So I'd suggest letting me get to it in a while, I'll rebase it, look at your patch, and refactor it.

az-raven:
Great! I'll do my best to help out with those as soon as I handle my homework for the week.

az-raven:
> I don't think there's any need for slope on auto/navigation/holes.

Hi - I don't mean to intrude, but I feel as though slope filtering on navigation would be extremely useful for a variety of use cases. I believe it would particularly simplify the workflow for large terrains or when trying to be consistent with what areas are navigable. That said, I would still love to hear your thoughts on the matter.

TokisanGames:
Godot navigation is not robust. One shouldn't just draw a massive area covering all paths and pass that to Godot and expect it will generate a good result. It needs to be refined a bit, and segmented into areas.

Second, slope occurs per pixel, so will produce non-contiguous application. This is fine for texturing but terrible for holes and navigation generation. 



Third, when users remove slope enabled navigation, they're going to leave pixels behind. We don't need to create more problems and complaints due to the navigation server doing weird things because we're feeding it garbage data.

az-raven:
That makes sense. Thank you for the clarification!

TokisanGames:
Updated the slope method based on discord discussions. This has resulted in a double slider that eliminates the need for invert. Set the range you want and paint. It's the most intuitive option I've seen so far. Please try it out in the push build.

image

I'll tackle foliage tomorrow probably.

TokisanGames:
I have reinstated the std:: namespace on isnan for linux. You should be able to build now.

I have enabled the invert key CTRL to inverse the operation, so now if the range is 37-60 the inverted range will be "not 37-60" or 0-37 and 60-90.

TokisanGames:
Instancer has been implemented works. Thanks @az-raven.

I instead used Alt to reverse the slope. Paint color + CTRL, Instancer + CTRL already removes, so conflicted. I need to document all of the keyboard operations, so drafted them below.

This is almost ready to merge. The last thing is that the embedded editor picker icon doesn't work in compatibility mode. I might have to revert the file.

**General Keys**
* <kbd>LMB</kbd> - Click the terrain to positively apply the current tool.
* <kbd>Ctrl + LMB</kbd> - **Inverse** the tool. Removes regions, height, color, wetness, autoshader, holes, navigation, foliage.
* <kbd>Shift + LMB</kbd> - Change to the **Smooth** sculpting tool if shift is the _only_ modifier pressed.
* <kbd>Ctrl + Z</kbd> - Undo. You can also view the operations in Godot's `History` panel.
* <kbd>Ctrl + Shift + Z</kbd> - Redo.
* <kbd>Ctrl + S</kbd> - Save the scene and all data.

**Sculpting**
* <kbd>Alt + LMB</kbd> - **Lift floors.** This lifts up lower portions of the terrain without affecting higher terrain around it. Use it along the bottom of cliff faces. See [videos demonstrating before and after](https://github.com/TokisanGames/Terrain3D/pull/409). 
* <kbd>Ctrl + Alt + LMB</kbd> - **Flatten peaks.**. The inverse of the above. This reduces peaks and ridges without affecting lower terrain around it.

**Slope Enabled Operations**
These operations support: **Paint**, **Spray**, **Color**, **Wetness**, **Instancer**.
* <kbd>LMB</kbd> - Operate within the set slope.
* <kbd>Alt + LMB</kbd> - Operate outside the set slope.
* <kbd>Ctrl + Alt + LMB</kbd> - Inversely operate outside the set slope. eg. Remove color or foliage with the opposite slope.

**Instancer**
* <kbd>LMB</kbd> - Add the selected mesh instance to the terrain.
* <kbd>Ctrl + LMB</kbd> - Remove instances of the selected type.
* <kbd>Ctrl + Shift + LMB</kbd> - Remove instances of any type.

TokisanGames:
* Fixed the wonky slider.
* Fixed the broken slope sculpt brush picker icons

This one is finished. Thanks for the help guys. Any more bug fixes or improvements can go in another PR. Good job everyone.

PR #457, opened 2024-08-19
Enable use of pre-processor in the generated shader

Xtarsia:
Allows use of #define / #ifdef etc pre-processor statements in the generated shader and the //INSERT: system.


TokisanGames:
Wow, so simple, and it works well. 

So, the issue was that the rendering server's version of the shader (rid) doesn't preprocess the shaders, right? (That's an issue we should report.) And you've solved it by using a regular Shader object.

In that case, I think we can simplify this a bit more and remove more code and variables. Please review my changes and the total PR before I merge.

Xtarsia:
Yes, much cleaner overall now too. pre-processor commands still function as they should with the generated shader code.

TokisanGames:
Great work. Thanks!

Issue #314, opened 2024-02-02
Mobile render shows stepped heights


TokisanGames:
### Description

The mobile renderer displays stepped heights. I can't find anyway to get rid of it. It's in the vertex shader. Resculpting or troubleshooting the shader by adding highp everywhere doesn't help. It looks like floating point precision, 

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/57e9e0a1-b7c0-45e6-a638-3e0dfc5e953e)

Apparently it only affects some of the demo regions.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/b7cdb1d4-aa23-43d3-bdfb-f3702d1697b1)

The precision gets worse the farther away from (0,0,0) we go.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/d3e93287-eb32-4622-b0a3-4213698151c5)

Switching back to the Forward renderer, any sculpted heights that appeared stepped are still there, though a little smoothed. And then they can be smoothed out.

There may be no fix, but we should at least document why. Presumably the renderer or engine has some low precision in texture arrays, texture formats, vertex locations or something.

FishOfTheNorthStar:
Hi, first post here ( I think ) but I've been using Terrain3D a while now.  Great project thanks a lot.  I'm going to start trying to feed some of the little changes and adjusts I've made back into the project soon, so I thought I'd have a look at open issues.

So about this here, i've been struggling with 16 bit floats a bit lately and here's what I've found.  First off, you don't get 16 actual bits of accuracy, of course.  More like 12ish?  distinct values from 0->4096, after that it's pretty murky, because of how floats dedicate bits to an exponent here that isn't actually getting used for normalized values.  That's my understanding anyways.  So this is how I mitigate that on my end for terrain generation, maybe it gives you some ideas.

Let's say I'm starting with true floats in normal range 0->1, and I want to encode that value using 16 bits (all of them), without any floating point nonsense at all.  First I'll convert it to an int between 0->65535, easy enough.  Then I split that into two byte values, high and low order (divide by 255 for high order, then multiply that (int value) by 255 and subtract it from the 16 bit int again to get the low order).  So now you have two bytes, and all 16 bits used, and can encode that into 8 bit channels which (hopefully), mobile can universally figure out how to access without mucking up the values anywhere.  To store that, I typically use Image.Format.La8, for 8 bits in red and 8 bits in alpha.  I put the high order in red, low in alpha.

[...]

PR #458, opened 2024-08-19
remove region_blend_map

Xtarsia:
swap to region blend function, removed region blend map.
removed 2 uniforms in lieu of a fixed blend distance.
~~removed offset from get_region_uv2() as it was causeing an artifact.~~

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -86,26 +86,22 @@ varying float v_region_border_mask;
 // Takes in UV world space coordinates, returns ivec3 with:
 // XY: (0 to _region_size) coordinates within a region
 // Z: layer index used for texturearrays, -1 if not in a region
-ivec3 get_region_uv(vec2 uv) {
-	uv *= _region_texel_size;
-	ivec2 pos = ivec2(floor(uv)) + (_region_map_size / 2);
-	int bounds = int(pos.x >= 0 && pos.x < _region_map_size && pos.y >= 0 && pos.y < _region_map_size);
+ivec3 get_region_uv(const vec2 uv) {
+	ivec2 pos = ivec2(floor(uv * _region_texel_size)) + (_region_map_size / 2);
+	int bounds = int(uint(pos.x | pos.y) < uint(_region_map_size));
 	int layer_index = _region_map[ pos.y * _region_map_size + pos.x ] * bounds - 1;
-	return ivec3(ivec2((uv - _region_offsets[layer_index]) * _region_size), layer_index);
+	return ivec3(ivec2(mod(uv,_region_size)), layer_index);
 }
 
 // Takes in UV2 region space coordinates, returns vec3 with:
 // XY: (0 to 1) coordinates within a region
 // Z: layer index used for texturearrays, -1 if not in a region
-vec3 get_region_uv2(vec2 uv) {
-	// Vertex function added half a texel to UV2, to center the UV's.  vertex(), fragment() and get_height()
-	// call this with reclaimed versions of UV2, so to keep the last row/column within the correct
-	// window, take back the half pixel before the floor(). 
-	ivec2 pos = ivec2(floor(uv - vec2(_region_texel_size * 0.5))) + (_region_map_size / 2);
-	int bounds = int(pos.x >= 0 && pos.x < _region_map_size && pos.y >= 0 && pos.y < _region_map_size);
+vec3 get_region_uv2(const vec2 uv2) {
+	// Remove Texel Offset to ensure correct region index.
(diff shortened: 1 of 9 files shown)


Xtarsia:
> > removed offset from get_region_uv2() as it was causeing an artifact.
> 
> Thank you for the update. Is this the artifact making the terrain stepped in the mobile renderer? #314

sadly not.

That issue can be fixed if the new normal method is employed, and the heightmap filtering can be set to nearest, along with removing the 0.5 offset when UV2 is calculated.

TokisanGames:
> -ivec2 pos = ivec2(floor(uv - vec2(_region_texel_size * 0.5))) + (_region_map_size / 2);
> +ivec2 pos = ivec2(floor(uv)) + (_region_map_size / 2);

I see the artifact when this is reverted. This was introduced by PR #348 which I merged based on Fish's recommendation and your approval, but still don't understand the original problem he posed, nor did I see any difference. Please review Fish's arguments there again (related to this offset) and confirm that what he wrote is not relevant or necessary.

Xtarsia:
hmmm actually I think it is wrong. the top image is actually correct, because the bilinear interpolation is (correctly) not occuring across region boundarys.

Xtarsia:
yeah, without it is correct, can can clearly see how the map pixels are centered over the vertices.

image


the color map not blending across regions is technically correct too (and could be fixed via conditional lookups, but not sure its worth the cost, perhaps worth opening an issue for it)

TokisanGames:
From a user perspective, when painting, regions should be invisible.

**Blending**
When painting directly across a region boundary in one swipe, it won't be noticeable.
image

Unknowingly painting along the region boundary will produce an artifact and maybe confusion.
image

The offset didn't fix blending over the region boundary before, so this has always been an issue. But I've never noticed it and it is easily fixable by the user. So I think we can ignore this artifact for now.

**Offset**
However, this does not look correct. 
image

With regions on both sides, the color should go right up to the region boundary so it is consistent. The code may be mathematically correct, but from a user experience, this looks wrong. Looking at the four edges of a region, it's consistently offset by (-0.5, -0.5) vertices on all region boundaries, which the offset does fix. I don't see this stretching.

Xtarsia:
its certainly a confusing issue,

vertex indices go from 0 to 1023 along x/z, so;
The bottom left edge of any region has control/height/color map values from that region.
The top right edge of any region has control/height/color map values from the next region.

imo, for true correctness, the vertex grid region lines (the thick white ones only) should be offset half a texel.

But what looks good to user perspective, does matter more (and smoke and mirrors is what game dev is all about tbh) so i've undone the undo of the undo (?) again and will leave it at that before this gets stuck in the weeds.

src/shaders/main.glsl
```diff
@@ -86,26 +86,22 @@ varying float v_region_border_mask;
 // Takes in UV world space coordinates, returns ivec3 with:
 // XY: (0 to _region_size) coordinates within a region
 // Z: layer index used for texturearrays, -1 if not in a region
-ivec3 get_region_uv(vec2 uv) {
-	uv *= _region_texel_size;
-	ivec2 pos = ivec2(floor(uv)) + (_region_map_size / 2);
-	int bounds = int(pos.x >= 0 && pos.x < _region_map_size && pos.y >= 0 && pos.y < _region_map_size);
+ivec3 get_region_uv(const vec2 uv) {
+	ivec2 pos = ivec2(floor(uv * _region_texel_size)) + (_region_map_size / 2);
+	int bounds = int(uint(pos.x | pos.y) < uint(_region_map_size));
```
TokisanGames, line 91: Wow. I had to think about this one, and I still don't understand how it works if x or y are negative. Did you come up with this?

TokisanGames:
Ah, perhaps we need to rethink where that vertex border should be considering image map pixels, collision, and blending between vertices. I want to think more about it. For now, thanks for removing it so we can merge this one in.

Please remove the `get_region_blend_map()` block from CodeGenerated.gd. And you can also remove the TextureRect from UI.tscn.

I think some people would like to blend in the noise and not have a hard coded moat around their regions. Something along these lines, but also blending within the region a bit. This just blends outside the region.
```
uniform float noise_blend1 : hint_range(0, 0.5) = 0.;
float region_blend(vec2 uv2) {
	...
	//vec2 w = smoothstep(vec2(0.), vec2(1.), fract(uv2));
	vec2 w = smoothstep(vec2(noise_blend1*2.), vec2(1.), fract(uv2)+vec2(noise_blend1));
```

Xtarsia:
Rebased and should be good to go.

TokisanGames:
Great work! Thank you.

PR #463, opened 2024-08-26
adjust UV2 calculation in vertex. Fixes #314

### LINKED ISSUE #314: Mobile render shows stepped heights

Xtarsia:
using fma() here avoids precision loss.

+ 0 ~ 15fps in mobile, no change for forward+

-----
Admin edit:
Fixes #314

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -134,7 +134,7 @@ void vertex() {
 	UV = round(v_vertex.xz * _mesh_vertex_density);
 
 	// UV coordinates in region space + texel offset. Values are 0 to 1 within regions
-	UV2 = (UV + vec2(0.5)) * _region_texel_size;
+	UV2 = fma(UV, vec2(_region_texel_size), vec2(0.5 * _region_texel_size));
 
 	// Discard vertices for Holes. 1 lookup
 	v_region = get_region_uv(UV);


TokisanGames:
Interesting. In my list if optimizations, I noted we should move to fma everywhere we can since it's supposed to be 10% faster. However it's not supported by the compatibility renderer. Will that make it difficult for you supporting it? Will we have 2 shaders, or can we split it with defines and inserts? Maybe we can redefine fma.

Xtarsia:
it may end up having to be a seperate shader, tho, this line

```glsl
UV2 = UV * _region_texel_size + 0.5 * _region_texel_size;
```
works just as well (though its slightly slower)

TokisanGames:
Excellent work, thank you @Xtarsia. Also looks fine out at -8k, -8k.

Another shader is fine. Currently compatibility mode complains about fma, textureQueryLod, and dFdyCoarse. However, adding this works: `#define fma(a,b,c) a*b+c;`


@Saul2022 is this the last standing mobile issue? I think everything else has been resolved.

Issue #290, opened 2023-12-31
Editor tablet pen support


dkaste:
### Description

I'm not sure how common it is to edit terrain with a graphics tablet, but it seems pretty natural to me. Godot automatically provides pen information through `InputEventMouseMotion`, which should make this feature pretty straightforward to add.

Pen pressure in particular should be very easy to support. Simply multiply the brush opacity by `InputEventMouseMotion.pressure`, or something along those lines.

There's also pen "tilting" which could change the angle of the brush, I suppose.

TokisanGames:
Zylann also uses pressure from the input event.

https://github.com/Zylann/godot_heightmap_plugin/blob/74437e6d1c106adee38f9c0fd0e775759b6e2974/addons/zylann.hterrain/tools/plugin.gd#L502

TokisanGames:
Pen pressure support added in https://github.com/TokisanGames/Terrain3D/commit/a67d5801a6d9edc7973f4634d396b7566bd228ea

If pen pressure is non-zero, strength is ignored. This means though that pen pressure cannot draw stronger than 100%, unlike the mouse.

@aristides86 @dkaste Try it in a nightly build. Download the artifact.
https://github.com/TokisanGames/Terrain3D/actions/runs/10465611812

aristides86:
Thank you. Pressure works but could it be that it breaks strength for the mouse now? In the video the first stroke made with the stylus with increasing pressure, the second with the mouse with 1% strength. Strength works well in 0.9.2-beta release.

https://github.com/user-attachments/assets/81a84fe6-7e7b-4eeb-9a0e-af3fac543f85

TokisanGames:
This looks beautiful. Please send me info so I can include it here
https://terrain3d.readthedocs.io/en/latest/docs/games.html

In 0.9.3-dev raise/lower strength was reduced 90%. However, you can increase strength up to 10,000%.

I can reproduce the issue. The current code says:
* If mouse_pressure is >0 use it, else use brush_strength

The problem is that if the mouse is clicked, pen pressure is 1 instead of 0. And when the tablet is pressed, godot reports that as the left mouse button. I see no way to determine if the user is moving the pen or the mouse when pressure is 0 or 1.

So which is better:
* effective strength = mouse_pressure * brush_strength
* effective strength = mouse_pressure if 0-.999; brush_strength if 1
* or ?

[...]

Issue #286, opened 2023-12-27
Can work like UE's alpha brush


CsloudX:
### Description

Use alpha brush, we can brush montain every easy.

https://github.com/TokisanGames/Terrain3D/assets/35988767/fd76ab3e-1087-4242-a8cf-0724cc2e6faf

It like need this tasks:

- [ ] Let can set brush size to larger (Current can only set less than 200m, for montain, it was too small)
- [ ] When mouse pressed , brush once (Current brush when mouse move)

TokisanGames:
That is not a brush. That is an alpha stamp. This means you provide a 2k-4k height image for a stamp.

The current brushes are very small, 100px or 256px. Even if you did stamp them very large, they'd be ugly and blocky. 

Alpha stamps are not supported, per se. It's not just these 4 options you need. There are other things needed, notably 2k-4k alpha stamps. 

Currently, you can import stamps in any region using the importer. Otherwise you're best bet is to work in a photo editing app for now. Then you can place your 2k-4k stamp wherever you like, at any rotation and import that heightmap. 

Additionally, 
Larger brush size needs #174.

You can set Brush rotation by disabling jitter and enabling align to view in the advanced settings. Paintable UV rotation is planned.

TokisanGames:
Action on mouse click instead of movement was added some time ago.
Size > 100% added some time ago.
Strength > 100% enabled in 7e75cd47c945f0ec199f2ec5c067b8f7b8b4a9c8.

You could technically do this now, though it's not as smooth as it could be. 

1. Insert higher resolution alpha stamps into the brush directory, no built in tool to manage them
2. Set brush size and strength very high.
2. Align the camera to the angle you want and disable jitter. No specific angle option.
3. Click once.

One of the default low res alpha mask zoomed up, stamped, then smoothed:

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/e7a6dfac-54c2-4fc6-9c08-a02bd81c7956)

This ticket will remain open until it's a smoother process, maybe brush management at least.

TokisanGames:
#468 Made a huge improvement for alpha stamping that eliminates the need for smoothing if using a small resolution 100x100 stamp at 1024x1024. However, it still needs smoothing if stamping at 2k+. See pics.

For future improvements, we need:
* An alpha stamp mode that will:
    * Upscale the selected brush to 4096x4096
    * Stop hiding the decal when placing
    * Snap to region?
[...]

PR #468, opened 2024-08-27
Upscale brush images when loading

### LINKED ISSUE #445: Use linear filtering for painting height
### Description

When painting the world and using a very large brush size like everything over 150 meters, the aliasing of the brushes become painfully obvious. This can be seen in this image
image

I used one of the basic round brushes to create this cliff looking terrain and due to the large brush size, the falloff becomes very step-like. Having the option to paint using a linearly sampled brush texture would get rid of this issue, as such step-like patterns would become interpolated and thus smoothed over.

The image is an extreme case of the problem, but it is still very obvious even with less strong brush stregnth as seen here
image
I used the very smooth round brush for making this image with a brush size of 500 meters.

I would also advise giving the user the option to switch between nearest-neighbour (the current sampling method) and linear sampling in the brush settings list (at the bottom of the 3d viewport).

### Notes
- I am aware that the smoothing tool can also get rid of this issue, but having the option to immediately paint with smooth terrain makes the editing process a lot quicker.

Xtarsia:
closes #445

I tested 2k and 4k and saw no noticeable improvement over 1k, however project load time when opening the editor did increase a little with 2k and a lot with 4k.

### DIFF
--- a/project/addons/terrain_3d/src/tool_settings.gd
+++ b/project/addons/terrain_3d/src/tool_settings.gd
@@ -243,6 +243,8 @@ func add_brushes(p_parent: Control) -> void:
 			if !dir.current_is_dir() and file_name.ends_with(".exr"):
 				var img: Image = Image.load_from_file(BRUSH_PATH + "/" + file_name)
 				img = Terrain3DUtil.black_to_alpha(img)
+				if img.get_width() < 1024 and img.get_height() < 1024:
+					img.resize(1024, 1024, Image.INTERPOLATE_CUBIC)
 				var tex: ImageTexture = ImageTexture.create_from_image(img)
 
 				var btn: Button = Button.new()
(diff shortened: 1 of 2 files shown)


TokisanGames:
Wow, this is a huge improvement.

Here's the argument for 4k. **_I stamped out this 4k terrain with 1 click_** using the 100x100 Terrain3 brush, size 4096, strength 50,000%. This is an amazing result and so much easier than trying to hand sculpt this.

However, you can see in the corner that when all the brushes are 4k, the editor took forever to open and it runs like garbage. It runs fine in game. So 1024 sized brushes are fine, but we need an alpha stamp mode that will upscale as large as we want.

image


1k brush @ 1024. Nice.

image

2k brush @ 1024 required a lot of smoothing afterwards. Not good. Easier to change the code, restart, stamp, change it back.

image

This should only upscale if less than 1024x1024. It shouldn't downscale. Many users have 2k or 4k alpha stamps and we don't want to disallow that. I fixed that and also bumped up the max up to 100,000% which equates to 1000m per click.

Alpha stamp changes moving forward noted in https://github.com/TokisanGames/Terrain3D/issues/286#issuecomment-2315027242

(Note brush strength is currently broken #290)

PR #476, opened 2024-09-02
Single to multiple regions conversion wizard

TokisanGames:
Builds off of #374. Converts single storage resource file to multiple storage files and facilitates setup.

TODO
* [x] Fix wizard menu
* [x] Add wizard button to inspector https://x.com/passivestar_/status/1795886728646975820 if storage_dir is empty
* [x] Update conversion wizard in GDScript
* [x] Port wizard to C++
* [x] Rename Terrain3DStorage to Terrain3DData
* [x] Strip Terrain3DStorage down to convertable resource
* [x] Convert demo
* [x] Update docs


Xtarsia:
image

Xtarsia:
Instant crash with no logs:
open the artifact project, open node3d.tsn, add 1 region, change mesh size .

Xtarsia:
attempting to add regions by clicking outside the area, and then saving:

image

TokisanGames:
The Directory setup wizard now works to upgrade an old version. Here are our alerts to begin our upgrade process. I don't think it can be any simpler than this when the user loads an old scene, and an inspector button is redundant.
* This pops up whenever a scene w/ a 0.9.2 storage loads.
* Console warning saying Use Terrain3D Tools / Directory Setup to upgrade
* Scene tree Terrain3D node yellow icon says the same, any time Data Directory is empty.
* Release notes saying the same

The popup is automatic and the old storage file is auto popuplated. All the user needs to do is specify a new directory, click ok. On save, it will write to disk. 

image

PR #482, opened 2024-09-06
Subdivide LOD1 and higher

Xtarsia:
Makes distant LOD levels look much better with a minimal Triangle count increase, so much so that useing half the mesh size yields visually similar results, as seen in this comparison:

mesh size 32 (LOD set to 7):
image

original, mesh size 64 (LOD set to 6):
image

Also, maintains 1:1 with current collision generation
image

### DIFF
--- a/src/geoclipmap.cpp
+++ b/src/geoclipmap.cpp
@@ -9,6 +9,106 @@
 // Private Functions
 ///////////////////////////
 
+// Half each triangle, have to check for longest side.
+void GeoClipMap::_subdivide_half(PackedVector3Array &vertices, PackedInt32Array &indices) {
+
+	// Hash for map for quick vertex search, for loop was very very slow!
+	struct Vector3Hash {
+		std::size_t operator()(const Vector3 &v) const {
+			std::size_t h1 = std::hash<float>()(v.x);
+			std::size_t h2 = std::hash<float>()(v.y);
+			std::size_t h3 = std::hash<float>()(v.z);
+			return h1 ^ (h2 << 1) ^ (h3 << 2);
+		}
+	};
+
+	PackedVector3Array new_vertices;
+	PackedInt32Array new_indices;
+
+	std::unordered_map<Vector3, int, Vector3Hash> vertex_map;
+
+	auto midpoint = [](const Vector3 &p1, const Vector3 &p2) -> Vector3 {
+		return (p1 + p2) / 2.0f;
+	};
+
+	auto find_or_add_vertex = [&vertex_map, &new_vertices](const Vector3 &vertex) -> int {
+		auto it = vertex_map.find(vertex);
+		if (it != vertex_map.end()) {
+			return it->second;
+		} else {
+			int index = new_vertices.size();
+			vertex_map[vertex] = index;
+			new_vertices.push_back(vertex);
+			return index;
+		}
+	};
+
+	for (int i = 0; i < indices.size(); i += 3) {
+		int id_0 = indices[i];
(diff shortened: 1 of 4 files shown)


TokisanGames:
I tested this a bit, took video and analyzed the footage. I think this is a good improvement. Maybe we'll look at it again if we switch to a symmetric grid but for now I like what is going on.

The full mesh @ 64 before was 741k vertices. 
The new mesh @ 32 is 345k, with a 10-20% speed improvement.
The new mesh @ 48 is 760k, with a 0-20% speed improvement.

32 is too chunky for my taste. Let's put the demo at the same vertex count, with a 48 mesh size.

The only thing I don't like is the edges of the last LOD get sawtoothed on one side. Can you fix this?

{DA8ECF8C-0CC4-4E43-8D72-8808B4C22D07}

src/geoclipmap.cpp
```diff
@@ -9,6 +9,106 @@
...
+};
+
+// Half each triangle, finer grain control, have to check for longest side.
+static void _subdivide_half(PackedVector3Array &vertices, PackedInt32Array &indices) {
+	PackedVector3Array new_vertices;
+	PackedInt32Array new_indices;
+
+	std::unordered_map<Vector3, int, Vector3Hash> vertex_map;
+
+	auto midpoint = [](const Vector3 &p1, const Vector3 &p2) -> Vector3 {
```
TokisanGames, line 30: Not a fan of these autos, but looking at the type in MSVC, I suppose they are fine.

src/geoclipmap.cpp
```diff
@@ -9,6 +9,106 @@
...
+	std::size_t operator()(const Vector3 &v) const {
+		std::size_t h1 = std::hash<float>()(v.x);
+		std::size_t h2 = std::hash<float>()(v.y);
+		std::size_t h3 = std::hash<float>()(v.z);
+		return h1 ^ (h2 << 1) ^ (h3 << 2);
+	}
+};
+
+// Half each triangle, finer grain control, have to check for longest side.
+static void _subdivide_half(PackedVector3Array &vertices, PackedInt32Array &indices) {
```
TokisanGames, line 24: Let's put this and the struct above into the class. No reason to have them floating outside, is there?

src/terrain_3d.cpp
```diff
@@ -242,21 +242,36 @@ void Terrain3D::_build_meshes(const int p_mesh_lods, const int p_mesh_size) {
 				if (l != 0 && (x == 1 || x == 2) && (y == 1 || y == 2)) {
 					continue;
 				}
-
-				RID tile = RS->instance_create2(_meshes[GeoClipMap::TILE], scenario);
+				RID tile;
+				if (l == 0) {
```
TokisanGames, line 246: Please change `l` here and above and below to `lod`.

TokisanGames (changes requested):

Xtarsia:
the sawtooth is a result of adding extra vertices at the yellow dots ive marked in this image:

image

I can't think of a reasonable way to avoid them.

src/geoclipmap.cpp
Xtarsia: Me either, but I too came to the conclusion that they were fine.

Xtarsia:
Rebased, review points addressed.

TokisanGames:
> I can't think of a reasonable way to avoid them.

What about adding the final square edge vertices?

Xtarsia:
jagged edges will be resolved by #488 😃

TokisanGames:
Thank you!

PR #488, opened 2024-09-13
Fix background mode none region borders

Xtarsia:
examples with mesh size 8;

before:
image

after:
image


TokisanGames:
Wow... really, really nice. This is an improvement that makes a game production ready. It was so sloppy before.

PR #492, opened 2024-09-16
Improve brushing

### LINKED ISSUE #192: Lock  to a specific material when painting color in terrain.
It may seem confusing but the idea is that  when painting a color you can go to the surface list, select the material you want and it will only paint on that material., taking advantage that now you can assign names to the textures and maybe in the future could be applied to the foliage feature when it comes to paint on a specific material, though 
 there can be an option in the color brush sethings where you can unlock that and paint on whatever material you want.

--------
Admin edit:
* Limit painting of color, texture, scale, rotation, foliage to a texture layer mask that are already present on the terrain (eg. a layer chooser in the inspector like collision mask or render layers)

### LINKED ISSUE #481: Spray w/ auto direction overwrites the direction of base texture
### Description

Spraying with auto direction overwrites the direction on the base layer below. Spray should only change direction on the Spray pixels (where brush_alpha >. 1), I think. Due to blending, it might not work as we imagine. Might have to bump the threshold up to 0.3 or so. 

https://discord.com/channels/691957978680786944/1065519581013229578/1275318563149840437

Xtarsia:
Changed alpha threshold to 0.5 along with extra conditions to make painting feel a bit more intuitive. 

~~blending may be a little slower again with the default circle 0 brush, useing circle 1 is similar to before~~

painting over the auto shader, or painting base onto base, or spray onto spray now follows the percieved brush shape much better.

angle and scale now also much more closley follow the visibility of the selected texture too.

images with height textures that are not centered around 0.5 maybe end up with a small inward, or outward halo for angle/scale values, however this is still an improvement over the entire brush square getting overwritten as it did before this change.

closes #481 tho a pure solution would be seperate values for base/over but that is probably not needed.

----
Admin edit
Fixes #192

### DIFF
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -365,7 +365,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
 											autoshader = false;
 										}
 									} else {
-										// Else overlay and base are separate, set overlay texture and increase blend value										
+										// Else overlay and base are separate, set overlay texture and increase blend value
 										blend = CLAMP(blend + brush_value, 0.f, 1.f);
 										if (blend > 0.5f && brush_alpha > 0.5f) {
 											overlay_id = asset_id;
@@ -436,26 +436,38 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
 				dest = Color(as_float(bits), 0.f, 0.f, 1.f);
 
 			} else if (map_type == TYPE_COLOR) {
-				switch (_tool) {
-					case COLOR:
-						dest = src.lerp((_operation == ADD) ? color : COLOR_WHITE, brush_alpha * strength);
-						dest.a = src.a;
-						break;
-					case ROUGHNESS:
-						/* Roughness received from UI is -100 to 100. Changed to 0,1 before storing.
-						 * To convert 0,1 back to -100,100 use: 200 * (color.a - 0.5)
-						 * However Godot stores values as 8-bit ints. Roundtrip is = int(a*255)/255.0
-						 * Roughness 0 is saved as 0.5, but retreived is 0.498, or -0.4 roughness
-						 * We round the final amount in tool_settings.gd:_on_picked().
-						 */
-						if (_operation == ADD) {
-							dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * roughness), brush_alpha * strength);
-						} else {
-							dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * 0.5f), brush_alpha * strength);
-						}
-						break;
-					default:
-						break;
+				// Filter by visible texture
+				bool can_write = true;
+				if (enable_texture) {
+					Ref<Image> map = region->get_map(TYPE_CONTROL);
+					real_t src_ctrl = map->get_pixelv(map_pixel_position).r;
+					int tex_id = (get_blend(src_ctrl) > 110 + int(_brush_data.get("margin", 0))) ? get_overlay(src_ctrl) : get_base(src_ctrl);
+					if (tex_id != asset_id) {
+						can_write = false;
+					}
+				}
+				if (can_write) {
+					switch (_tool) {
+						case COLOR:
+							dest = src.lerp((_operation == ADD) ? color : COLOR_WHITE, brush_alpha * strength);
+							dest.a = src.a;
+							break;
+						case ROUGHNESS:
+							/* Roughness received from UI is -100 to 100. Changed to 0,1 before storing.
+							 * To convert 0,1 back to -100,100 use: 200 * (color.a - 0.5)
+							 * However Godot stores values as 8-bit ints. Roundtrip is = int(a*255)/255.0
+							 * Roughness 0 is saved as 0.5, but retreived is 0.498, or -0.4 roughness
+							 * We round the final amount in tool_settings.gd:_on_picked().
+							 */
+							if (_operation == ADD) {
+								dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * roughness), brush_alpha * strength);
+							} else {
+								dest.a = Math::lerp(real_t(src.a), real_t(.5f + .5f * 0.5f), brush_alpha * strength);
+							}
+							break;
+						default:
+							break;
+					}
 				}
 			}
 			backup_region(region);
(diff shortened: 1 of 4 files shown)


src/terrain_3d_editor.cpp
```diff
@@ -317,8 +317,8 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
 				bool navigation = is_nav(src.r);
 				bool autoshader = is_auto(src.r);
 
-				real_t alpha_clip = (brush_alpha > 0.1f) ? 1.f : 0.f;
-				uint32_t dest_id = uint32_t(Math::lerp(base_id, asset_id, alpha_clip));
+				real_t alpha_clip = (brush_alpha > 0.5f) ? 1.f : 0.f;
+				uint32_t dest_id = asset_id;
```
TokisanGames, line 321: You could just drop dest_id and only use asset_id. 
asset_id should probably be set const to be safe. 
And put it on the right side of the `if (dest_id == base_id)` on L366

TokisanGames:
Ok pretty good. Lots of nice improvements here. There are a couple of odd effects that should be looked at though, but overall It's very close to what we want. I like how the tightening of the alpha clamp makes the brush feel more true to the brush shape.

Strength feels good overall. I want every setting to do something. So 1% should never not do anything, and 100% should be fast, but reasonable. Even 1000+% should be a super max, but still clamped to a not insane number. I recently clamped height/slope because 1000% spit out stupid values. Now they are just faster values than 100% but no more than any sane person would want.

I've drawn out all of the cases to pinpoint what currently happens and what should or not happen. I bolded the expected conditions that should change.

The setting in the left column is the only one selected (none, texture, scale, angle).

### Manually painted area
Large section of **painted** grass w/ a strip of **painted** rock

**Paint**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture|Changes base texture|Ok
|Scale|Changes scale of all textures|Ok, but **filtered by texture is probably better**
|Angle|Changes angle of all textures|Ok, **as above**. See spray scale/angle

**Spray**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture, selected id == base|Decreases blend|Ok
|Texture, selected id != base|Changes overlay texture, increases blend|Ok
|Scale|Changes scale of selected texture|Ok, expected and good
|Angle|Changes angle of selected texture|Ok, expected and good

### Autoshaded area 
**Paint**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture|Disables auto, changes base texture|Ok
|Scale|Changes scale| Unexpected, but ok
|Angle|Changes angle|Unexpected, but ok

**Spray**
|Setting|Result|Expected
|--|--|--
|Nothing w/ selected id==base id|Nothing|Ok
|Nothing w/ selected id!= base id|Disables autoshader|**Unexpected, should do nothing**
|Texture|Disables auto, changes spray texture|Ok
|Scale|Changes scale| Unexpected, but ok
|Angle|Changes angle| Unexpected, but ok

### Border of Manual/Autoshaded
|Setting|Result|Expected
|--|--|--
|Spray Texture @10%| Disables autoshader, sprays, slight halo if base is different|Ok
|Spray Texture @5% | Very slowly disables autoshader, then sprays | **A bit weak**
|Spray Texture @4% | Does nothing to autoshader | **Unexpected problem**

This bottom range for disabling the autoshader should move down 4% so the current 5% is the new 1%.

TokisanGames (changes requested):
Ok pretty good. Lots of nice improvements here. There are a couple of odd effects that should be looked at though, but overall It's very close to what we want. I like how the tightening of the alpha clamp makes the brush feel more true to the brush shape.

Strength feels good overall. I want every setting to do something. So 1% should never not do anything, and 100% should be fast, but reasonable. Even 1000+% should be a super max, but still clamped to a not insane number. I recently clamped height/slope because 1000% spit out stupid values. Now they are just faster values than 100% but no more than any sane person would want.

I've drawn out all of the cases to pinpoint what currently happens and what should or not happen. I bolded the expected conditions that should change.

The setting in the left column is the only one selected (none, texture, scale, angle).

### Manually painted area
Large section of **painted** grass w/ a strip of **painted** rock

**Paint**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture|Changes base texture|Ok
|Scale|Changes scale of all textures|Ok, but **filtered by texture is probably better**
|Angle|Changes angle of all textures|Ok, **as above**. See spray scale/angle

**Spray**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture, selected id == base|Decreases blend|Ok
|Texture, selected id != base|Changes overlay texture, increases blend|Ok
|Scale|Changes scale of selected texture|Ok, expected and good
|Angle|Changes angle of selected texture|Ok, expected and good

### Autoshaded area 
**Paint**
|Setting|Result|Expected
|--|--|--
|Nothing|Nothing|Ok
|Texture|Disables auto, changes base texture|Ok
|Scale|Changes scale| Unexpected, but ok
|Angle|Changes angle|Unexpected, but ok

**Spray**
|Setting|Result|Expected
|--|--|--
|Nothing w/ selected id==base id|Nothing|Ok
|Nothing w/ selected id!= base id|Disables autoshader|**Unexpected, should do nothing**
|Texture|Disables auto, changes spray texture|Ok
|Scale|Changes scale| Unexpected, but ok
|Angle|Changes angle| Unexpected, but ok

### Border of Manual/Autoshaded
|Setting|Result|Expected
|--|--|--
|Spray Texture @10%| Disables autoshader, sprays, slight halo if base is different|Ok
|Spray Texture @5% | Very slowly disables autoshader, then sprays | **A bit weak**
|Spray Texture @4% | Does nothing to autoshader | **Unexpected problem**

This bottom range for disabling the autoshader should move down 4% so the current 5% is the new 1%.

Xtarsia:
Things should behave more consistently now.

the case where spray did nothing at all, was related to this clamp:
``real_t spray_strength = CLAMP(strength * 0.05f, 0.003f, .25f);``

since 1/255 is 0.0039 if the value was between 0.003 and 0.0039 then blend would never accumulate over time.

setting the minimum to 0.004 resolved that.

i've added additional cases to make dealing with the auto shader more consistent too. whilst painting angle/scale seemed ok, it always ended up feeling a bit off to me, since the conditions would be based on the invisible state of the base/over/blend.

TokisanGames:
Added filtering to color and wetness to fix #192. However, we aren't doing a weighted blend on the colormap, so if the color is too contrasted, it will get blocky.
Image
{54629409-E6AC-42C7-ACBC-888CD6CC76A6}

TokisanGames:
Great, thank you!

PR #494, opened 2024-09-17
control map api changes

Xtarsia:
Need some feedback about function names, everything tested and working ok.

this method of taking the original control, and masking before or-ing could be added to the util docs page too.
``(control & ~(0x1F << 27)) | (base >> 27 & 0x1F)``

### DIFF
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -638,26 +638,6 @@ real_t Terrain3DData::get_height(const Vector3 &p_global_position) const {
 	}
 }
 
-real_t Terrain3DData::get_angle(const Vector3 &p_global_position) const {
-	float src = get_pixel(TYPE_CONTROL, p_global_position).r; // Must be 32-bit float, not double/real
-	if (std::isnan(src)) {
-		return NAN;
-	}
-	real_t angle = real_t(get_uv_rotation(src));
-	angle *= 22.5; // Return value in degrees.
-	return angle;
-}
-
-real_t Terrain3DData::get_scale(const Vector3 &p_global_position) const {
-	float src = get_pixel(TYPE_CONTROL, p_global_position).r; // Must be 32-bit float, not double/real
-	if (std::isnan(src)) {
-		return NAN;
-	}
-	std::array<real_t, 8> scale_values = { 0.0f, 20.0f, 40.0f, 60.0f, 80.0f, -60.0f, -40.0f, -20.0f };
-	real_t scale = scale_values[get_uv_scale(src)]; //select from array UI return values
-	return scale;
-}
-
 Vector3 Terrain3DData::get_normal(const Vector3 &p_global_position) const {
 	if (get_region_idp(p_global_position) < 0 || is_hole(get_control(p_global_position))) {
 		return Vector3(NAN, NAN, NAN);
@@ -1132,8 +1112,23 @@ void Terrain3DData::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_control", "global_position"), &Terrain3DData::get_control);
 	ClassDB::bind_method(D_METHOD("set_roughness", "global_position", "roughness"), &Terrain3DData::set_roughness);
 	ClassDB::bind_method(D_METHOD("get_roughness", "global_position"), &Terrain3DData::get_roughness);
-	ClassDB::bind_method(D_METHOD("get_angle", "global_position"), &Terrain3DData::get_angle);
-	ClassDB::bind_method(D_METHOD("get_scale", "global_position"), &Terrain3DData::get_scale);
+
+	ClassDB::bind_method(D_METHOD("set_control_base_id", "global_position", "texture_id"), &Terrain3DData::set_control_base_id);
+	ClassDB::bind_method(D_METHOD("get_control_base_id", "global_position"), &Terrain3DData::get_control_base_id);
+	ClassDB::bind_method(D_METHOD("set_control_overlay_id", "global_position", "texture_id"), &Terrain3DData::set_control_overlay_id);
+	ClassDB::bind_method(D_METHOD("get_control_overlay_id", "global_position"), &Terrain3DData::get_control_overlay_id);
+	ClassDB::bind_method(D_METHOD("set_control_blend", "global_position", "blend_value"), &Terrain3DData::set_control_blend);
+	ClassDB::bind_method(D_METHOD("get_control_blend", "global_position"), &Terrain3DData::get_control_blend);
+	ClassDB::bind_method(D_METHOD("set_control_angle", "global_position", "degrees"), &Terrain3DData::set_control_angle);
+	ClassDB::bind_method(D_METHOD("get_control_angle", "global_position"), &Terrain3DData::get_control_angle);
+	ClassDB::bind_method(D_METHOD("set_control_scale", "global_position", "percentage_modifier"), &Terrain3DData::set_control_scale);
+	ClassDB::bind_method(D_METHOD("get_control_scale", "global_position"), &Terrain3DData::get_control_scale);
+	ClassDB::bind_method(D_METHOD("set_control_hole", "global_position", "enable"), &Terrain3DData::set_control_hole);
+	ClassDB::bind_method(D_METHOD("get_control_hole", "global_position"), &Terrain3DData::get_control_hole);
+	ClassDB::bind_method(D_METHOD("set_control_navigation", "global_position", "enable"), &Terrain3DData::set_control_navigation);
+	ClassDB::bind_method(D_METHOD("get_control_navigation", "global_position"), &Terrain3DData::get_control_navigation);
+	ClassDB::bind_method(D_METHOD("set_control_auto", "global_position", "enable"), &Terrain3DData::set_control_auto);
+	ClassDB::bind_method(D_METHOD("get_control_auto", "global_position"), &Terrain3DData::get_control_auto);
 
 	ClassDB::bind_method(D_METHOD("get_normal", "global_position"), &Terrain3DData::get_normal);
 	ClassDB::bind_method(D_METHOD("get_texture_id", "global_position"), &Terrain3DData::get_texture_id);
--- a/src/terrain_3d_data.h
+++ b/src/terrain_3d_data.h
@@ -147,8 +147,24 @@ public:
 	uint32_t get_control(const Vector3 &p_global_position) const;
 	void set_roughness(const Vector3 &p_global_position, const real_t p_roughness);
 	real_t get_roughness(const Vector3 &p_global_position) const;
-	real_t get_angle(const Vector3 &p_global_position) const;
-	real_t get_scale(const Vector3 &p_global_position) const;
+
+	// Control Map
+	void set_control_base_id(const Vector3 &p_global_position, const uint8_t p_base);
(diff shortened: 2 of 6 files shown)


src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+	set_control(p_global_position, (control & ~(0xFF << 14)) | enc_blend(blend));
+}
+
+inline real_t Terrain3DData::get_control_blend(const Vector3 &p_global_position) const {
+	return real_t((get_control(p_global_position) >> 14 & 0xFF)) / 255.f;
+}
+
+inline void Terrain3DData::set_control_base(const Vector3 &p_global_position, const uint8_t p_base) {
+	uint32_t control = get_control(p_global_position);
+	uint8_t base = uint8_t(CLAMP(p_base, 0.f, 31.f));
```
TokisanGames, line 296: This should be an integer clamp, not float. Maybe the uint8_t should cast the 0 and 31 instead of the result as well, otherwise we might cast twice.

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+	uint8_t uvrotation = uint8_t(CLAMP(Math::round(p_angle / 22.5f), 0.f, 15.f));
+	set_control(p_global_position, (control & ~(0xF << 10)) | enc_uv_rotation(uvrotation));
+}
+
+// Expects scale as a percentage modifier
+inline void Terrain3DData::set_control_scale(const Vector3 &p_global_position, const real_t p_scale) {
+	uint32_t control = get_control(p_global_position);
+	std::array<uint32_t, 8> scale_align = { 5, 6, 7, 0, 1, 2, 3, 4 };
+	uint8_t uvscale = scale_align[uint8_t(CLAMP(Math::round((p_scale + 60.f) / 20.f), 0.f, 7.f))];
+	set_control(p_global_position, (control & ~(0x7 << 7)) | enc_uv_scale(uvscale));
```
TokisanGames, line 280: According to the docs this should be `(x & 0x7) <<6`. 7 matches util.h. So the docs are wrong.

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+
+// Expects 0.0 to 1.0 range
+inline void Terrain3DData::set_control_blend(const Vector3 &p_global_position, const real_t p_blend) {
+	uint32_t control = get_control(p_global_position);
+	uint8_t blend = uint8_t(CLAMP(Math::round(CLAMP(1.f - p_blend, 0.f, 1.f) * 255.f), 0.f, 255.f));
+	set_control(p_global_position, (control & ~(0xFF << 14)) | enc_blend(blend));
+}
+
+inline real_t Terrain3DData::get_control_blend(const Vector3 &p_global_position) const {
+	return real_t((get_control(p_global_position) >> 14 & 0xFF)) / 255.f;
```
TokisanGames, line 291: Out of bounds, get_control will return UINT32_MAX. This will return 1.0. NAN might be better.

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
 	return (std::isnan(val)) ? UINT32_MAX : as_uint(val);
 }
 
+// Expects angle in degrees
+inline void Terrain3DData::set_control_angle(const Vector3 &p_global_position, const real_t p_angle) {
+	uint32_t control = get_control(p_global_position);
+	uint8_t uvrotation = uint8_t(CLAMP(Math::round(p_angle / 22.5f), 0.f, 15.f));
+	set_control(p_global_position, (control & ~(0xF << 10)) | enc_uv_rotation(uvrotation));
+}
```
TokisanGames, line 273: If global_position is out of bounds, get_control will return UINT32_MAX, set_pixel will error with "no region". That's fine.

src/terrain_3d_data.cpp
```diff
@@ -622,7 +622,7 @@ real_t Terrain3DData::get_angle(const Vector3 &p_global_position) const {
 	return angle;
 }
 
-real_t Terrain3DData::get_scale(const Vector3 &p_global_position) const {
+real_t Terrain3DData::get_control_scale(const Vector3 &p_global_position) const {
```
TokisanGames, line 625: get_control_scale/angle could be moved to the inline functions so they're all together. Also I note that these two are written in a different style than the others (eg using get_pixel instead of get_control, which is inlined but has extra stuff)

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+	set_control(p_global_position, (control & ~(0x1F << 27)) | enc_base(base));
+}
+
+inline uint8_t Terrain3DData::get_control_base(const Vector3 &p_global_position) const {
+	return uint8_t((get_control(p_global_position) >> 27 & 0x1F));
+}
+
+inline void Terrain3DData::set_control_overlay(const Vector3 &p_global_position, const uint8_t p_overlay) {
+	uint32_t control = get_control(p_global_position);
+	uint8_t overlay = uint8_t(CLAMP(p_overlay, 0.f, 31.f));
```
TokisanGames, line 306: integer clamp

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+	return (get_control(p_global_position) >> 2 & 0x1) == 1;
+}
+
+inline void Terrain3DData::set_control_navigation(const Vector3 &p_global_position, const bool p_navigation) {
+	uint32_t control = get_control(p_global_position);
+	set_control(p_global_position, (control & ~(0x1 << 1)) | enc_nav(p_navigation));
+}
+
+inline bool Terrain3DData::get_control_navigation(const Vector3 &p_global_position) const {
+	return (get_control(p_global_position) >> 1 & 0x1) == 1;
```
TokisanGames, line 338: Hmm, what do we do about out of bounds for these? False I guess.

TokisanGames: Let's reorder these the same as bit order or descending as is here. Hole, navigation, auto.

src/terrain_3d_data.h
```diff
@@ -249,6 +265,79 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+	uint32_t control = get_control(p_global_position);
+	uint8_t overlay = uint8_t(CLAMP(p_overlay, 0.f, 31.f));
+	set_control(p_global_position, (control & ~(0x1F << 22)) | enc_base(overlay));
+}
+
+inline uint8_t Terrain3DData::get_control_overlay(const Vector3 &p_global_position) const {
+	return uint8_t((get_control(p_global_position) >> 22 & 0x1F));
+}
+
+inline void Terrain3DData::set_control_auto(const Vector3 &p_global_position, const bool p_auto) {
```
TokisanGames, line 314: Let's reorder these in the same order as the control map format doc; descending bit order. Set before get. In the declaration and definition.

src/terrain_3d_data.cpp
```diff
@@ -1106,8 +1106,23 @@ void Terrain3DData::_bind_methods() {
...
+
+	ClassDB::bind_method(D_METHOD("set_control_angle", "global_position", "degrees"), &Terrain3DData::set_control_angle);
+	ClassDB::bind_method(D_METHOD("get_control_angle", "global_position"), &Terrain3DData::get_control_angle);
+	ClassDB::bind_method(D_METHOD("set_control_scale", "global_position", "percentage_modifier"), &Terrain3DData::set_control_scale);
+	ClassDB::bind_method(D_METHOD("get_control_scale", "global_position"), &Terrain3DData::get_control_scale);
+	ClassDB::bind_method(D_METHOD("set_control_blend", "global_position", "value"), &Terrain3DData::set_control_blend);
+	ClassDB::bind_method(D_METHOD("get_control_blend", "global_position"), &Terrain3DData::get_control_blend);
+	ClassDB::bind_method(D_METHOD("set_control_base", "global_position", "id"), &Terrain3DData::set_control_base);
+	ClassDB::bind_method(D_METHOD("get_control_base", "global_position"), &Terrain3DData::get_control_base);
+	ClassDB::bind_method(D_METHOD("set_control_overlay", "global_position", "id"), &Terrain3DData::set_control_overlay);
```
TokisanGames, line 1118: How about texture_id instead. And blend_value or blend instead of just value.

TokisanGames:
Thank you. The API is fine. If you rebase it, I can generate the XML and add in a commit so you can at least put in the return values. I'll fill out the rest.

TokisanGames (changes requested):
.

src/terrain_3d_data.h
TokisanGames: This still isn't correct. Out of bounds we get UINT32_MAX.
UINT32_MAX >> 1 = 0111 1111 1111 1111... 1111 & 0001 == 1

TokisanGames:
Auto/nav/hole need another adjustment. I added XML. You could fill out the description or at least return values and I'll wrap up anything left.

src/terrain_3d_data.h
Xtarsia: Ah, I should add the same ternery for these then too.
uint32_t control = get_control(p_global_position);
return control == UINT32_MAX ? false : is_auto(control);

though.. when we are useing all control bits, is it not possible that UINT32_MAX is the correct stored value in some situations?

src/terrain_3d_data.h
```diff
@@ -250,6 +266,96 @@ inline uint32_t Terrain3DData::get_control(const Vector3 &p_global_position) con
...
+
+inline uint8_t Terrain3DData::get_control_overlay_id(const Vector3 &p_global_position) const {
+	uint32_t control = get_control(p_global_position);
+	return control == UINT32_MAX ? UINT32_MAX : get_overlay(control);
+}
+
+// Expects 0.0 to 1.0 range
+inline void Terrain3DData::set_control_blend(const Vector3 &p_global_position, const real_t p_blend) {
+	uint32_t control = get_control(p_global_position);
+	uint8_t blend = uint8_t(CLAMP(Math::round(CLAMP(p_blend, 0.f, 1.f) * 255.f), 0.f, 255.f));
```
TokisanGames, line 294: Aren't these clamps redundant? I don't think there's any value p_blend could be where the inner clamp makes a difference.

Xtarsia: Yep, you're right.

src/terrain_3d_data.h
TokisanGames: Technically possible, but that means there's a hole there. It also means the the autoshader is on and overlay is 100% blended and base id is hidden. So, we could disallow base and overlay to be 31 when blend is 1. Base could reset to id 0 if it becomes a problem.

Xtarsia:
Hopefully ready for @TokisanGames to apply that final bit of polish now..

TokisanGames:
Very good, thank you.

PR #495, opened 2024-09-18
Arraymesh Generator sets scene root as owner

### LINKED ISSUE #480: Arraymesh generator doesn't set owner in some cases
### Description

It seems that if Terrain3D is the scene root, it won't set the owner properly. It will if it's a child node. 

https://discord.com/channels/691957978680786944/1130291534802202735/1280604554316218461

Xtarsia:
closes #480

image

### DIFF
--- a/project/addons/terrain_3d/src/baker.gd
+++ b/project/addons/terrain_3d/src/baker.gd
@@ -48,6 +48,9 @@ func bake_mesh_popup() -> void:
 
 
 func _bake_mesh() -> void:
+	if plugin.terrain.data.get_region_count() == 0:
+		push_error("Terrain3D has no active regions to bake")
+		return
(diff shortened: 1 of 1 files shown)


TokisanGames:
Thank you. Can you please also add these to the PR:
* Fix the same thing for Bake Occluder3D and Setup Navigation (which reparents Terrain3D to a nav node)
* When there are no regions, avoid Godot's errors below, and print our own clear non-error message
```
Bake ArrayMesh
ERROR: Condition "!(format & Mesh::ARRAY_FORMAT_TEX_UV)" is true.
   at: generate_tangents (scene/resources/surface_tool.cpp:1151)
ERROR: Condition "index_array.size() == 0" is true.
   at: optimize_indices_for_cache (scene/resources/surface_tool.cpp:1283)

Bake Occluder has the above plus
SCRIPT ERROR: Assertion failed.
          at: _bake_occluder (res://addons/terrain_3d/src/baker.gd:97)
```

PR #500, opened 2024-09-25
Compatibility renderer support

### LINKED ISSUE #217: Support Compatibility Renderer
Switch to compatibility mode and the textures are black. Supposedly texture arrays are not supported. However height and the colormap are there, so some texture arrays are being read. So it's unclear what is supported exactly.

Also opengl doesn't support FMA, so a macro should be made to replace fma with `*+` via code or have two separate shaders.

Waiting for OpenGL support in Godot:
* Tracker https://github.com/godotengine/godot/issues/66458
* TextureArray support https://github.com/godotengine/godot/issues/71029 
* TextureArray PR https://github.com/godotengine/godot/pull/80363

Xtarsia:
Admin edit:
Fixes #217

* Enables compatibility support
* Modifies custom shaders w/ special defines
* Improves GPU mouse and shader; supports compatibility mode
* Fixes not-normalized normals for all renderers
* Hides the region border normal fix outside of regions (was noticable at small region sizes)

----

Detects whether Godot is using the compatibility renderer.

If so,
The generated shader will have a set of #defines added that allow full support for the standard shader.
The shader override will have the #defines visible at the begining on the shader, before shader_type.

The editor will inject code at the begining, and the end either the standard shader, or any provided override shader
in order to draw a "decal" on the terrain.

in the interest of simplicity, the gdscript changes are as small as possible, tho some things are a bit inneficient,
performace is fine and its only during editing anyways.

~~The slope tool doesnt have propper decal support.~~

adding terrain3d.get_compatibility() seems a bit dubious to me, but saved repeating a ton of code in gdscript, and ensures that the editor is always on the same page.

Finally, there are no warnings at the moment when the albedo and normal arrays are in compressed mode. for 4.3 this is fixed anyways, but for 4.2 all textures will be black. The workaround is to re-import all textures as "Lossless". Resizing to half resolution results in the same VRAM use for Lossless vs Compressed at the original size.

EDIT: recently updated test version is available: https://github.com/Xtarsia/Terrain3D/actions/runs/11079134556

### DIFF
--- a/src/shaders/gpu_depth.glsl
+++ b/src/shaders/gpu_depth.glsl
@@ -7,28 +7,33 @@ R"(
 shader_type spatial;
 render_mode unshaded;
 
-uniform sampler2D depth_texture : source_color, hint_depth_texture, filter_nearest, repeat_disable;
+uniform highp sampler2D depth_texture : source_color, hint_depth_texture, filter_nearest, repeat_disable;
 
 uniform float camera_far = 100000.0;
-
-// Mobile renderer HDR mode has limit of 1 or 2. Pack full range depth to RG
-// https://gamedev.stackexchange.com/questions/201151/24bit-float-to-rgb
-vec3 encode_rg(float value) {
-    vec2 kEncodeMul = vec2(1.0, 255.0);
-    float kEncodeBit = 1.0 / 255.0;
-    vec2 color = kEncodeMul * value / camera_far;
-    color = fract(color);
-    color.x -= color.y * kEncodeBit;
-	return vec3(color, 0.);
-}
+uniform bool compatibility = false;
 	
 void fragment() {
 	float depth = textureLod(depth_texture, SCREEN_UV, 0.).x;
+	if (compatibility) {
+	 depth = depth * 2.0 - 1.0;
+	}
 	vec3 ndc = vec3(SCREEN_UV * 2.0 - 1.0, depth);
 	vec4 view = INV_PROJECTION_MATRIX * vec4(ndc, 1.0);
 	view.xyz /= view.w;
 	float depth_linear = -view.z;
-	ALBEDO = encode_rg(depth_linear); // Encoded value for Mobile
+
+	// Normalize depth to the range 0 - 1
+	highp float scaledDepth = clamp(depth_linear / 100000.0, 0.0, 1.0);
+
+	// Encode using 127 steps, which map to the 128 - 255 range.
+	// Avoids precision loss for compatability and mobile renderer
+	// 21bit depth value
+	highp float r = (floor(scaledDepth * 127.0) + 128.0) / 255.0;
+	highp float g = (floor(fract(scaledDepth * 127.0) * 127.0) + 128.0) / 255.0;
+	highp float b = (floor(fract(scaledDepth * 127.0 * 127.0) * 127.0) + 128.0) / 255.0;
+		
+	ALBEDO = vec3(r, g, b); // Return encoded value, required for Mobile & Compatibility renderers
+
 }
 
 )"
--- a/src/terrain_3d.h
+++ b/src/terrain_3d.h
@@ -96,6 +96,7 @@ private:
 	RenderingServer::ShadowCastingSetting _cast_shadows = RenderingServer::SHADOW_CASTING_SETTING_ON;
 	GeometryInstance3D::GIMode _gi_mode = GeometryInstance3D::GI_MODE_STATIC;
 	real_t _cull_margin = 0.0f;
+	bool _compatibility = false;
 
 	// Mouse cursor
 	SubViewport *_mouse_vp = nullptr;
@@ -208,6 +209,7 @@ public:
 	GeometryInstance3D::GIMode get_gi_mode() const { return _gi_mode; }
 	void set_cull_margin(const real_t p_margin);
 	real_t get_cull_margin() const { return _cull_margin; };
+	bool is_compatibility_mode() const { return _compatibility; };
 
 	// Processing
 	void snap(const Vector3 &p_cam_pos);
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -5,10 +5,10 @@
 #include <godot_cpp/classes/gradient.hpp>
 #include <godot_cpp/classes/image_texture.hpp>
 #include <godot_cpp/classes/noise_texture2d.hpp>
-#include <godot_cpp/classes/rendering_server.hpp>
-#include <godot_cpp/classes/resource_saver.hpp>
 #include <godot_cpp/classes/reg_ex.hpp>
 #include <godot_cpp/classes/reg_ex_match.hpp>
+#include <godot_cpp/classes/rendering_server.hpp>
+#include <godot_cpp/classes/resource_saver.hpp>
 
 #include "logger.h"
 #include "terrain_3d_material.h"
@@ -154,13 +154,13 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 	Ref<RegEx> regex;
 	Ref<RegExMatch> match;
 	regex.instantiate();
-	int idx;
-	// Insert after render_mode ;
-	insert_names.clear();
-	regex->compile("render_mode[^;]*;");
+
+	// Insert after render_mode
+	regex->compile("render_mode.*;?");
 	match = regex->search(shader);
-	idx = match.is_valid() ? match->get_end() : -1;
+	int idx = match.is_valid() ? match->get_end() : -1;
 	if (idx < 0) {
+		LOG(DEBUG, "No render mode; cannot inject editor code");
 		return shader;
 	}
 	if (_compatibility) {
@@ -173,11 +173,13 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 		shader = shader.insert(idx, "\n\n" + insert);
 		idx += insert.length();
 	}
+
 	// Insert before vertex()
 	regex->compile("void\\s+vertex\\s*\\(");
 	match = regex->search(shader);
 	idx = match.is_valid() ? match->get_start() - 1 : -1;
 	if (idx < 0) {
+		LOG(DEBUG, "No void vertex(); cannot inject editor code");
 		return shader;
 	}
 	insert_names.clear();
@@ -189,9 +191,11 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 		shader = shader.insert(idx, "\n" + insert);
 		idx += insert.length();
 	}
-	// Inserted at the end of the shader
+
+	// Insert at the end of the shader, before the end of `fragment(){ }`
 	idx = shader.rfind("}");
 	if (idx < 0) {
+		LOG(DEBUG, "No ending bracket; cannot inject editor code");
 		return shader;
 	}
 	insert_names.clear();
@@ -265,11 +269,11 @@ void Terrain3DMaterial::_update_shader() {
 		if (_shader_override->get_code().is_empty()) {
 			if (_compatibility) {
 				code = _generate_shader_code();
-				// Insert after render_mode ;
-				regex->compile("render_mode[^;]*;");
+				// Insert after render_mode
+				regex->compile("render_mode.*;?");
 				match = regex->search(code);
 				if (match.is_valid()) {
-					_shader_override->set_code(code.insert(match->get_end(), "\n\n" + String(_shader_code["EDITOR_COMPATIBILITY_DEFINES"])));
+					_shader_override->set_code(code.insert(match->get_end(), "\n\n" + String(_shader_code["EDITOR_COMPATIBILITY_DEFINES"]).strip_edges()));
 				}
 			} else {
 				_shader_override->set_code(_generate_shader_code());
@@ -278,11 +282,12 @@ void Terrain3DMaterial::_update_shader() {
 		code = _shader_override->get_code();
 		if (_compatibility) {
 			if (!code.contains("COMPATIBILITY_DEFINES")) {
-				// Insert after render_mode ;
-				regex->compile("render_mode[^;]*;");
+				// Insert after render_mode
+				regex->compile("render_mode.*;?");
 				match = regex->search(code);
 				if (match.is_valid()) {
-					_shader_override->set_code(code.insert(match->get_end(), "\n\n" + String(_shader_code["EDITOR_COMPATIBILITY_DEFINES"])));
+					_shader_override->set_code(code.insert(match->get_end(), "\n\n" + String(_shader_code["EDITOR_COMPATIBILITY_DEFINES"]).strip_edges()));
+					LOG(WARN, "Inserting compatibility defines into your shader. Save to finalize");
 					code = _shader_override->get_code();
 				}
 			}
@@ -472,12 +477,11 @@ void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
 				int chars = code.find("// END_COMPAT_DEFINES") - idx;
 				code = code.erase(idx, chars + 21); // + length of "// END_COMPAT_DEFINES"
 				_shader_override->set_code(code);
-				LOG(WARN, "Compatibility Renderer not detected: removed COMPATIBILITY_DEFINES in override shader");
+				LOG(WARN, "Compatibility renderer not detected. Removed COMPATIBILITY_DEFINES in override shader. Save to finalize");
 				_update_shader();
 			}
 		}
 	}
-
 }
 
 Terrain3DMaterial::~Terrain3DMaterial() {
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -7,6 +7,8 @@
 #include <godot_cpp/classes/noise_texture2d.hpp>
 #include <godot_cpp/classes/rendering_server.hpp>
 #include <godot_cpp/classes/resource_saver.hpp>
+#include <godot_cpp/classes/reg_ex.hpp>
+#include <godot_cpp/classes/reg_ex_match.hpp>
 
 #include "logger.h"
 #include "terrain_3d_material.h"
@@ -18,7 +20,6 @@
 
 void Terrain3DMaterial::_preload_shaders() {
 	// Preprocessor loading of external shader inserts
-
 	_parse_shader(
 #include "shaders/uniforms.glsl"
 			, "uniforms");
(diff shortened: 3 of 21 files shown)


src/terrain_3d_material.cpp
```diff
@@ -381,14 +401,15 @@ void Terrain3DMaterial::_set_shader_parameters(const Dictionary &p_dict) {
...
-void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
+void Terrain3DMaterial::initialize(Terrain3D *p_terrain, const bool p_compatibility) {
 	if (p_terrain != nullptr) {
 		_terrain = p_terrain;
 	} else {
 		LOG(ERROR, "Initialization failed, p_terrain is null");
 		return;
 	}
 	LOG(INFO, "Initializing material");
+	_compatibility = p_compatibility;
```
TokisanGames, line 412: Since get_compatibility is public, let's drop the parameter and do this:
`_compatibility = _terrian->get_compatibility()`

src/terrain_3d.h
```diff
@@ -208,6 +209,7 @@ class Terrain3D : public Node3D {
 	GeometryInstance3D::GIMode get_gi_mode() const { return _gi_mode; }
 	void set_cull_margin(const real_t p_margin);
 	real_t get_cull_margin() const { return _cull_margin; };
+	bool get_compatibility() const { return _compatibility; };
```
TokisanGames, line 212: Maybe `is_compatibility` is more consistent

src/terrain_3d.h
```diff
@@ -48,6 +48,7 @@ class Terrain3D : public Node3D {
 	String _data_directory;
 	bool _is_inside_world = false;
 	bool _initialized = false;
+	bool _compatibility = false;
```
TokisanGames, line 51: This should probably be in the `//Rendering` section

src/shaders/compatibility.glsl
```diff
@@ -0,0 +1,48 @@
...
+
+//INSERT: EDITOR_SETUP_DECAL
+uniform sampler2D _editor_decal : source_color, filter_linear, repeat_disable;
+uniform vec2 _editor_decal_position;
+uniform float _editor_decal_size;
+uniform float _editor_decal_rotation;
+uniform vec3 _editor_decal_color : source_color;
+uniform bool _editor_decal_visible = false;
+
+// expects UV.
```
TokisanGames, line 27: Does this mean UV (0-region_size in region space), not UV2 (0-1 in local space), and not world space? We should define these in the shader design md doc so I'm not confused again.

src/shaders/gpu_depth.glsl
```diff
@@ -10,25 +10,30 @@ render_mode unshaded;
 uniform sampler2D depth_texture : source_color, hint_depth_texture, filter_nearest, repeat_disable;
 
 uniform float camera_far = 100000.0;
-
-// Mobile renderer HDR mode has limit of 1 or 2. Pack full range depth to RG
```
TokisanGames, line 14: This first part about the mobile HDR mode is still true and good to keep somewhere, maybe where the depth texture is actually read. I don't know why I wrote 1 or 2 like I didn't know :/  . I'm sure there's a similar limit for compatibility.

src/shaders/gpu_depth.glsl
```diff
@@ -10,25 +10,30 @@ render_mode unshaded;
...
+
+	// Encode using 127 steps, which map to the 128 - 255 range.
+	// Avoids precision loss for compatability and mobile renderer
+	// 21bit depth value
+	highp float r = (floor(scaledDepth * 127.0) + 128.0) / 255.0;
+	highp float g = (floor(fract(scaledDepth * 127.0) * 127.0) + 128.0) / 255.0;
+	highp float b = (floor(fract(scaledDepth * 127.0 * 127.0) * 127.0) + 128.0) / 255.0;
+		
+	ALBEDO = vec3(r, g, b); // Return encoded value for Mobile / Compatibility
+
```
TokisanGames, line 36: In my notes I have these options:
* `ALBEDO = vec3(depth_linear); // Absolute value for Forward+ HDR` since the HDR viewport can have any 16-bit value
* The encoded version to fit it within the mobile HDR range
* And this to get the world position instead of depth:
```
	float depth = textureLod(depth_texture, SCREEN_UV, 0.).x;
	vec3 ndc = vec3(SCREEN_UV * 2.0 - 1.0, depth);
	vec4 world = CAMERA * INV_PROJECTION_MATRIX * vec4(ndc, 1.0);
	vec3 world_position = world.xyz / world.w;	
	ALBEDO = world_position;
```

src/terrain_3d_material.cpp
```diff
@@ -149,6 +154,7 @@ String Terrain3DMaterial::_generate_shader_code() const {
 
 String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 	String shader = p_shader;
+	// inserted at the end of the shader
```
TokisanGames, line 157: It's a very small, nitpicking thing, but I'd prefer comments start with a capital. I usually leave off the periods if it's one line, but use them if it's a paragraph. It's so small, but it reminds me of Steve Jobs demanding the inside of macs and iphones look nice too. On an open source project, we're literally building a tool for people to look at the inside of, and we are building a beautiful tool. The work you've done here is fantastic and appreciated by everyone.

src/terrain_3d_material.cpp
```diff
@@ -202,11 +208,25 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
...
+	idx = p_shader.find(";") + 1;
+	insert_names.clear();
+	if (_compatibility) {
+		insert_names.push_back("EDITOR_SETUP_DECAL");
+	}
+	for (int i = 0; i < insert_names.size(); i++) {
+		String insert = _shader_code[insert_names[i]];
+		shader = shader.insert(idx, "\n" + insert);
+		idx += insert.length();
+	}
```
TokisanGames, line 229: Maybe this beginning section should go before the end section? 

Doesn't this find `shader_type spatial;` and you probably want it after `render_mode blend_mix,depth_draw_opaque,cull_back,diffuse_burley,specular_schlick_ggx,skip_vertex_transform;`?
`p_shader.find(";", p_shader.find(";") + 1) + 1`?

Though it could come after `*/` since we should always have a /* */ comment at the top. And our comments above the raw string in main.glsl should note that the first */ and last } in fragment are sentinels.

src/shaders/compatibility.glsl
```diff
@@ -0,0 +1,48 @@
...
+#define textureQueryLod(a, b) vec4(0.0)
+#define texelOffset(b, c) ivec3(ivec2(b.xy * _region_size + c -0.4979), int(b.z))
+#define textureGather(a, b) vec4( \
+    texelFetch(a, texelOffset(b, vec2(0,1)), 0).r, \
+    texelFetch(a, texelOffset(b, vec2(1,1)), 0).r, \
+    texelFetch(a, texelOffset(b, vec2(1,0)), 0).r, \
+    texelFetch(a, texelOffset(b, vec2(0,0)), 0).r  \
+)
+
+//INSERT: EDITOR_SETUP_DECAL
```
TokisanGames, line 19: Though this is only for compatibility now, we're eventually going to use it for everything. And it is an editor function inserted differently than compatibility defines above. So maybe this block should be placed in `editor_functions.glsl` now.

TokisanGames (changes requested):
Very nice. This is a good one.  ~~to be our 500th PR!~~ (_edit_: oh # 500 means our 500th document including issues and discussions. This is our... 144th PR)

I added a note that it will fix a ticket.

Please also try exporting the demo to webgl. I reviewed code but haven't tested it yet. I'll also test the export.

You can also add some notes to platform.md that it is now supported, and what they need to do for 4.2 and 4.3. And any notes for WebGL as well.

I thought you said 4.2 must use '(vram) uncompressed' before. Here you say 'Lossless'. Do both work? Whichever does, make sure to get the exact name for the documentation and our discord guidance.

src/shaders/compatibility.glsl
Xtarsia: clarified this comment, its actually texture uv in world space.

src/shaders/compatibility.glsl
Xtarsia: makes sense, moved EDITOR_SETUP_DECAL to editor_functions.glsl

src/shaders/gpu_depth.glsl
Xtarsia: If i remember rightly, Forward+ is 16bit, mobile is 10bit and compatibility is 8bit.

however compatibility has a strange color space conversion that trims blacks, and causes some values between 8 and 127 to sometimes be rounded up or down to the nearest even number, which was causeing havok with the depth encode.

this didnt occur in the top half of the range. Thankfully, using effectivly 7 bits of all 3 color channels is still more than enough, and works perfectly for all 3 render modes.

comment updated with some more info!

src/shaders/gpu_depth.glsl
Xtarsia: I thought about encoding position, but there just isnt enough bits to hold xyz, it could be possible to pass just the y component instead.

src/terrain_3d.h
Xtarsia: I thought about that, but decided not to initially as all the variables in that section are represented in the inspector.

i'll leave that decision to you.

src/terrain_3d_material.cpp
Xtarsia: I'm all for correcting any typos / grammar / spelling / formatting etc. iv'e missed!

src/terrain_3d_material.cpp
Xtarsia: I've swapped the order round, and changed to insert just before vertex(). inserting too early meant no access to uniforms or functions from later on in the shader!

src/terrain_3d.h
TokisanGames: Yes, please move it down. You categorized the function there as well

Xtarsia:
Im unable to get web export working, i suspect due to needing custom export template build for terrain 3D, or something along those lines.

As for textures import settings for 4.2, both options work.

I'll add some initial things to platform.md in the mean time.

TokisanGames:
* In compatibility mode, the minimum.gdshader complains about fma, even though it isn't connected to anything. I had to comment it out.

* Here are my notes on getting webgl working. It isn't successful yet.

### Docs
Building instructions
https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_web.html

Godot Export for web docs
https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_web.html

### Building

To build the extension for web you must install emscripten, set up the build environment, then build the web assembly version. I installed with scoop, which is how my python is also installed, per the Godot building instructions. 

```
$ scoop install emscripten
# then I restarted my terminal to get the updated path to finish installing emscripten

$ emsdk install latest
$ emsdk activate latest

# finally setup the build environment
$ source ~/scoop/apps/emscripten/current/emsdk_env.sh
$ scons platform=web target=template_debug

$ ls project/addons/terrain_3d/bin/*.wasm
project/addons/terrain_3d/bin/libterrain.web.debug.wasm32.wasm
```

### Exporting

* Select [custom dlink unthreaded template](https://forum.godotengine.org/t/gdextension-not-loading-in-web-export/81867/5), which I found here: `C:/Users/Cory/AppData/Roaming/Godot/export_templates/4.3.stable/web_dlink_nothreads_debug.zip`
* Vram compressed textures don't work for me on Edge. They don't load in the log. They stop complaining with VRam uncompressed.
* Export and web load reports this message, but [tickets have said](https://github.com/godotengine/webrtc-native/issues/136#issuecomment-1918640657) we can ignore that.
```
ERROR: No suitable library found for GDExtension: res://addons/terrain_3d/terrain.gdextension. Possible feature flags fe
   at: (./editor/plugins/gdextension_export_plugin.h:129)
```

### Upload to Webserver
After export, the full package must be uploaded to a webserver. Itch or your own. That server must be configured with SharedArrayBuffer & Cross Origin Isolation.
In nginx this can be enabled with:
```
    location ^~ /terrain3d/ {
        add_header "Cross-Origin-Opener-Policy" "same-origin";
        add_header "Cross-Origin-Embedder-Policy" "require-corp";
    }

```
On Itch.io , edit game, find the Embed Options section and enable:

`SharedArrayBuffer support - (Experimental) This may break parts of the page or your project. Only enable if you know you need it.`


### Running

I built w/ 4.2.2 and seemed to mostly load but error messages complained that my scene files were built for 4.3?? So I built and exported everything with 4.3: godot source, godot-cpp, Godot binary, Godot export templates.

Edge complained about being unable to load the textures if compressed with vram compressed. But no complains using vram uncompressed.

Currently it loads to the godot logo, then a grey screen. The javascript console reports this:
{047F5A68-6253-4B55-B5FD-8F977D9C0DCB}

TokisanGames:
* Nice work. Aside from webgl, it works well. It even stutters just like I remember in Godot 3. :/

* The lack of visual indication for the slope tool is a problem. Drawable is not intuitive without it. If you don't already know how it works it will be easy to be confused by it. What can we do about this?

* Figuring out how to build for web won't hold this PR or release up.

* It renders very differently. Brighter, stronger normals, half the speed.

image

src/terrain_3d_material.cpp
```diff
@@ -149,11 +148,39 @@ String Terrain3DMaterial::_generate_shader_code() const {
...
+	if (_compatibility) {
+		insert_names.push_back("EDITOR_COMPATIBILITY_DEFINES");
+	}
+	for (int i = 0; i < insert_names.size(); i++) {
+		String insert = _shader_code[insert_names[i]];
+		shader = shader.insert(idx, "\n" + insert + "\n");
+		idx += insert.length();
+	}
+	// Insert before vertex()
+	idx = shader.find("void vertex() {") - 1;
```
TokisanGames, line 165: They could code like a neanderthal and write: `void  vertex ( )  {` or worse:
```
void vertex()
{
// do cave man shading
}
```
Probably best to go with regex:
``` 
var regex = RegEx.new()
regex.compile("void\\s+vertex\\s*\\(")
var result = regex.search(shader)
if result:
	print(result.get_start()) 
```
https://docs.godotengine.org/en/stable/classes/class_regex.html#class-regex

Xtarsia: I think adding "//SHADER_INJECTION_POINT_VERTEX // do not remove this line. It must be directly before the vertex() function!"

and inserting there would be most straight forwards.

Xtarsia: the regex wasnt too bad to add, its cleaner from the user side as well.

src/terrain_3d_material.cpp
```diff
@@ -149,11 +150,44 @@ String Terrain3DMaterial::_generate_shader_code() const {
...
+			insert_names.push_back("EDITOR_COMPATIBILITY_DEFINES");
+		}
+	}
+	for (int i = 0; i < insert_names.size(); i++) {
+		String insert = _shader_code[insert_names[i]];
+		shader = shader.insert(idx, "\n" + insert + "\n");
+		idx += insert.length();
+	}
+	// Insert before vertex()
+	RegEx regex;
```
TokisanGames, line 169: This line crashes Godot for me in 4.3 with a console error:
```
ERROR: BUG: Godot Object created without binding callbacks. Did you forget to use memnew()?
   at: (godot-cpp\src\classes\wrapped.cpp:87)
```
This works:
```
	Ref<RegEx> regex;
	regex.instantiate();
	regex->compile("void\\s+vertex\\s*\\(\\s*\\)[\\s\\n]*\\{");
	Ref<RegExMatch> match = regex->search(shader);
```

src/terrain_3d_material.cpp
```diff
@@ -149,11 +150,44 @@ String Terrain3DMaterial::_generate_shader_code() const {
...
 		return shader;
 	}
-	Array insert_names;
+	insert_names.clear();
+	if (_compatibility) {
+		insert_names.push_back("EDITOR_SETUP_DECAL");
+	}
+	for (int i = 0; i < insert_names.size(); i++) {
+		String insert = _shader_code[insert_names[i]];
+		shader = shader.insert(idx - 1, "\n" + insert);
```
TokisanGames, line 182: idx is already -1 from L172. This is now -2 from "void vertex", when I think it should just be 0. Insert "\nmystuff\n\n" at the "void vertex" 0 position. My original debug view inclusion of inserting -1 before "}" is probably wrong.

src/terrain_3d_material.cpp
```diff
@@ -216,7 +253,16 @@ void Terrain3DMaterial::_update_shader() {
 	String code;
 	if (_shader_override_enabled && _shader_override.is_valid()) {
 		if (_shader_override->get_code().is_empty()) {
-			_shader_override->set_code(_generate_shader_code());
+			if (_compatibility) {
+				_shader_override->set_code(_generate_shader_code().insert(0, _shader_code["EDITOR_COMPATIBILITY_DEFINES"]));
```
TokisanGames, line 257: I inserted the minimum shader into the override slot and it dumps an error, then inserts shader defines into my file, above `shader type`. I think it should be after render_mode.

Xtarsia:
I think I need to tidy up the regex stuff a bit still.

TokisanGames:
New url. Back tick \` will also capture the mouse, and arrow keys will turn the camera.
https://tokisan.com/terrain3d_demo/demo.html

It seems that all the samplers need highp. Including those in uniforms.glsl.

Here's the full console log, after adding additional highps:
[console_log.txt](https://github.com/user-attachments/files/17179088/console_log.txt)


The last remaining shader error is:
```
demo.js:474  ERROR: 0:528: '[' : array index for samplers must be constant integral expressions
also 535, 536.
```

These lines are:
```
demo.js:459 528: 			vec2 m_decal_uv=(((vec2(((m_cosa * m_uv.x) - (m_sina * m_uv.y)), ((m_sina * m_uv.x) + (m_cosa * m_uv.y))) - (vec2(((m_cosa * m_dus_editor_decal_position[m_i].x) - (m_sina * m_dus_editor_decal_position[m_i].y)), ((m_sina * m_dus_editor_decal_position[m_i].x) + (m_cosa * m_dus_editor_decal_position[m_i].y))) * m_dus_vertex_density)) * m_size) * m_dus_vertex_spacing);

demo.js:459 535: 			float m_decal=smoothstep(0.10000000149012, 1.0, texture(m_dus_editor_decal[m_i], (m_decal_uv + 0.5)).r);
demo.js:459 536: 			m_albedo=mix(m_albedo, m_dus_editor_decal_color[m_i].rgb, (m_decal * m_dus_editor_decal_color[m_i].a));
```

I'm not sure why webgl doesn't like these array indices compared to our other array indices.

> texture(m_dus_editor_decal[m_i], (m_decal_uv + 0.5)).r`
I think this should be `texture(_editor_decal, vec3(decal_uv + 0.5, float(i)))`. In my own custom shader, accessing an array index this way was working, but when I put this into the glsl file and compiled, Godot reported:
`SHADER ERROR: Invalid arguments for the built-in function: "texture(sampler2D[3],vec3)".`

src/terrain_3d_material.cpp
Xtarsia: the 1 \n before is warrented, and the -1 aswell to ensure the insert happens before the } as without it, it will insert after the closing brace.

the inserts include a new line anyways.

idx is assigned to the new find value rather than accumulating now anyways.

Xtarsia:
This PR https://github.com/godotengine/godot/pull/85430 should make correctly determining when to use compatibility modifications much simpler.

TokisanGames (approved):
Great, thanks for all the updates. I've made a few changes and am squashing now.

TokisanGames:
Ok, down to 4 clean, topical commits. Thanks for all the work on this. It will help a lot of people.

Issue #26, opened 2023-01-28
Shader optimizations


TokisanGames:
### Performance

- [x] 1. Precalculate this in C++ whenever terrain_size is changed and pass in the uniform: `terrain_size_inv` rather than in the shader. WIll save millions of divides per second.
https://github.com/TokisanGames/Terrain3D/blob/0bb046eb2e0da0c137e91f74215f574f353f95fb/src/terrain_material.cpp#L98

----

- [x] 2. Replace a*b+c with fma(a,b,c), which is allegedly 10% faster. Not supported by webgl (but neither are texture arrays)
https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/shading_language.html

e.g.
* https://github.com/outobugi/GDExtensionTerrain/blob/0bb046eb2e0da0c137e91f74215f574f353f95fb/src/terrain_material.cpp#L73
* https://github.com/outobugi/GDExtensionTerrain/blob/0bb046eb2e0da0c137e91f74215f574f353f95fb/src/terrain_material.cpp#L80
* https://github.com/outobugi/GDExtensionTerrain/blob/0bb046eb2e0da0c137e91f74215f574f353f95fb/src/terrain_material.cpp#L141

* Maybe also make an FMA macro that uses fma or `*+` depending on compatibility or not

----

- [x] 3. ~~Lookup region only once per vertex. Move into vertex() and save as a varrying for the colormap in fragment.~~ - Looks bad on lower LODs when using this to calculate pixel data as the vertices are far apart.

-----
- [x] 4. Consider replacing `texturegrad` with `texture`, if we don't need it for 3D Projection. See #169 update 2 and https://github.com/outobugi/Terrain3D/issues/169#issuecomment-1658107044

-----
- [x] 5. As above remove all derivative functions and at least get them out of branches. Test optimizing branches

https://community.khronos.org/t/if-statements-in-shaders-confused-by-results/73518/15

-------
- [x] 6. Replace 16x16 Region map with an array. No need for a texture lookup

-----
- [x] 7. Replace 512x512 region blend map w/ a zoomed UV lookup of the region map, or a mathematical version from an array.

------
- [x] 8. ~~Since we're blending 4 adjacent vertices, move lookups to vertex() and weight in fragment.~~ - Doesn't look good on lower LODs as the vertices are farther apart, but see redesign below.

------
- [ ] 9. #245 

------
[...]

PR #506, opened 2024-10-02
Replacing Shader Code for FMA

Malidos:
I added fma into all shader snippets I could identify, fulfilling #26.2, but since fma only accepts variables of the same type I had to replace some float constants with vectors.

fma combines multiplication and addition into a single operator which should double the speed of that specific operation, but since the Shader contains many other more complex operations, this resulted in an **insignificant performance increase of about 1%** which makes me doubt whether this should be merged at all, but that's up to you.

It should be compatible with the changes in #500 because it introduces a replacement macro for fma.

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -194,14 +194,12 @@ vec3 get_normal(vec2 uv, out vec3 tangent, out vec3 binormal) {
 }
 
 vec3 unpack_normal(vec4 rgba) {
-	vec3 n = rgba.xzy * 2.0 - vec3(1.0);
-	n.z *= -1.0;
+	vec3 n = fma(rgba.xzy, vec3(2.0, 2.0, -2.0), vec3(-1.0, -1.0, 1.0));
 	return n;
 }
 
 vec4 pack_normal(vec3 n, float a) {
-	n.z *= -1.0;
-	return vec4((n.xzy + vec3(1.0)) * 0.5, a);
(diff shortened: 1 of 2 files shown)


TokisanGames:
Thank you. When making PRs you need to make a branch in your repository, then make a (Pull) Request to merge your branch into our main. Never your main to my main, and never add merge commits into your PRs. Never use git merge, nor "sync" on gh desktop. You can probably fix it by dropping the merge commit; I can probably as well. 

Read and follow the [Godot PR Workflow](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html) (and all our [contributor docs](https://terrain3d.readthedocs.io/en/stable/docs/contributing.html)).

At the least, I won't merge until 500 is merged, then we can rebase this and look at it then. I'll want to try it on my system and get more data points.

Why the changes from vec4s to mat4s?

@Xtarsia Thoughts?

Xtarsia:
any case where there are extra steps needed just to make fma() work, are usually not worth it. the same applies to nesting fma() instructions, which can prevent the compiler from being able to parallelize something it might have before.

the swap to mat4 and the following nested fma() are likley not gaining anything - and may even be slowing things down. But other areas like the rotation functions could well be faster.

it is handy to have this laid out with all currently possible places where fma() could be put.

TokisanGames:
@Xtarsia Ok, I'll go through and take the simpler ones, especially in utility functions, and skip the nested fmas and conversions to mat4s and vec3s before release.

PR #508, opened 2024-10-04
Link Terrain3D object when creating UndoRedo action to mark scene as unsaved

### LINKED ISSUE #184: Implement _get_unsaved_status()
Right now if there have been changes made to a scene then we close the scene, it won't ask to save changes. The only solution is available in 4.2 using _get_unsaved_status(). Implement when we move to 4.2

https://github.com/godotengine/godot-proposals/issues/2153

Xtarsia:
closes #184 

This moves the undo/redo history from global to the scene containing the object linked.

### DIFF
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -508,7 +508,7 @@ void Terrain3DEditor::_store_undo() {
 	EditorUndoRedoManager *undo_redo = _terrain->get_plugin()->get_undo_redo();
 	String action_name = String("Terrain3D ") + OPNAME[_operation] + String(" ") + TOOLNAME[_tool];
 	LOG(DEBUG, "Creating undo action: '", action_name, "'");
-	undo_redo->create_action(action_name);
+	undo_redo->create_action(action_name, UndoRedo::MERGE_DISABLE, _terrain);
 
 	LOG(DEBUG, "Storing undo snapshot: ", _undo_data);
 	undo_redo->add_undo_method(this, "apply_undo", _undo_data);


src/terrain_3d_editor.cpp
```diff
@@ -508,7 +508,7 @@ void Terrain3DEditor::_store_undo() {
 	EditorUndoRedoManager *undo_redo = _terrain->get_plugin()->get_undo_redo();
 	String action_name = String("Terrain3D ") + OPNAME[_operation] + String(" ") + TOOLNAME[_tool];
 	LOG(DEBUG, "Creating undo action: '", action_name, "'");
-	undo_redo->create_action(action_name);
+	undo_redo->create_action(action_name, UndoRedo::MERGE_DISABLE, get_terrain());
```
TokisanGames, line 511: You could also just provide `_terrain`.

But more importantly how does this tell godot to mark scenes as edited? I haven't tried it yet. The docs only say, "If custom_context object is provided, it will be used for deducing target history (instead of using the first operation).".

Is there a PR or some other documentation that shows how custom_context object connects to save status?

Xtarsia: Based off this comment on the PR mentioned before, just gave it a go and the results were desirable:

https://github.com/godotengine/godot-proposals/issues/2153#issuecomment-1685358047

TokisanGames:
Great, thank you.

src/terrain_3d_editor.cpp
TokisanGames: Thanks for the link. I filed a doc issue:
https://github.com/godotengine/godot-docs/issues/10043

Issue #441, opened 2024-07-30
Add option to toggle key bindings for Maya


TokisanGames:
### Description

Change ALT+LMB to something else like drag RMB.
Context https://github.com/TokisanGames/Terrain3D/pull/409#issuecomment-2257093148
Code https://github.com/tcoxon/Terrain3D/blob/d616eec395be9fb13c773a472ffd52526e8e717a/project/addons/terrain_3d/src/ui.gd#L403-L408

The setting can go in the menu. Save it to EditorSettings (see assetdock for how).

jesus-g20:
Hi, I am planning to work on this issue. Can I proceed or can I be assigned?

TokisanGames:
Yes, thanks for volunteering. Here are some options for implementation:
* Base
    * Add an Editor Setting "config/Alternate key bindings" that changes ALT+LMB to RMB when enabled. 
    * No gui. 
    * Change the [docs](https://github.com/TokisanGames/Terrain3D/blob/main/doc/docs/user_interface.md) to let people know the option is there.

* Optional Additional
    * In the Terrain3D Menu, make an entry called `Options...`. That dialog box will have the above setting on it (still an Editor Setting in the background), and more in the future. Change the docs to descibe this menu option instead of the editor settings.

TokisanGames:
@jesus-g20 If you can get this in by the next week or two it will go out in the next release.

TokisanGames:
MMB and RMB are occupied for most keymappings. I'm allowing the choice of Alt modifier key. See details here https://github.com/TokisanGames/Terrain3D/pull/509#issuecomment-2426776061.

PR #509, opened 2024-10-04
Added alternate key binding setting and updated documentation

jesus-g20:
This pull request adds a new editor setting (config/alternate_key_binding) to toggle between ALT+LMB and RMB for terrain editing operations. Documentation has also been updated to reflect this new option.

### DIFF
--- a/jesus
+++ b/jesus
@@ -1,0 +1,2 @@
+* [32mfeature-alternate-keybinding[m
+  main[m
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -37,6 +37,9 @@ func _init() -> void:
 func _enter_tree() -> void:
 	editor = Terrain3DEditor.new()
 	editor_settings = EditorInterface.get_editor_settings()
+	    # Initialize the new key binding option
+    if not editor_settings.has_setting("config/alternate_key_binding"):
+        editor_settings.set_setting("config/alternate_key_binding", false)
 	ui = UI.new()
 	ui.plugin = self
 	add_child(ui)
--- a/project/addons/terrain_3d/editor.gd
+++ b/project/addons/terrain_3d/editor.gd
@@ -259,8 +259,22 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> 
 		if p_event.get_button_index() == MOUSE_BUTTON_RIGHT and p_event.is_released():
 			ui.last_rmb_time = Time.get_ticks_msec()
 		ui.update_decal()
-			
-		if p_event.get_button_index() == MOUSE_BUTTON_LEFT:
+
+		    # Fetch the alternate key binding setting
+    	var use_rmb = editor_settings.get_setting("config/alternate_key_binding")
+
+			# If alternate keybinding is enabled, use RMB instead of ALT + LMB
+		    if use_rmb and p_event.get_button_index() == MOUSE_BUTTON_RIGHT:
+        if p_event.is_pressed():
+            if not editor.is_operating():
+                editor.start_operation(mouse_global_position)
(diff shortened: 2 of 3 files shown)


jesus
```diff
@@ -0,0 +1,2 @@
+* [32mfeature-alternate-keybinding[m
+  main[m
```
TokisanGames, line 2: This file doesn't need to be included. You can edit your commit, drop it, and force push your branch to the server.

project/addons/terrain_3d/editor.gd
```diff
@@ -37,6 +37,9 @@ func _init() -> void:
 func _enter_tree() -> void:
 	editor = Terrain3DEditor.new()
 	editor_settings = EditorInterface.get_editor_settings()
+	    # Initialize the new key binding option
+    if not editor_settings.has_setting("config/alternate_key_binding"):
+        editor_settings.set_setting("config/alternate_key_binding", false)
```
TokisanGames, line 42: Your indents are off. Whatever IDE you're using, the space/tab settings it has inserted are different from every other in this file. This is true in the block below as well.

We are using the regular Godot editor. This code will probably cause Godot to break on other systems as it complains about the indent mismatch. I'm surprised it works on your system.

project/addons/terrain_3d/editor.gd
```diff
@@ -256,8 +259,22 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) ->
...
+        if p_event.is_pressed():
+            if not editor.is_operating():
+                editor.start_operation(mouse_global_position)
+            editor.operate(mouse_global_position, p_viewport_camera.rotation.y)
+            return AFTER_GUI_INPUT_STOP
+        elif editor.is_operating() and p_event.is_released():
+            editor.stop_operation()
+            return AFTER_GUI_INPUT_STOP
+
+		elif not use_rmb and p_event.get_button_index() == MOUSE_BUTTON_LEFT:
```
TokisanGames, line 277: I'm not sure about this implementation. I don't want to have start_operation/operate/stop_operation in two places. 

The original task in #441 was to change ALT+LMB (lift floors/flatten peaks mode) to RMB. That code is in [ui.gd](https://github.com/TokisanGames/Terrain3D/blob/main/project/addons/terrain_3d/src/ui.gd#L448-L450). But that isn't even referenced to determine if ALT is pressed.

I need to change how ALT mode is implemented anyway for another task as my next step. So maybe hold off on reimplementing for now. You could clean up the tabs and remove the jesus file in the meantime.

jesus-g20: Hello, thank you for your feedback. Sorry to have such mistakes, is my first time collaborating to an open source. Please let me know if I can help in anything else when you implement the change of ALT mode. In the meantime, I will clean up the tabs and remove the file.

TokisanGames:
I've made my change, renaming lift_flatten to alt_mode in https://github.com/TokisanGames/Terrain3D/pull/514/commits/870ced27af011eb17d07cfdbb6c8a7f71662a714.

You can now rebase your PR `git pull --rebase upstream main` and continue fixing your implementation, then force push into this branch.

Don't duplicate the operation function calls. All you need to do is swap the functionality of ALT and RMB. If in alternate mode then when MOUSE_BUTTON_RIGHT is checked instead check ALT, and vice versa.

Some references are embedded in conditionals. It might be best to check alternate mode at the beginning of such a function, then check the appropriate button and set a boolean. Then all other conditionals in that function can just check the boolean.

Also it's not ideal to check something with a string in a function that's called continuously like mouse motion input. `editor_settings.get_setting("config/alternate_key_binding")`. Instead call this in `editor._edit()` and set a variable so during mouse motion only a boolean is checked.

We'll release when all of the items in the 0.9.3 column on the roadmap are completed.

TokisanGames:
Thanks for the start on this.

PR #516, opened 2024-10-08
Terrain Shader updates

### LINKED ISSUE #65: Implement 3D projection
https://github.com/cdxntchou/IndexMapTerrain

Xtarsia:
_Admin edit:_ 
Fixes #65 

- 3D Projected Textureing
- Ground Truth Fragment normals (required for projection)
- Screenspace derivative based branching of the bilinear blend. (FPS boost)
- reimplements the lookup grid with alternative method to match normal lookups.
- Removes get_height() and get_normal() functions in favour in doing them inline, access to texture fetches for other funcs
- region border fix removed, as made redundant
- vertex normals are present as they are almost free anyways. (despite not being used atm, likley needed for webGL)
- added toggle to macro variation
- added Texture LOD bias
- added depth blur effect (can use useful in conjunction with DOF shaders, or at small values preventing pixel shimmer at distance)

overall should give a decent FPS boost across the board.
everything works across compat/mobile/forward+

no projection:
image

projection (0.5 threshold):
image

texture LOD bias 0.8, depth blur 8.0:
image

a note on painted UV rotations and projected UVs - they work sort of, but at high surface slopes, the control map is still very skewed, as seen here:
image

It is possible to paint uv rotation, but it requires care. The dynamic mode is relative to the xz plane, not the projected plane so it is un-reliable. with some close attention, good results can be achieve if really specific things are desired on steep faces, tho issues are unavoidable at the crossover between different projections:
image

EDIT: fixing debug views to work with this, included a small upgrade to the control map texture view:
image


TokisanGames:
Looking great. Needs a rebase, which will apply base texture = 0 on paint. It can also be squashed. Leave new features or old bug fixes in separate commits, but squash your fixes for this PR. Thanks!

* Let's call Texture LOD Bias, Texture sharpness 

* Macrovation doesn't get projected, so gets stretched into these weird vertical striations

image
{8C5083C4-3601-4FDC-89A5-11076A49E400}

* Is there a benefit to limiting 3D projection to just one texture?
* I think the Projection threshold should default to 0.8 to align with the autoshader slope.
* Can/should projection work on the worldnoise?

* Unrelated to this PR, but I want to mention it so we can discuss. There's a subtle flicker in textures that I'd like to track down sometime. It's like a moire effect, so it might have to do with texture filtering. It's present all over whether projection or uv rotation are used or not, but it is most obvious if you make a tall column so the rock texture is stretched, disable projection, reduce lod bias to minimum, then move the camera towards and away from the column.

{33F8D973-25DD-4B6F-BD42-5F2D61028BA6}

(On the otherhand, just looking at this static image looks like it's flickering to my eyes, but I asked my gf to look and she said this is static, but moving the camera in and out in Godot definitely flickers).

src/shaders/main.glsl
```diff
@@ -314,117 +283,186 @@ void get_material(vec2 base_uv, uint control, ivec3 iuv_center, vec3 normal, out
...
 	// Macro variation. 2 Lookups
-	float noise1 = texture(noise_texture, rotate(uv*noise1_scale * .1, cos(noise1_angle), sin(noise1_angle)) + noise1_offset).r;
-	float noise2 = texture(noise_texture, uv*noise2_scale * .1).r;
-	vec3 macrov = mix(macro_variation1, vec3(1.), clamp(noise1 + v_vertex_xz_dist * .0002, 0., 1.));
-	macrov *= mix(macro_variation2, vec3(1.), clamp(noise2 + v_vertex_xz_dist * .0002, 0., 1.));
-
+	vec3 macrov = vec3(1.0);
+	if (enable_macro_variation) {
+		float noise1 = texture(noise_texture, rotate(uv*noise1_scale * .1, cos(noise1_angle), sin(noise1_angle)) + noise1_offset).r;
+		float noise2 = texture(noise_texture, uv*noise2_scale * .1).r;
```
TokisanGames, line 453: Please add spaces in uv*noise1/2

Xtarsia:
> Let's call Texture LOD Bias, Texture sharpness

Whilst that is the end result, I actually think it should be Texture_Mipmap_Bias which will match the same setting found in `rendering/textures/default_filters/texture_mipmap_bias` the value is a 1:1 modifier, eg setting the project value to -0.5 is directly countered by setting the terrain shader version 1.5.

> Please update the minimum.gdshader, stripping out all material functions.

Noted!

> Macrovation doesn't get projected, so gets stretched into these weird vertical striations

2 ways to go about this, either mix out the effect the surface angle increases, which has almost no cost, or make it part of the material, adding 3 extra noise samples. The former looks fine to me.

> Is there a benefit to limiting 3D projection to just one texture?

Since this piggy backs of the existing bilerp and the new normal method (which itself is faster), its just some uv math and no additional lookups, so I dont think so.

> I think the Projection threshold should default to 0.8 to align with the autoshader slope.

Easily done!

> Can/should projection work on the worldnoise?

It can.. but it would require "sampling" the world noise 8x instead of 3x, or just ignoring the derivative breaks, which are noticable.
I actually want to make world noise reads in fragment optional, using just vertex normals for them is quite a perf win without much visual loss at low octaves (4 or less).

> ..There's a subtle flicker in textures that I'd like to track down..

I have noticed this too, it seems related to mipmap selection / anisotropy and is mitigated with positive mipmap bias, at the cost of sharpness. It does need a bit more investigation.

TokisanGames:
> mix out the effect the surface angle increases, which has almost no cost,

This is fine. They can add variation on the colormap if needed.

Xtarsia:
World noise now has a toggle for fragment normals.

minimum shader updated.

Squashed.

TokisanGames:
* What is projection_angular_division? I think above 6 or so looks unusable. 1.0 looks better than other numbers.
 
* No macrovariation on rock faces looks rather bland. I'm going to add in a variable slope filter for macrov and default to 33%, which avoids striations on vertical, but if they never use cliffs, they can utilize up to 100%.

* Texture depth blur should start at zero, so I'm adjusting that.

Xtarsia:
> * What is projection_angular_division? I think above 6 or so looks unusable. 1.0 looks better than other numbers.

Its how many surface angles the projection can be divided into per axis, high values essentially give per-domain projection, which works better with small texture scales - where a 1k texture is scaled so that it covers only 1m x 1m for instance.


>     * No macrovariation on rock faces looks rather bland. I'm going to add in a variable slope filter for macrov and default to 33%, which avoids striations on vertical, but if they never use cliffs, they can utilize up to 100%.

Fair.


>     * Texture depth blur should start at zero, so I'm adjusting that.

would have to add 1.0 in the shader to the uniform, but it would make the slider more intuitive.

TokisanGames:
Great, thank you! I did a lot of testing with this and think it's good.

The demo rock texture is unattractive and needs to be replaced. Using it with 3D projection doesn't look good. But then I turn it off and it looks worse. It looks ok with the grass, and with the rock I'm using in my game.

I'm dropping "texture" from texture_mipmap_bias, texture_depth_blur, and adding bias_distance to replace the hard coded number as it affects both depth_blur and bias.

PR #518, opened 2024-10-09
base-equal-to-over-spray-condition-fix

Xtarsia:
getting base id and over id on the control map can occur when spraying, and then rolling the same id over the same area.
then, if base == over and useing the spray tool for another texture, it behaves like the roller tool instead.

this corrects that behaviour.


TokisanGames:
Thanks!

PR #521, opened 2024-10-10
Fix Instancer working w/ vertex spacing

### LINKED ISSUE #479: Vertex spacing >= 2 issues
### Update transforms 
Resize to 2+, paint foliage, resculpt and it won't update transforms.
It seems to work at 1.1, 1.2, 1.5, 1.8; it's a little offset at 1.95.

Note that around 2+, the transforms might be saved in another region, though they shouldn't. Add_transforms now filters through the region separator, but that doesn't take into account vertex spacing.

### Loading

~~Regions don't load properly at startup w/ transforms if vertex spacing >=2~~
Edit: Seems fine now

TokisanGames:
Superseded by #523 
Fixes #479 

To test at spacing <1, 1, >1
* painting, w/ slope and ALT slope
* erasing, w/slope and ALT slope
* Saving & reloading
* Removing regions where instances were created at all scales to ensure the instances are attached to the proper region
* Sculpting underneath instances

Key reminder: ALT to inverse slope. CTRL+SHIFT to remove all instances.

The only thing I've seen is if you remove a region when vertex_spacing != 0, the instances in an adjacent region transform slightly. They aren't removed nor moved, but they seem to grow, shrink, or shear slightly. It isn't noticable at vertex_spacing 1. It may be due to the way we're scaling the MMI laterally only.

Remaining issues from first commit:
* update_transforms < 1 doesn't work
* Vertex spacing transforms the MMI!! And working on fixing this broke everything else.


TokisanGames:
Superseded by https://github.com/TokisanGames/Terrain3D/pull/523

PR #523, opened 2024-10-11
Instancer storage rework

### LINKED ISSUE #479: Vertex spacing >= 2 issues

Xtarsia:
Fixes #479 

Implements region localised storage of instancer transforms.

See discussion
https://github.com/TokisanGames/Terrain3D/discussions/524

- [x] 1. Embed vertex_spacing in transforms, relative to local region.
Add vertex_spacing to Terrain3DRegion.
Only change vertex_spacing of transforms on load/set_vs().

- [x] 2. Split MM data down to 32x32 storage
append_location -> append_region -> append_cell -> storage in 32x32

- [x] 3. Change MM/MMI generation off of new data structure

- [x] 4. Add region container in tree / instancer

- [x] 5. Support changing region size

- [x] 6. Upgrade old data from 0.9.2 and 0.9.3

Bug squash and misc:
- [x] Hand removing instances doesn't fully delete regions from tree.
- [x] Update transforms isn't accurate at 0.5 vertex spacing.
- [x] Changing region size twice (1024-512-1024) broken.
- [x] Remove_instances operates much faster than add_instances now - should be same speed, and not "blocky".
- [x] Naming convention as per https://github.com/TokisanGames/Terrain3D/pull/523#discussion_r1802691896
- [x] Expose visibility distance for each mesh asset
- [x] Fix visibility ranges on MMIs
- [x] Dictionary issues, test std::unordered_map
- [x] Crash changing vertex spacing
- [x] Removal slope filter being smaller than the add slope filter is an annoyance
- [x] Swap_ids() and other functions haven't been touched yet (comment out region::get_multimeshes() to reveal all)
- [x] Remove_instances: Crash in w/ remove_all
- [x] Remove instances: remove not completely filtered by mesh_id
- [x] Remove instances: warning - not erasing _instances[mesh_id]
- [x] Crashes on close pretty frequently, after backup_region() runs. Potentially corrupts data
- [x] Transfer data buffers directly into Multimesh rather than per instance - unnecessary set_buffer has a memcpy
- [x] Remove_instances: Sparse removal of instances shows the cell grid

### DIFF
--- a/src/terrain_3d_instancer.cpp
+++ b/src/terrain_3d_instancer.cpp
@@ -132,7 +132,7 @@ void Terrain3DInstancer::_backup_region(const Ref<Terrain3DRegion> &p_region) {
 	}
 }
 
-Ref<MultiMesh> Terrain3DInstancer::_create_multimesh(const int p_mesh_id, const TypedArray<Transform3D> &p_xforms, const TypedArray<Color> &p_colors) const {
+Ref<MultiMesh> Terrain3DInstancer::_create_multimesh(const int p_mesh_id, const TypedArray<Transform3D> &p_xforms, const PackedColorArray &p_colors) const {
 	Ref<MultiMesh> mm;
 	IS_INIT(mm);
 	Ref<Terrain3DMeshAsset> mesh_asset = _terrain->get_assets()->get_mesh_asset(p_mesh_id);
@@ -254,7 +254,7 @@ void Terrain3DInstancer::add_instances(const Vector3 &p_global_position, const D
 	Terrain3DData *data = _terrain->get_data();
 
 	TypedArray<Transform3D> xforms;
-	TypedArray<Color> colors;
+	PackedColorArray colors;
 	for (int i = 0; i < count; i++) {
 		Transform3D t;
 
@@ -364,7 +364,7 @@ void Terrain3DInstancer::remove_instances(const Vector3 &p_global_position, cons
 
 		LOG(EXTREME, "Removing ", count, " instances from ", p_global_position);
 		TypedArray<Transform3D> xforms;
-		TypedArray<Color> colors;
+		PackedColorArray colors;
 		for (int i = 0; i < multimesh->get_instance_count(); i++) {
 			Transform3D t = multimesh->get_instance_transform(i);
 			// If quota not yet met, instance is within a cylinder radius, and can work on slope, remove it
@@ -392,7 +392,7 @@ void Terrain3DInstancer::remove_instances(const Vector3 &p_global_position, cons
 void Terrain3DInstancer::add_multimesh(const int p_mesh_id, const Ref<MultiMesh> &p_multimesh, const Transform3D &p_xform) {
 	LOG(INFO, "Extracting ", p_multimesh->get_instance_count(), " transforms from multimesh");
 	TypedArray<Transform3D> xforms;
-	TypedArray<Color> colors;
+	PackedColorArray colors;
 	for (int i = 0; i < p_multimesh->get_instance_count(); i++) {
 		xforms.push_back(p_xform * p_multimesh->get_instance_transform(i));
 		Color c = COLOR_WHITE;
@@ -404,7 +404,7 @@ void Terrain3DInstancer::add_multimesh(const int p_mesh_id, const Ref<MultiMesh>
 	add_transforms(p_mesh_id, xforms, colors);
 }
 
-void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Transform3D> &p_xforms, const TypedArray<Color> &p_colors) {
+void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Transform3D> &p_xforms, const PackedColorArray &p_colors) {
 	IS_DATA_INIT_MESG("Instancer isn't initialized.", VOID);
 	if (p_xforms.size() == 0) {
 		return;
@@ -418,7 +418,7 @@ void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Tr
 	Dictionary colors_dict;
 	Ref<Terrain3DMeshAsset> mesh_asset = _terrain->get_assets()->get_mesh_asset(p_mesh_id);
 
-	// Separate incoming transforms/colors into Dictionary { region_loc => Array[Transform3D] }
+	// Separate incoming transforms/colors by region Dict{ region_loc => Array() }
 	LOG(INFO, "Separating ", p_xforms.size(), " transforms and ", p_colors.size(), " colors into regions");
 	for (int i = 0; i < p_xforms.size(); i++) {
 		// Get adjusted xform/color
@@ -433,12 +433,13 @@ void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Tr
 		Vector2i region_loc = _terrain->get_data()->get_region_location(trns.origin);
 		if (!xforms_dict.has(region_loc)) {
 			xforms_dict[region_loc] = TypedArray<Transform3D>();
-			colors_dict[region_loc] = TypedArray<Color>();
+			colors_dict[region_loc] = PackedColorArray();
 		}
 		TypedArray<Transform3D> xforms = xforms_dict[region_loc];
-		TypedArray<Color> colors = colors_dict[region_loc];
+		PackedColorArray colors = colors_dict[region_loc];
 		xforms.push_back(trns);
 		colors.push_back(col);
+		colors_dict[region_loc] = colors; // Note similar bug as godot-cpp#1149 needs this for PCA
 	}
 
 	// Merge incoming transforms with existing transforms
@@ -446,7 +447,7 @@ void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Tr
 	for (int i = 0; i < region_locations.size(); i++) {
 		Vector2i region_loc = region_locations[i];
 		TypedArray<Transform3D> xforms = xforms_dict[region_loc];
-		TypedArray<Color> colors = colors_dict[region_loc];
+		PackedColorArray colors = colors_dict[region_loc];
 		LOG(DEBUG, "Adding ", xforms.size(), " transforms to region location: ", region_loc);
 		append_location(region_loc, p_mesh_id, xforms, colors);
 	}
@@ -454,7 +455,7 @@ void Terrain3DInstancer::add_transforms(const int p_mesh_id, const TypedArray<Tr
 
 // Appends new transforms to existing multimeshes
 void Terrain3DInstancer::append_location(const Vector2i &p_region_loc, const int p_mesh_id,
-		const TypedArray<Transform3D> &p_xforms, const TypedArray<Color> &p_colors, const bool p_clear, const bool p_update) {
+		const TypedArray<Transform3D> &p_xforms, const PackedColorArray &p_colors, const bool p_clear, const bool p_update) {
 	IS_DATA_INIT(VOID);
 	Ref<Terrain3DRegion> region = _terrain->get_data()->get_region(p_region_loc);
 	if (region.is_null()) {
@@ -465,12 +466,12 @@ void Terrain3DInstancer::append_location(const Vector2i &p_region_loc, const int
 }
 
 void Terrain3DInstancer::append_region(const Ref<Terrain3DRegion> &p_region, const int p_mesh_id,
-		const TypedArray<Transform3D> &p_xforms, const TypedArray<Color> &p_colors, const bool p_clear, const bool p_update) {
+		const TypedArray<Transform3D> &p_xforms, const PackedColorArray &p_colors, const bool p_clear, const bool p_update) {
 	Dictionary mesh_dict = p_region->get_multimeshes();
 
 	// Collect old data
 	TypedArray<Transform3D> old_xforms;
-	TypedArray<Color> old_colors;
+	PackedColorArray old_colors;
 	if (!p_clear) {
 		Ref<MultiMesh> multimesh = mesh_dict.get(p_mesh_id, Ref<MultiMesh>());
 		if (multimesh.is_valid()) {
@@ -543,7 +544,7 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
 				}
 				Ref<Terrain3DMeshAsset> mesh_asset = _terrain->get_assets()->get_mesh_asset(mesh_id);
 				TypedArray<Transform3D> xforms;
-				TypedArray<Color> colors;
+				PackedColorArray colors;
 				LOG(EXTREME, "Multimesh ", mesh_id, " has ", mm->get_instance_count(), " to review");
 				for (int i = 0; i < mm->get_instance_count(); i++) {
 					Transform3D t = mm->get_instance_transform(i);
@@ -584,7 +585,7 @@ void Terrain3DInstancer::copy_paste_dfr(const Terrain3DRegion *p_src_region, con
 	for (int i = 0; i < keys.size(); i++) {
 		int mesh_id = keys[i];
 		TypedArray<Transform3D> xforms;
-		TypedArray<Color> colors;
+		PackedColorArray colors;
 		Ref<MultiMesh> src_mm = src_mms[mesh_id];
 		if (src_mm.is_null()) {
 			LOG(ERROR, "Region has null multimesh for mesh_id ", mesh_id);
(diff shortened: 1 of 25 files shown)


TokisanGames:
Good start. I'll work on 4, placement in the tree by region.

Xtarsia:
Great!
It occurs to me that each cell will need its own internal location relative the region.

Doing updates would need to place each mmi cell by (region_size * location + 32 * cell_location) * vertex scaling.

The vertex scaling update will have to operate locally per cell rather than region too.

The cell dictionary should be keyed by cell location relative to the region i belive.

Probably need another pass to fix the access / updates etc to so its relative to cells rather than regions once cells are in.

TokisanGames:
I have this to return cell location.
```
#define CELL_SIZE 32
Vector2i Terrain3DInstancer::get_cell(const Vector3 &p_global_position) {
	int region_size = _terrain->get_region_size();
	real_t vertex_spacing = _terrain->get_vertex_spacing();
	Vector2i cell;
	cell.x = (UtilityFunctions::floori(p_global_position.x / vertex_spacing) % region_size) / CELL_SIZE;
	cell.y = (UtilityFunctions::floori(p_global_position.z / vertex_spacing) % region_size) / CELL_SIZE;
	return cell;
}
```

I'm still just poking around the edges. I'm going to build up _instances simultaneously w/ _multimeshes so it doesn't break the old until I know the new works.

Xtarsia:
Thinking about it more, I think the transforms in each cell should be relative to the region still.

TokisanGames:
The last commit is the first intrusive change to the data structure. It works a little bit, with a lot broken. I'm done for the day in case you want to review it or take a crack and it. I'll start looking again in 10-12 hours.

It works enough to paint foliage on the ground at the 0, 0 cell of any region. Outside of that first cell, the height starts to increase linearly. I'm accumulating height somewhere.

Undo sort of works, but doesn't fully remove everything, which means destroy and/or destroy_by_location aren't 100%.

That's about it. Not much else works. Lots of bugs.

If you click the Demo node, there is a dump_tree bool you can click to print the Terrain3D tree and see the MMIs.

Xtarsia:
Added a 3rd modified value to the tuple (making it a triple now i guess?) allows much more targeted updates, especially when there are a lot of cells & transforms present.

update vertex spacing and update transforms are now both working.

removes instances ive started on, but it has issues, only seems to work in region 0, and causes the entire region to clear / suddenly have blank arrays.

i didnt touch undo, some issues persist.

TokisanGames:
When I loaded up my previous data it crashed when there was no tuple, so I fixed it. Also it was very slow to add. I removed, which cleared them all, then it stopped giving me so many problems by starting with new data.

Issues:
* Foliage used to stay in it's region and not be allowed to create transforms outside it. Now you can add transforms outside your region. They may also be going outside their grid.

* I'm crashing a lot on closing the editor. MSVC doesn't point to any of our code. It's likely in the allocation/deallocation.

* Sometimes I'm not able to reach some areas

After this I clicked around, changed vertex scaling a few times, then was able to paint here.

{BF510881-DC95-451E-B3C6-91366FB09811}
{4F022C24-64E9-430A-802F-9C69F0BAE054}

Xtarsia:
Before my comits last night, cell mmis were being translated to their cell positions, which would have required offsetting the transforms per cell rather than per region. Keeping transforms per region makes things easier all around.

However loading the current iteration with old data would mean it makes assumptions about the location of the mmis when updating.

For now i suggest we test with fresh data comit until things are stable and everything is working. Then look at converting old data.

Xtarsia:
https://github.com/TokisanGames/Terrain3D/pull/523/commits/ed06faf1c8dc9e2e6ca3ccb7f130bc00f46e9b87 does fully clear when removing, but causes crashes if adding afterwards?

I think we need a better way to remove the mmi for a cell once its empty.

Xtarsia:
Things are a bit of a mess atm with the nested for loops. Vertex spacing is broken still for remove/update.

Rather than looping every cell, a similar lookup method to get_region_loc(), would make it possible to calculate the potential cell ids needed, and then check if they exist, bringing the worst case of 16k cells for 2k regions checked per mesh to only cells in the size of the area checked.

TokisanGames:
I spent all day banging my head on the crash on close and don't have much to show for it, except that it is an [engine bug](https://github.com/godotengine/godot/issues/98182).

It crashes in Resource::GDClass because Godot frees the GDExtension first, then asks a Resource what class it is. It has a non-null _extension pointer which triggers the conditions to trigger it.

I can't tell what class it is, other than a Resource. I've done extensive testing trying to tell, printing out instance IDs from all Objects and Resources we're creating, but haven't found anything yet.


* One note is that after painting instances and saving. Closing crahes. But loading it up again, closing again doesn't crash. It only does if more instances are added. This suggests the trigger is in the add_instances chain rather than the load_region/update_mmi chain. Or maybe something to do in the replacement of the MMI data in update_mmi.

* The problem sometimes occurs on main. I was able to reproduce it, then seemingly fix it, then I couldn't reproduce the original anymore. Since the problem object is a Resource, I attempted to unref the MultiMesh before deleting the MMI, which seemed to fix the crash on main, but later when testing undoing the fix, I couldn't get it to crash on main anymore. In the PR, I tried unreffing the MM before deleting and to the old one anytime a new one was created and assigned to the MM, but it doesn't help the PR. 

Other issues:

* Can't add to cells that have been reloaded from disk.

* Even if I close the scene and say don't save, Godot often saves foliage inside the region files. It doesn't seem to save the sculpting, unless it crashes, then it could be either way.

Xtarsia:
outside of upgrading old data, everything should be working now, and needs testing.

there is a strange issue when repeatedly changing region size without saving between, that causes instancer data to just vanish for some regions.

TokisanGames:
Lots of good stuff here. Things are coming along well. Most things are working including undo and storing data within the regions.

Bugs

* I still get crashes on close pretty frequently. It's interesting that it doesn't crash after loading saved data, only add_instances or update_transforms.
* hand removing instances doesn't fully delete regions from tree. `MMI/Region_00-00` still exists even if empty. I don't know about cleaning up the dicts. _delete_by_location has everything needed to clean up now so can be an example.
* Reducing region size did not reorganize the MMI children attached to the tree. I went to 512, which created a (0,-4) region. The regions appear to have the correct instances. However there is no `MMI/Region_00-04` attached to the tree
* update transforms isn't accurate at 0.5 vertex spacing. The first click often leaves foliage floating and doesn't update the cell until you keep operating. 
* Changing region size twice (1024-512-1024) prevented me from removing instances. I could add. Changing again made some regions worth of instances go away.
* This was an issue, then it wasn't, so I'm not sure now: Godot saves instances to disk even if we say don't save. If it crashes, it only saves partially. This is the biggest concern w/ crashing: corrupted data. So far I haven't found a definitive example of losing data after I've saved.

Usage
* remove_instances operates much faster than add_instances now. Before they operated at the same speed so you could thin out instances. If you wanted quick removal you could increase the speed. Now it is only fast.
* Removal slope filter being smaller than the add slope filter is an annoyance we should address. Just pressing alt is an easy workaround.
* It would be good to offset the MMIs to the center of their cell location so we can expose visible distance.

src/terrain_3d_instancer.cpp
```diff
@@ -573,6 +573,7 @@ void Terrain3DInstancer::remove_instances(const Vector3 &p_global_position, cons
 								MultiMeshInstance3D *mmi = cast_to<MultiMeshInstance3D>(mesh_mmis[cell]);
 								remove_from_tree(mmi);
 								memdelete_safely(mmi);
+								mesh_mmis.erase(cell);
```
TokisanGames: How could this fix a crash? 

mesh_mmis stores v2i (mesh_id, lod)
`Dictionary mesh_mmis = region_mmis[Vector2i(m, 0)];`

cells are v2i(0-31, 0-31)
`Vector2i cell = cell_queue[c]; -> c_locs.keys();`

This line should do nothing, not finding the cell key in the mesh dict, most of the time, but might occasionally delete a valid mesh/lod if the cell matches eg around (0,0,).

I'm working on consolidating mmi deletion from remove_instances and destroy_by_location into one function.

Xtarsia: Without that line, I belive that the mesh_mmis dict would be left with a not-really-null refrence to an mmi that is queue free'd?

when update mmis is called, it would try to access the mmi that was just freed, as it would still be keyed.

I may be wrong.

having some set/get/check/delete functions for the data is 100% needed.

TokisanGames: I see my confusion. We are using different naming conventions. 

In your code, you're doing this:
`Dictionary region_mmis = _mmi_nodes[region_loc]; `
Whereas I call it this
`Dictionary mesh_dict = _mmi_nodes[region_loc];`

_mmi_nodes contains a dictionary of mesh dictionaries keyed by region location. ie, location-> mesh dictionary. So I think mesh_dict is more accurate. Whereas region_mmis or region_dict is what we might call _mmi_nodes itself; a dictionary of region dictionaries keyed by location.

On the next level, you have this:
`Dictionary mesh_mmis = region_mmis[Vector2i(m, 0)];`
Whereas I've written:
`Dictionary cell_dict = mesh_dict[mesh_key];`

Here we have a dictionary of cells keyed by mesh_id,lod. So I think cell_dict is a more accurate name.

-----

Further, we are also using multiple nested dictionaries with similar but different structures to track both MMIs and triple[arrays].

`Region::_instances{mesh_id:int} -> cell{v2i} -> [ TypedArray<Transform3D>, PackedColorArray, modified:bool ]`
`_mmi_nodes{region_loc} -> mesh{v2i(mesh_id,lod)} -> cell{v2i} -> MultiMeshInstance3D`

I have written both of these in different areas:

`Dictionary mesh_dict = region->get_instances();`
`Dictionary cell_dict = mesh_dict[mesh_key];`

`Dictionary mesh_dict = _mmi_nodes[region_loc];`
`Dictionary cell_dict = mesh_dict[mesh_key];`

Clearly this is unclear when looking at code if mesh_dict or cell_dict are referring to the data set or the MMI set.

This is confusing to us and impossible for newcomers. So I propose we standardize on this naming convention for the dictionaries to differentiate the two in our variables:

```
// MM Resources stored in Terrain3DRegion::_instances as
// _instances{mesh_id:int} -> cell{v2i} -> [ TypedArray<Transform3D>, PackedColorArray, modified:bool ]

Dictionary mesh_inst_dict = region->get_instances();
Dictionary cell_inst_dict = mesh_inst_dict[mesh_id];
Array triple = cell_inst_dict[cell];
```
```
// MMI Objects attached to tree, freed in destructor, stored as
// _mmi_nodes{region_loc} -> mesh{v2i(mesh_id,lod)} -> cell{v2i} -> MultiMeshInstance3D

Dictionary mesh_mmi_dict = _mmi_nodes[region_loc];
Dictionary cell_mmi_dict = mesh_mmi_dict[mesh_key];
mmi = cell_mmi_dict[cell];
```

Xtarsia: Oh, yeah, hard agree on the naming convention.

TokisanGames:
* Crash on close seems nullified if instancer::_backup_region() isn't called. It will crash if there is some instance data, and it's backed up via instancer or editor. It could be that Godot is freeing Terrain named resources (eg Terrain3DRegion) from the undo data after unloading the GDExtension. I was able to get the crash on the main branch. I think the current setup has aggravated the situation. Maybe we can manually free the editor data on scene close or something.
* I found another crash. Add some instances, change vertex_spacing to 2, add more, reset to 1. If it doesn't crash then change spacing a bit more and it will crash in `_destroy_mmi_by_location()` calling `_destroy_mmi_by_cell()`. `access violation.
**cell_pair** was 0x10.` Likely update_mmis is inserting a null somewhere that is triggered by set_vertex_spacing. I'm not sure how to check if the cell_pair iterator is null.
* Fixed: ~~Labels and MMI visibility do not turn off with the Terrain3D node.~~

TokisanGames:
* Slope improvement is great.
* Vertex spacing is really solid now.

* Sparse removal doesn't work. It's not possible to thin out a large area with an even distribution. Instead, it produces hot spots in a grid shape. Moderately cover a large area, then try to evenly, lightly remove at 10%.
image

Xtarsia:
Solving that removal problem is going to need some work

Density per cell needs reducing by the ratio of (cell area inside the radius of the ring : full cell area)

TokisanGames:
The original went:
* Grab the transforms data
* For the quota, if there's a Transform in the ring, drop it

This goes:
* for each cell
  * Get the xforms (for that cell)
  * Remove the quota (from that cell)

What about:
* for the quota
  * while something not removed or processed a random order of each cell in the circle
      * remove an instance within the circle

TokisanGames:
There's a crash in remove_instances. If one selects CTRL+SHIFT to remove all and the region does not have all N mesh types in  the asset dock, it will crash on the bottom line with an invalid index.
```C++
int mesh_count = _terrain->get_assets()->get_mesh_count();
for (int r = 0; r < region_queue.size(); r++) {
	Ref<Terrain3DRegion> region = data->get_region(region_loc);
	_backup_region(region); // Note: region hasn't changed yet and might not. Move lower
	Dictionary mesh_inst_dict = region->get_instances();
	Array mesh_types = mesh_inst_dict.keys();
	if (mesh_types.size() == 0) {
		continue;
	}
	// For this mesh id, or all mesh ids
	for (int m = (modifier_shift ? 0 : mesh_id); m <= (modifier_shift ? mesh_count - 1 : mesh_id); m++) {
		int mesh_id = mesh_types[m]; // Crash
```

TokisanGames:
Your last fix did pretty well to fix the crash and being able to remove other non-selected meshes. Continuing to test things went well until I CTRL+SHIFT removed a selection of 5 mesh types. As I got to the end of the instances this warning popped up many times for all of the mesh ids. 
`WARNING: Terrain3DInstancer#5838:remove_instances: Region at: (0, 0) has instance dictionary for mesh id: 4 but has no cells.`

It did remove the instances properly. However running the dump_data function I see the mesh_inst_dicts are still present, though they have no cells. They need to be erased from the parent instances dict. Though I do see a check for it in the code, but apparently it's not happening (or being done to a copy).

I updated swap_ids and removed some other functions we don't need. We're getting close.

TokisanGames:
More testing on the crash on close:
* The object it's crashing on is a Terrain3DRegion. It's an engine bug, attempting to call is_class() on it after freeing the library.
* It has no issue with the first three loaded demo regions. Upon creating backups the questionable instance is either the first (original duplicate) or second (redo duplicate) created.
* I also figured out how to clear the undo history on exit but that doesn't solve the issue.
* I nullified region::duplicate() and even renamed it in case it was calling the base class, but no change.
* The crashing code is only supposed to run if --verbose is specified, in the section printing out leaked instances. If you run from the command line it shouldn't crash. However I run with the msvc debugger, and though I don't specify --verbose, it does trigger the specific code.
* I've narrowed down the circumstance that initiate the crash down to a single line:
```c++
void Terrain3DEditor::_store_undo() {
	IS_INIT_COND_MESG(_terrain->get_plugin() == nullptr, "_terrain isn't initialized, returning", VOID);
	if (_tool < 0 || _tool >= TOOL_MAX) {
		return;
	}
	LOG(DEBUG, "Finalize undo & redo snapshots");
	Dictionary redo_data;
	// Store current locations; Original backed up in start_operation()
	redo_data["region_locations"] = _terrain->get_data()->get_region_locations().duplicate();
	// Store original and current backups of edited regions
	_undo_data["edited_regions"] = _original_regions; // ** causes latent crash **
	redo_data["edited_regions"] = _edited_regions;
```

I can return right before that line or comment it out and it won't crash. If I return right after or leave it uncommented it will crash. 
* _original_regions and _edited_regions are both TypedArray<Terrain3DRegion>.
* _undo_data and redo_data are both dictionaries, but the first one is a class member.

Alternatively, clearing the dictionary rather than reassigning it also prevents the crash. However it breaks our undo/redo. Still this suggests maybe the issue is in GDExtension instead of Godot.
```
void Terrain3DEditor::start_operation(const Vector3 &p_global_position) {
	IS_DATA_INIT_MESG("Terrain isn't initialized", VOID);
	LOG(INFO, "Setting up undo snapshot");
	_undo_data.clear();
	//_undo_data = Dictionary(); // New pointer instead of clear
```

I'm bug reporting now, then will attempt to figure out a solution to make a new data pointer without triggering the bug.

TokisanGames:
Using the default values of 20m/33% I made the following shapes. Then I found I could not remove any at 20m/33%, 100%, nor 1000%. 
{9A1E2CBB-CB47-4B45-84CC-D513993CF611}

I bumped it up to 100m/100% and found I could start to remove some instances if I centered the cursor in the red area so the edge of the circle was brushing the instances. It's still difficult to remove those that remain. It takes a lot of scrubbing on the edge of the ring, then it gradually removes more and more.

{986B0E34-D836-4734-8A7E-7275B64285D7}

Xtarsia:
should be working better now.

the remove rate decreasing as there are fewer and fewer instances remaining insiide the ring is invevitable with a low % strength, personally i find that more desirable for the most part, as over-shooting remove means undo or repaint.

I think its tollerable given the option to just use a higher strength value.

could also tweak the threshold a bit if wanted, currently its 0.15, but maybe 0.2 or even 0.225 could be ok, tho likley too fast compared to add instances rate.

src/terrain_3d_instancer.cpp
```diff
@@ -513,94 +794,169 @@ void Terrain3DInstancer::append_region(const Ref<Terrain3DRegion> &p_region, con
...
+			for (int c = 0; c < cell_queue.size(); c++) {
+				Vector2i cell = cell_queue[c];
+				Array triple = cell_inst_dict[cell];
+				TypedArray<Transform3D> xforms = triple[0];
+				PackedColorArray colors = triple[1];
+				TypedArray<Transform3D> updated_xforms;
+				PackedColorArray updated_colors;
+				for (int i = 0; i < xforms.size(); i++) {
+					Transform3D t = xforms[i];
+					real_t height = _terrain->get_data()->get_height(t.origin + global_local_offset);
```
Xtarsia, line 877: i belive this should actually include the same height offset subtraction that remove instances has for its slope check.

PR #537, opened 2024-11-01
Improve spray condition check

Xtarsia:
_Admin edit: Cherry-pick 0.9.3_

fixes use case of spraying onto freshly painted base texture, by overrwiting overlay texture when it is not visible (less than 0.5 blend)

still some undesireable edge cases can occur when intentionally "making a mess" with lots of different textures sprayed in close proximity but its not really avoidable, since these conditions are attempting to emulate the feel of using a spatmap with many channels, with the just the 2 in the index map.

Also added ~~alt~~ CTRL key modifier to invert the blend direction, which is handy for quickly un-spraying any over-shoot, also prevents any new ID changes on the control map during this so that it explicitly only effects the blend value.

modified the tool colors slightly, SEA_GREEN and FOREST_GREEN were not visually distinct to me

### DIFF
--- a/project/addons/terrain_3d/src/ui.gd
+++ b/project/addons/terrain_3d/src/ui.gd
@@ -15,8 +15,8 @@ const COLOR_LIFT := Color.ORANGE
 const COLOR_FLATTEN := Color.BLUE_VIOLET
 const COLOR_HEIGHT := Color(0., 0.32, .4)
 const COLOR_SLOPE := Color.YELLOW
-const COLOR_PAINT := Color.FOREST_GREEN
-const COLOR_SPRAY := Color.SEA_GREEN
+const COLOR_PAINT := Color.DARK_GREEN
+const COLOR_SPRAY := Color.PALE_GREEN
 const COLOR_ROUGHNESS := Color.ROYAL_BLUE
 const COLOR_AUTOSHADER := Color.DODGER_BLUE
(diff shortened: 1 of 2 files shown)


TokisanGames:
Alt is already used for the slope filter. However Ctrl is free for texture, and the video you showed on discord looks more like removal/inverse so I think that will fit better.

You mentioned we're getting into edge cases, but the way I see it is this is the primary case for using the spray tool. That is, paint two adjacent areas then spray the edge to blend. The tools should do that case perfectly, no matter the underlying base/overlay values. With only two textures it should be perfect. I think this changed after #518. 

I'm less concerned with mixing 3-4+ textures in one area, but it has worked well in the past and I was happy with it. 

I'll try it out a bit later today. I want to get this in soon and pushed into 0.9.3 as it seems broken now. I'll test this PR as well as pre-#518 with both 2-4 textures. 

To update you on my schedule, I want to get all the PRs merged in, but I'm working on getting my team what we need to get our demo pushed out and I need to work on a new video showing all the new features. You're free to work on whatever you're interested in and I'll get through the PRs eventually. Thanks!

Xtarsia:
I agree that the primary use case should take priority, and I belive this does cover that now.

Looking forwards to giving the demo a go :)

TokisanGames:
Excellent work as always!

PR #538, opened 2024-11-02
move cursor decals to shader

### LINKED ISSUE #533: Visibility ring around cursor when small
### Description

When the brush is small (<10m) and the camera is far, it can be hard to see where the cursor is. Even though the OS mouse is right there, having feedback from Terrain3D is necessary to feel like it's working. One of my artists asked for something like a second ring around the cursor for this visual indication. Could also be a ring and a cross hair.

It could just be a larger ring, but what might be better is something that stays the same size on screen like the region labels and region grid. In the shader this can be achieved by sizing it by `sqrt(camera distance)` like the region grid.

### LINKED ISSUE #534: Move cursor decals to shader
### Description

Let's continue moving all cursor usage to the shader for all renderers and drop decals.

Xtarsia:
removes all decals, closes #534

Adds a crosshair, with a threshold setting in the advanced menu, closes #533

Consolidate uniforms to just brush and ring textures

Also fixed derivative rotations
Also added more agressive branching to texture lookups in get_material()

Todo:

- [x] use crosshair instead of cross_hair
- [x] rework timer to lerp editor_decal_color uniform alpha to 0.
- [x] crosshair opacity to match editor_decal_color alpha
- [x] apply color space correction when updating editor_decal_color uniform RGB values if in compatibility
- [x] ensure correct position for bush operations


TokisanGames:
Great! 

* What do you think about taking care of #533 in this PR as well and adding a larger open crosshair or circle around the cursor when small (or always)?

* Please remove UI.gd:_on_setting_changed():update_decal(). It's annoying to have it popup all the time when changing texture colors. I have it on mine but havn't pushed it yet.

* Why do we need separate decals for the gradients? Are there any ideas or reasons for having a third decal? What do you think if instead of three open slots for textures, we specify them: 0= current brush, 1=ring, 2=crosshair/etc? Then in the shader we can name them such as use them by name rather than editor_decal[0]?

src/shaders/editor_functions.glsl
```diff
@@ -67,8 +68,10 @@ vec3 get_decal(vec3 albedo, vec2 uv) {
 				decal = texture(_editor_decal_2, decal_uv + 0.5).r;
 				break;
 		}
-		// Blend in decal; reduce opacity 55% to account for differences in Opengl/Vulkan and/or decals
-		albedo =  mix(albedo, _editor_decal_color[i].rgb, clamp(decal * _editor_decal_color[i].a * .55, 0., .55));
+		// Decal fadeout from last time stamp
+		float fade = clamp((_editor_decal_timestamp - TIME) * 8. - 6., 0., 1.);
```
TokisanGames, line 72: What do 8 and 6 represent?

Xtarsia: As only an offset time stamp is passed each update, those 2 values work with the offset to replicate the timer fade delay & speed of the original decal_timer.

Xtarsia:
ok im fairly happy with the cross hair now.

size 0;
afar:
image
upclose:
image

size 50;
afar:
image
upclose:
image

threshold setting in advanced, defaults to 16
image

Xtarsia:
shorter line length feels better
image

project/addons/terrain_3d/src/ui.gd
```diff
@@ -258,7 +243,6 @@ func _on_setting_changed() -> void:
 		return
 	brush_data = tool_settings.get_settings()
 	brush_data["asset_id"] = plugin.asset_dock.get_current_list().get_selected_id()
-	update_decal()
```
TokisanGames, line 261: Fade on timeout isn't working, but what I believe this did is showed the decal when the brush_size gets larger or smaller. Currently the decal size doesn't change on adjustment.

project/addons/terrain_3d/src/ui.gd
```diff
@@ -309,182 +297,122 @@ func update_decal() -> void:
...
+	editor_decal_size[0] = maxf(brush_data["size"], .5)
 	if brush_data["align_to_view"]:
 		var cam: Camera3D = plugin.terrain.get_camera();
 		if (cam):
-			decal.rotation.y = cam.rotation.y
+			editor_decal_rotation[0] = cam.rotation.y
 	else:
-		decal.rotation.y = 0
-
+		editor_decal_rotation[0] = 0
```
TokisanGames, line 314: Please change to 0.

TokisanGames (changes requested):
Thanks for working on this. In my review copy I rebased it, then tested.

* crosshair(s) seems to be most widely used as a single word. So all cross_hair references could drop the underscore.
 
* I'd like to see the crosshair lines the same opacity as the brush so they match better. 

* I don't see the fade on timeout, though I see there's code in there. The only way to make it disappear is a single right click, but it reappears when the mouse moves. There are times when the camera is close and the cursor is big, and we want to adjust something in the inspector and see the results on the terrain w/o the decal. Currently the process is to wait until it disappears then adjust. We need either the timeout, or maybe a better alternative is to hide the cursor as soon as the mouse leaves the viewport, or both.

* In this PR, with Paint mode, if you right-click-hold and move the camera, let go (cursor invisible), click-hold and paint, the cursor will appear at a wrong spot and paint, then jump to where it should be and continue painting.

* Comptability mode, Raise 100% strength, white decal Brightness is way too bright. As well as slope, spray, foliage ring. The other colors are acceptable. Forward foliage ring is a bit bright as well.
 
* I'm wondering if the lines should change thickness/size based on the brush size. 🤔 

* Can be rebased and squashed to one commit.

src/shaders/editor_functions.glsl
Xtarsia: Reworked to use a timer and tween, similar to original implementation, this method was not reliable.

Xtarsia:
> * In this PR, with Paint mode, if you right-click-hold and move the camera, let go (cursor invisible), click-hold and paint, the cursor will appear at a wrong spot and paint, then jump to where it should be and continue painting.

This happens on main too. Its related to get_intersection() GPU mode problems.

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -157,36 +157,39 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) ->
 
 	_read_input(p_event)
 	
+	## Setup active camera & viewport
+	# Always update this for all inputs, as the mouse position can move without
+	# necessarily being a InputEventMouseMotion object. get_intersection() also
```
TokisanGames, line 162: > as the mouse position can move without necessarily being a InputEventMouseMotion object

👀🤯

TokisanGames:
Given the new changes to the shader, these tasks could be addressed, if you're up for it:
* Audit the shader comments where they say `3 lookups` and make sure those are accurate for each section
* Review [shader design](https://terrain3d.readthedocs.io/en/latest/docs/shader_design.html) and offer updates to things that have changed, or new information that would be useful to devs
* Ensure the tip on [adding a custom map](https://terrain3d.readthedocs.io/en/latest/docs/tips.html#add-a-custom-texture-map) is still accurate.

Xtarsia:
what I wrote in the docs might need some more work.

the last commit is a bit cheeky, but updates the demo textures, and applies some detiling to them. Also made the grass a bit smaller too.

doc/docs/shader_design.md
```diff
@@ -74,6 +82,8 @@ All splat maps are sampled, and of the 16-32 values, the 4 strongest are blended
 
 *Side note:* Storing blend values in 3-bits is possible, where each of the 8 numbers represents an index in a array of 0-1 values: `{ 0.0f, .125f, .25f, .334f, .5f, .667f, .8f, 1.0f }`. In the future, this may be baked at runtime. However, editing using a 3-bit array of fixed values was exceedingly difficult and unsuccessful.
 
+*Side note 2:* The bilinear blend can be skipped for distant terrain, allowing only 1/4 of the samples normally required.
+
```
TokisanGames, line 86: Like the first side node, this is a potential future improvement that isn't implemented? I can section these two side notes off under a clearer heading, after the line below.

BTW, I'm waiting for you to tell me this PR is ready for another review.

Xtarsia: Hmm this probably could be changed, added it more as another reason to use index map over splat map - it is already implemented.

Xtarsia:
@TokisanGames Rebased, and hopefully finished.

TokisanGames:
> fixed derivative rotations

Looks great. Much more stable now.

> added more agressive branching to texture lookups in get_material()

Fantastic

Addressed issues:
* Show decals on modifier presses
* Don't show decal on inspector settings change (eg uv scale), only on tool settings

TokisanGames:
There are times when I get these errors starting up, closing a scene, or saving the code.

```
SCRIPT ERROR: Invalid assignment of index '0' (on base: 'Array[bool]') with value of type 'bool'.
          at: update_decal (res://addons/terrain_3d/src/ui.gd:321)
SCRIPT ERROR: Out of bounds get index '0' (on base: 'Array[Color]')
          at: @editor_decal_fade_setter (res://addons/terrain_3d/src/ui.gd:63)
```

TokisanGames:
The size of these arrays is indeed 0. I setup them in _init() and it worked on a first restart. But then I closed one scene and opened another scene and they're reporting size 0 again.

Xtarsia:
```gdscript
func set_visible(p_visible: bool, p_menu_only: bool = false) -> void:
	editor_decal_color.resize(3)
	editor_decal_position.resize(3)
	editor_decal_rotation.resize(3)
	editor_decal_size.resize(3)
	editor_decal_visible.resize(3)
	terrain_menu.set_visible(p_visible)

```

```gdscript
var editor_decal_fade: float :
	set(value):
		if (editor_decal_color.size() > 0):
			editor_decal_fade = value
			editor_decal_color[0] = Color(
				editor_decal_color[0].r,
				editor_decal_color[0].g,
				editor_decal_color[0].b,
				value
			)
			RenderingServer.material_set_param(mat_rid, "_editor_decal_color", editor_decal_color)
```

this seems to cover enough bases.

TokisanGames:
I have no issues with these changes. Do you get any errors with just these, and nothing in update_decal?

```python
 var editor_decal_fade: float :
        set(value):
                editor_decal_fade = value
+               if not plugin.terrain:
+                       return
                editor_decal_color[0] = Color(

func set_visible(p_visible: bool, p_menu_only: bool = false) -> void:
                ...
                visible = p_visible
                toolbar.set_visible(p_visible)
                tool_settings.set_visible(p_visible)
+               if editor_decal_color.size() != 3:
+                       # These sizes get reset to 0 on user scene closes. No idea why
+                       editor_decal_position.resize(3)
+                       editor_decal_rotation.resize(3)
+                       editor_decal_size.resize(3)
+                       editor_decal_color.resize(3)
+                       editor_decal_visible.resize(3)

```

Xtarsia:
Yes, if i have both the demo, and navdemo scenes open at the same time with terrain3d selected in both scenes, switch back and forth, and then close 1, I get errors again until i deselect and reselect the terrain3d node in the remaining scene.

TokisanGames:
Excellent work as always. Thank you.

PR #540, opened 2024-11-02
Emit edited AABB on undo/redo

### LINKED ISSUE #539: Terrain3DObjects: Undo after sculpt operation does not move object (until next sculpt operation)
### Terrain3D version

v0.9.3

### System information

macOS 14.5, Godot 4.3

Yes

### Issue description

When a sculpting operation is undone (ctrl+z), the objects under a Terrain3DObject do not move back to the position they were in before the sculpting operation. Only with the next sculpting operation the object is moved back to the original offset.

### Logs

_No response_

TokisanGames:
Fixes #539 
Cherry pick 0.9.3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,9 @@
 # Terrain3D
-project/test/
-project/addons/godot-jolt/
 project/addons/terrain_3d/bin/
 project/_dev/
 project/_export/
 project/_tests/
+project/test/
 src/_archive/
 _misc/
 _patches/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,11 +1,10 @@
 # Terrain3D
 project/test/
-project/addons/godot-jolt/
-project/addons/terrain_3d/bin/
```
TokisanGames, line 4: add this back in

Xtarsia:
Tested, resolves the bug.

PR #541, opened 2024-11-03
Fix channel packer scaling with display_scale

### LINKED ISSUE #531: Channel_packer doesn't handle display_scale properly
### Terrain3D version

master

### System information

v4.4

Yes

### Issue description

in display_scale = 100%, normal, other scale have bug

`V05_8ZDI`PN8BHVT10{~18
%UZZJA CE$RZ9Z9(14 G A8
_ A207G8{1TX7 LEAYDU $U

### Logs

none

benjiwolff:
Fixes #531.

The issue was that the `custom_minimum_size` of all panels and the window size does not scale with `display_scale`. Using panel containers to wrap sections instead of panels and using `Wrap Controls` on the window, solves the issue.

Display scale 200% on a 14" MacBook
image


TokisanGames:
Very good, thank you. I changed the window to be resizable and made the panelcontainer full rect. Now the window can resize as well.

PR #560, opened 2024-11-28
Small fix to terrain_3d_objects.gd

artoonu:
Add signal connection check before attempting to disconnect. Attempting to disconnect a non-existent connection caused an error print to the console.


artoonu:
Not sure if I clicked the correct things in GitHub, but I think it's changed

TokisanGames:
Thanks.

PR #564, opened 2024-12-05
force decompress textures when building arrays in compatibility

Xtarsia:
Fixes crashes, and removes the requirement to manually re-import textures for compatibility mode.

When the arrays are built, if in compatibility and the texture format is compressed, decompress before building the array that is sent to generated_texture.create()

Deprecated in 4.4 (hopefully)

### DIFF
--- a/src/terrain_3d_assets.cpp
+++ b/src/terrain_3d_assets.cpp
@@ -155,6 +155,7 @@ void Terrain3DAssets::_set_asset(const AssetType p_type, const int p_id, const R
 }
 
 void Terrain3DAssets::_update_texture_files() {
+	IS_INIT(VOID);
 	LOG(DEBUG, "Received texture_changed signal");
 	_generated_albedo_textures.clear();
 	_generated_normal_textures.clear();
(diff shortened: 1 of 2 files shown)


TokisanGames:
Great, thank you!

src/terrain_3d_assets.cpp
```diff
@@ -192,6 +195,11 @@ void Terrain3DAssets::_update_texture_files() {
 				return;
 			}
 			Ref<Image> img = albedo_tex->get_image();
+			//DEPRECATED 1.0 - remove in Godot 4.4
+			if (_terrain->is_compatibility_mode() && img->is_compressed()) {
+				warn_compatibility_decompress = true;
+				img->decompress();
```
TokisanGames, line 201: This probably doesn't work on export. The de/compression libraries are stripped out of the export templates.

TokisanGames: Also I don't know about mobile/compatibility as [The formats ETC1 and ETC2 are not supported for decompress](https://docs.godotengine.org/en/stable/classes/class_image.html#class-image-method-decompress). Export to mobile won't work because of the above, but if they are using the editor on mobile, they should be using PNG textures. It would make sense that they would default to using ETC2 textures rather than DXT5, which is unsupported on most mobiles.

Xtarsia: The primary intention of this was to avoid "crash to desktop"

I did quickly test an exported project, and vram format compression/decompression worked for me. It may just be all the png/jpg etc that are excluded, tho I didn't look any more into that specifically.

Perhaps the warning message could be more explicit? Even changed to an error?

PR #578, opened 2024-12-14
Fix gamma causing nans

Xtarsia:
* Resolves this error when using gamma != 1.0 for brushes: 

image

* Adds a check to correct nan / inf if present in an existing heightmap.


TokisanGames:
Great sleuth work. Thanks!

PR #591, opened 2025-01-07
Add config warnings for texture size/format/mipmap mismatch

### LINKED ISSUE #542: Add Configuration Warnings
Admin edit:
Add configuration warnings for:
* No mipmaps
* Mismatched texture size
* Mismatched texture format
* Compatibility mode && albedo/normal is compressed (and log error)

-----

### Terrain3D version

v0.9.3

### System information

Godot 4.3 - Steam Deck

Not applicable

### Issue description

I was testing on steam deck and ran into a performance issue that basically boiled down to the fact that the texture didn't have mipmaps enabled. It was really bad, from like 60 fps down to 20 fps.

There might be an export type to tell the editor the texture is used for 3d, but if not, would be nice to have a configuration warning somewhere, because adding textures to the terrain doesn't automatically "prepare" them for 3D use like Godot does when you add them to a material.

### Logs

_No response_

TokisanGames:
Fixes #542 

Add SceneTree (config) warnings for:
* Texture size mismatch
* Texture format mismatch
* Texture mipmap mismatch

Add console warnings for textures:
* No mipmaps
* Not square
* Not power of 2

@Xtarsia Would you review this please? Are there any config warnings we should add? 

Compatibility mode && albedo/normal is compressed is probably fine as a warning since it works and can be ignored. Config warnings is a nag screen, so should be considered "must dos".

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -32,6 +32,12 @@ using namespace godot;
 #define V3_ZERO Vector3(0.f, 0.f, 0.f)
 #define V3_MAX Vector3(FLT_MAX, FLT_MAX, FLT_MAX)
 
+// Terrain3D::_warnings is uint8_t
+#define WARN_MISMATCHED_SIZE 0x01
+#define WARN_MISMATCHED_FORMAT 0x02
+#define WARN_MISMATCHED_MIPMAPS 0x04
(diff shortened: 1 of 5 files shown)


TokisanGames:
Great idea, thanks.

PR #594, opened 2025-01-15
Correctly identify fragment closing brace to allow custom light functions

Xtarsia:
Shader inserts (Editor functions & debug views) currently find the last " } " in the shader, however this prevents use of a custom light() function.

Identifying the correct " } " fixes this issue.

### DIFF
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -194,7 +194,25 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 	}
 
 	// Insert at the end of the shader, before the end of `fragment(){ }`
-	idx = shader.rfind("}");
+	// Check for each nested {} pair until the closing } is found.
+	regex->compile("void\\s*fragment\\s*\\(\\s*\\)\\s*{");
+	match = regex->search(shader);
+	idx = -1;
+	if (match.is_valid()) {
+		int start_idx = match->get_end() - 1;
+		int pair = 0;
+		for (int i = start_idx; i < shader.length(); i++) {
+			if (shader[i] == '{') {
+				pair++;
+			} else if (shader[i] == '}') {
+				pair--;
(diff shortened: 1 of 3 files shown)


TokisanGames:
This will break if there is no space after a statement's semicolon and the last `}`:


```
void fragment() {vec2 uv = UV;}
```
or
```
void fragment() {vec2 uv = UV

;}
```
or
```
void fragment() {vec2 uv = UV;

int a;}
```

```
  333 | void fragment() {vec2 uv = UV;
  334 |
  335 | int a
E 336->  ALBEDO = get_decal(ALBEDO, uv);
  337 |
  338 | ;}
SHADER ERROR: Expected a ',' or ';'.
          at: (null) (:335)
```

src/terrain_3d_material.cpp
```diff
@@ -194,7 +194,25 @@ String Terrain3DMaterial::_inject_editor_code(const String &p_shader) const {
 	}
 
 	// Insert at the end of the shader, before the end of `fragment(){ }`
-	idx = shader.rfind("}");
+	// Check for each nested {} pair until the closing } is found.
+	regex->compile("void\\s+fragment\\s*\\(");
+	match = regex->search(shader);
+	idx = -1;
+	if (match.is_valid()) {
+		int start_idx = shader.find("{", match->get_start());
```
TokisanGames, line 202: I think get_start returns the `v`. This is fine, but you could optionally expand the match to include `) {`, and use get_end() to start on the `{`.

Xtarsia: Good point!

Xtarsia:
Fixed the idx value adjustment, and it seems to work just fine when that space is missing now as well!

tested with this monstrosity:
image

also no space at the end, along with every debug view enabled at the same time.

TokisanGames:
Great work. Thanks!

Issue #100, opened 2023-07-02
Draw region grid in shader instead of gizmos


TokisanGames:
Todo
- [x] Different color for defined regions, and no regions (white, grey or maybe white at .50% opacity)
- [x] Different color when the mouse is over it for add (green) and remove (red) when using the region tool (mouse tracking added in #500)
- [x] Blue line for maximum bounds (though we may drop bounds in #491)
- [x] Reduce the region gizmos to draw the add region square only when background mode = NONE and the mouse is in an empty space with no terrain mesh

-------

Using the editor gizmos for regions is a bit ugly and doesn't occlude. This would look so much better if drawn directly on the terrain, which will traverse the heights and occlude when close. We could even inject that code into custom shaders when inside the editor.

![image](https://github.com/outobugi/GDExtensionTerrain/assets/632766/908eb99d-3108-423f-93c2-2713cf206d55)

A preview of shader based region grid is in [the grid debug view shader](https://github.com/outobugi/Terrain3D/commit/f1ecab4159218dab700f754412ad5232ca3fd879#diff-f40b75ff1c053ed3b8dbe514c95900394a8970e0ba207b04afb37a7f05ff6121). It nicely occludes and stays the same size regardless of distance. Just needs mouse position and highlight color and the region gizmo can be dispensed with.

![image](https://github.com/outobugi/Terrain3D/assets/632766/58597c29-bcd4-49f7-af2f-b308f141a9ad)

However, we can't draw on the terrain where no terrain exists. So I'm not sure what to do about the grid for empty regions. Perhaps we do a hybrid and use gizmos for empty slots.

TokisanGames:
A preview of shader based region grid is in [the grid debug view shader](https://github.com/outobugi/Terrain3D/commit/f1ecab4159218dab700f754412ad5232ca3fd879#diff-f40b75ff1c053ed3b8dbe514c95900394a8970e0ba207b04afb37a7f05ff6121). It nicely occludes and stays the same size regardless of distance. Just needs mouse position and highlight color and the region gizmo can be dispensed with.

![image](https://github.com/outobugi/Terrain3D/assets/632766/58597c29-bcd4-49f7-af2f-b308f141a9ad)

TokisanGames:
The region_grid is still a debug shader, but the setting is now exposed in the main inspector under Regions. I've updated the Op with remaining steps.

Xtarsia:
[...]

PR #599, opened 2025-01-19
Fix triangle pairs to generate edges for ArrayMesh, Navigation

### LINKED ISSUE #597: Baked navmesh doesn't equally fill the painted navigable area
### Terrain3D version

0.9.3a

### System information

Godot v4.3.stable.mono - Windows 10.0.26100 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4070 Ti SUPER (NVIDIA; 32.0.15.6094) - AMD Ryzen 9 9900X 12-Core Processor (24 Threads)

Yes

### Issue description

When baking a navmesh on a Terrain3D node, the navmesh is generated much closer to certain edges of the painted navigable area than others. For instance, on the following navmesh generated from baking a single, completely flat, 64 x 64 terrain region with default settings, the top and left edges are much closer to the edge of the region than the bottom and right edges:

Image

I'd expect a result closer to the following image with all of the edges of the navmesh being roughly equal distance from the edge of the terrain region:

Image

This issue is also present inside the navigation demo:

Image

It is also present within terrain regions and not just at the border between regions.

Image

### Logs

_No response_

Xtarsia:
Admin edit
* Fixes #597 extending triangle pair generation into the region edge, but not into holes. 
* Fixes navigation region edges and arraymesh edges
* Reduces lookups by 2/3rds, increasing speed by 33%.
* Related to #451

----

fixes #597 by adding a 0.5 offset to the uv when doing the control_map lookup.

Edit: And generating triangle for nav mesh if any vertex point has nav enabled.

image

### DIFF
--- a/project/demo/data/assets.tres
+++ b/project/demo/data/assets.tres
@@ -19,6 +19,7 @@ distance_fade_max_distance = 96.0
 [sub_resource type="Terrain3DMeshAsset" id="Terrain3DMeshAsset_2qf8x"]
 name = "TextureCard"
 generated_type = 1
+height_offset = 0.5
 material_override = SubResource("StandardMaterial3D_b2vqk")
 last_lod = 0
 last_shadow_lod = 0
@@ -29,6 +30,7 @@ visibility_range = 128.0
 name = "LODExample"
 id = 1
 scene_file = ExtResource("1_4jrdu")
+height_offset = 0.5
 last_lod = 3
 last_shadow_lod = 3
 
--- a/project/demo/src/Enemy.gd
+++ b/project/demo/src/Enemy.gd
@@ -47,7 +47,9 @@ func _physics_process(p_delta: float) -> void:
 
 	# Ensure enemy doesn't fall through terrain when collision absent
 	if get_parent().terrain:
-		global_position.y = maxf(global_position.y, get_parent().terrain.data.get_height(global_position))
+		var height: float = get_parent().terrain.data.get_height(global_position)
+		if not is_nan(height):
+			global_position.y = maxf(global_position.y, height)
 
 
 func _on_velocity_computed(p_safe_velocity: Vector3) -> void:
--- a/src/shaders/editor_functions.glsl
+++ b/src/shaders/editor_functions.glsl
@@ -5,7 +5,7 @@
 R"(
 //INSERT: EDITOR_NAVIGATION
 	// Show navigation
-	if(bool(texelFetch(_control_maps, get_region_uv(floor(uv), SKIP_PASS), 0).r >>1u & 0x1u)) {
+	if(bool(texelFetch(_control_maps, get_region_uv(floor(uv + 0.5), SKIP_PASS), 0).r >>1u & 0x1u)) {
 		ALBEDO *= vec3(.5, .0, .85);
 	}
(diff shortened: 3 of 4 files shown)


TokisanGames:
Thank you. Does this address the first case where it captures the top left vertex, but not the bottom right vertex? It looks like this is a cosmetic fix, but the navigation server should receive the bottom right vertices.

Xtarsia:
I don't think those edges exist. Its the 0 to (region_size - 1) indexing thing again.

This is 9 64m regions, 5 of which are filled entirely with nav (with only the offset in the shader).

image

I think we should be drawing the region gizmo, and the debug view grid with the 0.5 offset as well.

I did test the changes to _generate_triangles() but it had no effect, as it would just lookup that location, find no region (and thus no nav painted), and skip generating that triangle pair.

Xtarsia:
This might do the trick? 
image

TokisanGames:
I baked a 64m terrain shape. On the +X, +Z corner, the navigation shader misses the final vertices. The baked navigation misses the last two vertices.

{99C830CA-B4E7-49C7-B417-8E697E962F3C}

The -X, -Z corner is a lot closer

{413ABF24-676E-4806-9C79-213701220393}

This lines up the debug shader with the currently generated perfectly:
`if(bool(texelFetch(_control_maps, get_region_uv(uv*1.03-0.25, SKIP_PASS), 0).r >>1u & 0x1u)) {`

I made a 64m box, baked navigation on it. Top left corner looks the same as ours:

{9C09D8DF-348B-4BB0-9AE4-F7944BFAC737}

Bottom right corner goes 1 more vertex:

{869567C1-4B96-4EE5-9F50-F3193532388A}

I compared with main, and I don't see a difference with the baked mesh. Testing the function, the ORs and the ANDs both produce the exact same result on this nav mesh, since every vertex is set.

I also baked a LOD0 array mesh, and sure enough it's also missing the last row. Our generate triangles function is wrong.

Mesh, Nav and Collision all need information from the next vertex, which may be in the next region. Collision duplicates the last pixel if not found, which these need to do as well.

I'm working on duplicating the last vertex, and adjusting the shader to match.

TokisanGames:
> I think we should be drawing the region gizmo, and the debug view region grid with the 0.5 offset as well.

This sounds like the [discussion](https://github.com/TokisanGames/Terrain3D/discussions/464) we started but never really finished. I wanted to make images for all of the region layers but never got around to it. I've unlocked the discussion. Please elaborate on your idea there. We the region gizmos, I'd love to dump them in favor of #100 if we can figure out how to render it when using NONE.

TokisanGames:
@Xtarsia This is ready for another review.

It now extends the triangle pairs out to the edge of the mesh, making both `ArrayMesh` baking and `navigation` match the mesh and collision.

Triangles on the region boundaries get extended, while those on the edge of holes do not.

This matches baking a navmesh on a plane pretty closely.

`NavigationMesh/Agents/Radius` can be set to 0 for basically a full mesh bake for testing, but the default is good for operation.

{8A3EFAF3-9734-4CE0-A658-F59B38FCB25D}

TokisanGames:
Works fine in 4.4rc3. Merging.

PR #603, opened 2025-01-22
fix region edge behaviour

Xtarsia:
When background mode is none, get_region_uv() will return the +xz edges from 1 or 2 additional region_texels into empty regions to provide filler data for bridging vertices, and normal calculations.

Set NAN instead of 0.0 when collision reads from edge data if adjacent regions are empty

Average sculpt duplicates source entry for average instead of 0.0 at offsets along edges of regions for more intuitive bahvaior.

image

### DIFF
--- a/project/demo/CodeGeneratedDemo.tscn
+++ b/project/demo/CodeGeneratedDemo.tscn
@@ -18,10 +18,10 @@ script = ExtResource("1_h7vyv")
 [node name="Environment" parent="." instance=ExtResource("3_71ikj")]
 
 [node name="Player" parent="." instance=ExtResource("2_3v2uf")]
-transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 200, 0)
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 50, 0)
 
 [node name="Enemy" parent="." node_paths=PackedStringArray("target") instance=ExtResource("4_p8qry")]
-transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 10, 200, -10)
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 10, 50, -10)
 target = NodePath("../Player")
 
 [node name="UI" parent="." instance=ExtResource("4_x5ge4")]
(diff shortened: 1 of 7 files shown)


Xtarsia:
Will need to check debug views etc

TokisanGames:
Really, really nice @Xtarsia. Users can change cull_disabled in the shader to see underneath.

* control_angle, control_scale, autoshader are broken on `ivec3 __auv = get_region_uv(floor(uv));`

* navigation broken on `if(bool(texelFetch(_control_maps, get_region_uv(floor(uv)), 0).r >>1u & 0x1u)) {`

* I see what you did with smoothing, which is an improvement for NONE, but it's not intuitive why smoothing doesn't work to soften the edge of a ridge pushed up against a region boundary for the other background modes. We still assume sea level is at 0, which is fine for now, but I think we should smooth to sea level(0) for modes other than NONE. How you have it for NONE is fine.

* So the shader is copying the last vertex? Why not also copy the last vertex for collision so they are pixel perfect?

image

* Not a big deal, but this -x,+z and +x,-z corners are odd. Cut off and dark normal. In the demo as well if non-zero height.

{518D8AC4-14BA-426D-B0D1-A08C64BC9554}


* `Raising` heights along +X or +Z makes a strange lip 

{FE88697B-1D49-4AF7-82B1-5CA62AB63DD1}

Maybe it's just a copy of the previous value:

{5F489902-3DA3-4813-A9FE-CD41F62CCDF8}

But at some camera angles it looks lower:

{A269FA2A-ACDB-4134-9CFC-B2FE20AB3D85}

And other camera angles it looks higher: 🫤 

{6BA55675-D3DF-4C83-AFA7-7FE409521916}

TokisanGames:
Also I'm not opposed to us never providing a variable sea level, and always assuming it's 0. All GIS data and real world heightmaps define it at 0. Our heights are always real. If they want to change height data, we can provide a data modification tool in the menu.

Xtarsia:
Search direction is now aligned with either only X or Z, unless exactly diagonal (1 corner) - the duplicated edges are now not read from a diagonal anymore.

Updated the code generated demo.

The missing edges on 2 of the 4 corners is abit problematic, without doing as much branching as the collision is doing.

Now pulling a similar trick for collision

cirehildebrand:
Capture
Capture2

Adding region creates blend instead of vertical connection. Maybe intended but wanted to point out in case.

TokisanGames:
Very nice work. Thanks!

A fix for clipped corners is documented in https://github.com/TokisanGames/Terrain3D/discussions/608.

Issue #321, opened 2024-02-11
Configuration warnings may change implementation


TokisanGames:
### Update

The PRs that broke configuration warnings were reverted. 

A new design is in process for 4.4 at the earliest:
Proposal https://github.com/godotengine/godot-proposals/issues/9098
PR https://github.com/godotengine/godot/pull/90049

------
### Original

Configuration warnings changed from PackedStringArray to an Array that accepts a mix of Strings and Dictionaries { message: String, property?: String }.
If property is set, then show a clickable warning icon on property in the inspector.
Strings in this array are converted to an equivalent Dictionary with message filled.

https://github.com/godotengine/godot/pull/68420

m4gr3d:
Note that https://github.com/godotengine/godot/pull/68420 causes the `master` build of the Android editor to crash when loading the Terrain3D sample.

TokisanGames:
This is a non-issue until the pending Godot PRs are actually merged. We can reopen this then.

PR #604, opened 2025-01-25
Implement Instancer LODs

### LINKED ISSUE #617: Using the same .tscn with the same .mesh multiple times in the instancer combines them into only one foliage object.
### Terrain3D version

0.9.3a

### System information

Godot v4.3.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 Laptop GPU (NVIDIA; 32.0.15.6636) - 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (16 Threads)

Not applicable

### Issue description

My use case:
I have three differently colored types of bushes all using the same mesh.
Therefore, I load the .tscn that contains the mesh 3x into the foliage instancer each time with a differently colored material.

However, in the end, all three of these foliage slots show the same object with the same material. Painting in the world also paints only one of the three variations. 

Workaround:
duplicate the bush .tscn and give each foliage slot a different .tscn to refer to, even when they use the same mesh.

### Logs

_No response_

laurentsenta:
Admin edit
Fixes #617 

* Adds support for up to 10 LODs to the instancer
* Adds the ability to hide all instances `rendering/show_instances`
* Adds the ability disable MMI generation for individual meshes

-----

Contributes to https://github.com/TokisanGames/Terrain3D/issues/43
Introduce four levels of LODs for multimesh instances and min / max shadow lods.

See thread in [Discord](https://discord.com/channels/691957978680786944/1130291534802202735/1332695349705441301)

Task list:

- [x] Decide on using an array or hardcoded floats for the distances
- [x] Actually use the LOD'd mesh in the instancer
- [x] Shadow LOD
- [x] figure out the visibility margin thingy
- [x] Fix the "everything resets on loading" issue
- [x] Remove visibility range, and figure out upgrade path
- [x] Field validation
- [x] Update
- [x] Cleanup
- [x] Shadow min / max config
- [x] Field validation (bis)
- [x] Update the doc
- [x] Also update the notes on "in the future we'll chunk MMIs" this is already implemented

CleanShot 2025-01-25 at 19 36 50

### Regarding validation:
This PR introduces a bunch of fields that are related to each other:

- maximum lod value is valid if the scene is loaded and must be less than mesh count
- maximum shadow lod is valid if <= maximum lod and >= minimum shadow lod (and vice versa for minimum shadow lod)
- Cast Shadows validity depends on the shadow lods:
   - if cast shadow is double-sided or shadows only, it will interact weirdly with min/max shadow lods

Checking values in the setters will be super annoying to use, as in "I'm trying to lower the maximum lod, but nothing happens" > "yes, you need to lower the minimum shadow lod first".

This is a perfect use case for configuration warning, which is a work in progress:
- https://github.com/TokisanGames/Terrain3D/issues/321
- https://github.com/godotengine/godot/pull/90049


src/terrain_3d_mesh_asset.h
```diff
@@ -38,6 +38,13 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
 	Ref<PackedScene> _packed_scene;
 	Ref<Material> _material_override;
 	real_t _density = 10.f;
+	int _maximum_lod = 0;
+	int _shadow_lod = 0;
+
+	float _lod_0_distance_begin = 0.f;
+	float _lod_1_distance_begin = 0.f;
+	float _lod_2_distance_begin = 0.f;
+	float _lod_3_distance_begin = 0.f;
```
laurentsenta, line 47: To be discussed, it might be better to use an array and expose getters / setters for the inspector's UI

CleanShot_2025-01-25_at_17 32 452x

src/terrain_3d_instancer.cpp
```diff
@@ -85,55 +85,59 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
 					_terrain->get_mmi_parent()->add_child(node, true);
 				}
 
-				// Retrieve MMI or create one
+				// Retrieve MMI or create one for every lod
 				MeshMMIDict &mesh_mmi_dict = _mmi_nodes[region_loc];
 
-				int lod = 0; // TODO Hard coded LOD0 for now
```
laurentsenta, line 91: Standing on the shoulders of giants here, it feels like everything was prepared for this PR.

src/terrain_3d_mesh_asset.h
laurentsenta: (used array)

src/terrain_3d_mesh_asset.h
```diff
@@ -27,9 +27,11 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
 	};
 
 private:
+	static constexpr int MAX_LOD_COUNT = 4;
+
 	// Saved data
 	real_t _height_offset = 0.f;
-	real_t _visibility_range = 100.f;
+	real_t _visibility_range = 100.f; // TODO: drop and discuss storage migration
```
laurentsenta, line 34: Need maintainer input: should we propose an upgrade path, how?

src/terrain_3d_mesh_asset.cpp
```diff
@@ -203,6 +212,62 @@ void Terrain3DMeshAsset::set_cast_shadows(const GeometryInstance3D::ShadowCastin
...
+	_maximum_lod = CLAMP(p_lod, 0, MIN(MAX_LOD_COUNT - 1, get_mesh_count() - 1));
+	LOG(INFO, "Setting maximum LOD: ", _maximum_lod);
+	emit_signal("instancer_setting_changed");
+}
+
+void Terrain3DMeshAsset::set_shadow_lod(const int p_lod) {
+	_shadow_lod = CLAMP(p_lod, 0, get_maximum_lod());
+	LOG(INFO, "Setting shadow LOD: ", _shadow_lod);
+	emit_signal("instancer_setting_changed");
+}
```
laurentsenta, line 225: Need maintainer input: every time I reopen the project, the maximum lod and shadow lod values are reset to -1,
I think this is because the mesh is not loaded when the editor tries to set the instance's value.

If you change the values and then close and reopen the scene, that'll work.

Is there an idiomatic way to avoid this?

src/terrain_3d_mesh_asset.h
TokisanGames: Yes.
* Leave the old variable.
* Change the setter to call the new setter.
* Move both to the bottom of the file headed with a comment: `//DEPRECATED 1.0 - Remove 1.1`
* Remove the getter.
* _bind_methods(): Move the bound variables/functions to the bottom and add the comment.
* _bind_methods(): change the property usage to no_editor so it doesn't show in the inspector, eg.
`ADD_PROPERTY(PropertyInfo(Variant::ARRAY, "region_offsets", PROPERTY_HINT_ARRAY_TYPE, vformat("%tex_size/%tex_size:%tex_size", Variant::VECTOR2, PROPERTY_HINT_NONE), PROPERTY_USAGE_NO_EDITOR), "set_region_offsets", "get_region_offsets");`

* In the XML docs note it is deprecated and to see the new function.

laurentsenta: Done in https://github.com/TokisanGames/Terrain3D/pull/604/commits/33a46e3a599ec2c78a94196040f05609240e219d

src/terrain_3d_mesh_asset.cpp
TokisanGames: When Godot loads the scene file it loads all settings from the scene file. You don't really know which one it will load first. If this is loaded before the scene, get_mesh_count() will be zero, which sets your _maximum_lod = -1.

If packed_scene.is_null() then you shouldn't check get_mesh_count() and accept the value as it is. It won't be used while the scene is null.

When packed_scene is loaded, you can verify that maximum_lod is valid. Same w/ shadow lod.

laurentsenta: Done in https://github.com/TokisanGames/Terrain3D/pull/604/commits/fcaccb99a3bb2bb0b4cc93c7f97ad0fd8470e8b8

Xtarsia (commented):
Very nice work overall.

Cory will be along asking for a good squash no doubt..

The demo changes can likley be dropped unless you want to spend time tidying them up.

Docs effort is stellar!

TokisanGames:
> Cory will be along asking for a good squash no doubt..

It can be squashed down to 3 commits: code, docs, demo changes.

I'll review in a couple of days after Xtarsia is satisfied and I get more progress on my collision commit.

src/terrain_3d_instancer.cpp
```diff
@@ -85,60 +84,93 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
 
-				// Create MM and assign to MMI
-				mmi = cell_mmi_dict[cell];
-				mmi->set_multimesh(_create_multimesh(mesh_id, xforms, colors));
+					if (cell_mmi_dict.count(cell) == 0) {
+						mmi = memnew(MultiMeshInstance3D);
+						LOG(DEBUG, "No MMI found, Created new MultiMeshInstance3D: ", uint64_t(mmi));
+						// Node name is MMI3D_Cell##_##_Mesh#
+						String cstring = "_C" + Util::location_to_string(cell).trim_prefix("_");
+						mmi->set_name("MMI3D" + cstring + "_M" + String::num_int64(mesh_id));
```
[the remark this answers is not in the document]
laurentsenta: Fixed in https://github.com/TokisanGames/Terrain3D/pull/604/commits/0c39ac0cbef71fe118b154865c0de19b4338758c

src/terrain_3d_instancer.cpp
```diff
@@ -85,60 +84,93 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
+	real_t lod_end = p_ma->get_lod_visibility_range_end(p_lod);
+	if (lod_end > 0.0f) {
+		p_mmi->set_visibility_range_end(lod_end);
+	}
+
+	real_t visibility_margin = p_ma->get_visibility_margin();
+	if (visibility_margin > 0.0f) {
+		p_mmi->set_visibility_range_begin_margin(visibility_margin);
+		p_mmi->set_visibility_range_end_margin(visibility_margin);
+	}
```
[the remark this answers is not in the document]
laurentsenta: Fixed in https://github.com/TokisanGames/Terrain3D/pull/604/commits/e2df2ff53357bb78d001dda62a6e50a8a0160d2c

Saul2022:
Looking great,  just one curiosity, so to add  lods, we do it like we usually do with visibility ranges and hlod.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -300,6 +395,52 @@ void Terrain3DMeshAsset::set_generated_size(const Vector2 &p_size) {
 	}
 }
 
+GeometryInstance3D::ShadowCastingSetting Terrain3DMeshAsset::get_lod_cast_shadows(const int p_lod_id) const {
+	// Shadow casting is disabled, no question asked
+	if (get_cast_shadows() == GeometryInstance3D::SHADOW_CASTING_SETTING_OFF) {
```
TokisanGames, line 400: * Use member variables internally throughout, so we know nothing else is going on when retrieving the value. Change to `_cast_shadows`.
* Remove `no question asked`

src/terrain_3d_mesh_asset.cpp
```diff
@@ -300,6 +395,52 @@ void Terrain3DMeshAsset::set_generated_size(const Vector2 &p_size) {
...
+		return GeometryInstance3D::SHADOW_CASTING_SETTING_OFF;
+	}
+	// That lod id is strictly reserved for shadow casting at min shadow lod and below
```
TokisanGames, line 403: //This lod id is for the shadow impostor and is set to cast shadows only.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -300,6 +395,52 @@ void Terrain3DMeshAsset::set_generated_size(const Vector2 &p_size) {
...
+	if (p_lod_id == SHADOW_LOD_INSTANCE) {
+		return GeometryInstance3D::SHADOW_CASTING_SETTING_SHADOWS_ONLY;
+	}
+	// That lod is in the special case where each lod uses its own shadow
```
TokisanGames, line 407: This case is not where each lod uses it's own shadow. The condition as written out is only when lod0 uses its own shadow. Is that the condition we want? Any lod should be able to use its own shadow, as long as lod is <= max_shadow_impostor_lod (aka minimum_shadow_lod).

src/terrain_3d_mesh_asset.cpp
```diff
@@ -300,6 +395,52 @@ void Terrain3DMeshAsset::set_generated_size(const Vector2 &p_size) {
...
+	if (p_lod_id == 0 && p_lod_id == get_minimum_shadow_lod()) {
+		return get_cast_shadows();
+	}
+	// That lod relies on the min shadow lod
+	if (p_lod_id <= get_minimum_shadow_lod()) {
```
TokisanGames, line 412: // This lod uses the shadow impostor lod

src/terrain_3d_instancer.cpp
```diff
@@ -85,60 +84,92 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
-				triple[2] = false;
+					// Set the cell modified state to false
+					triple[2] = false;
+				}
 			}
 		}
 	}
 }
 
+void Terrain3DInstancer::_setup_mmi_lod(MultiMeshInstance3D *p_mmi, const Ref<Terrain3DMeshAsset> &p_ma, const int p_lod) {
```
TokisanGames, line 159: This should be named _setup_mmi_lod_ranges

src/terrain_3d_instancer.cpp
```diff
@@ -85,60 +84,92 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
+						modified = true;
+					}
+					// If data hasn't changed since last _update_mmis, skip
+					if (modified == false) {
+						continue;
+					}
+
+					// Create MM and assign to MMI
+					mmi = cell_mmi_dict[cell];
+					mmi->set_multimesh(_create_multimesh(mesh_id, lod, xforms, colors));
```
TokisanGames, line 139: This looks like you are creating 5 MultiMeshes for Shadow, and LOD0-3. So that's 5 meshes (necessary) and 5x the transforms and colors (MMI internals copy incoming data so we can't share it) to display 4 meshes.

You could instead reverse the for loop to count from last_lod to SHADOW_LOD. Then when creating the shadow MMI, grab the already created LOD# Multimesh reference and insert it instead of creating an identical copy.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -359,18 +515,36 @@ void Terrain3DMeshAsset::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("set_mesh_count", "count"), &Terrain3DMeshAsset::set_mesh_count);
 	ClassDB::bind_method(D_METHOD("get_thumbnail"), &Terrain3DMeshAsset::get_thumbnail);
 
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "name", PROPERTY_HINT_NONE), "set_name", "get_name");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "id", PROPERTY_HINT_NONE), "set_id", "get_id");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "height_offset", PROPERTY_HINT_RANGE, "-20.0,20.0,.005"), "set_height_offset", "get_height_offset");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "density", PROPERTY_HINT_RANGE, ".01,10.0,.005"), "set_density", "get_density");
-	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "visibility_range", PROPERTY_HINT_RANGE, "0.,4096.0,.05,or_greater"), "set_visibility_range", "get_visibility_range");
-	//ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "visibility_margin", PROPERTY_HINT_RANGE, "0.,4096.0,.05,or_greater"), "set_visibility_margin", "get_visibility_margin");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "visibility_margin", PROPERTY_HINT_RANGE, "0.,4096.0,.05,or_greater"), "set_visibility_margin", "get_visibility_margin");
```
TokisanGames, line 525: Sort all of the lod settings under the scene file, and put this above those properties to put them in a group `	ADD_GROUP("LODs", "");`. I don't know, but you might be able to then add `ADD_GROUP("", "");` to get out of the group.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -203,6 +207,89 @@ void Terrain3DMeshAsset::set_cast_shadows(const GeometryInstance3D::ShadowCastin
...
+int Terrain3DMeshAsset::get_valid_lod() const {
+	// As long as the packed scene was not loaded, we don't enforce the LOD ids
+	if (_packed_scene.is_null()) {
+		return MAX_LOD_COUNT - 1;
+	}
+
+	return MIN(MAX_LOD_COUNT - 1, get_mesh_count() - 1);
+}
+
+void Terrain3DMeshAsset::refresh_valid_lods() {
```
TokisanGames, line 219: * This function should be called _validate_lods()
* Use internal names, not functions.

TokisanGames (changes requested):
Alright, very good work on this. Thank you for the effort. Please rebase it as you address the notes.

* Playing with the options and thinking about the names, I think they can use some adjustments to be clearer. I'd prefer to move away from "higher, lower, minimum, maximum" because lods numbers are a reverse scale, where higher/better is a lower number. The docs say "lods above this...", does "above" mean "larger number / lower detail"? Or "higher detail / lower number"? Does "maximum lod" mean "maximum distance / lowest detail" or "maximum detail / lowest distance"? I think "last" is not only shorter, but more universal in meaning lowest detail, farthest away.
  * `maximum_lod` means "don't show lods lower than this lod". I think `last_lod` is a little better. It's also `lod_count-1`.
  * `maximum_shadow_lod` means "show shadows for this lod and higher(0) - ie don't show shadows for lods lower than this value". This would be better named `last_shadow_lod`.
  *  `minimum_shadow_lod` means "use this shadow impostor lod or lower(3) - ie when showing a lod higher(0) than this lod, use this one as its shadow impostor". This would be better named `max_shadow_imposter_lod`, or `max_shadow_lod`.

* If `minimum_shadow_lod` is > `maximum_shadow_lod`, the latter doesn't work. Or more clearly, if `max_shadow_imposter_lod` > `last_shadow_lod`, then it shows shadows beyond the `last_shadow_lod`. Given the above descriptions it should not. It should only show shadows for lods up to `last_shadow_lod`.

* I actually don't think we need a `maximum_shadow_lod`. Lights already have a shadow distance. We don't need to cull it ourselves. Having it prevents us from needing to create and store the MMI, but that generation is quick and only on start, and the memory should be shared with my related note elsewhere.

In my specifications here on [discord](https://discord.com/channels/691957978680786944/1332695349705441301/1332797693986209834) I expressed the need for `maximum_lod(last_lod)` and `shadow_lod`. The lods should show shadows at all distances that the user has configured their Lights for, it just may be at a lower level LOD.

* `MeshAsset/Cast Shadows / shadows only` doesn't work. `Shadows Off` does work.

* MeshAsset set_scene_file doesn't do any filtering or sorting on meshes named LOD0-LOD3. It happens that sometimes the LODS might be out of order for an artist or an engine reason. I would look for all of the mesh instances that end in LOD#, then sort them and use the first 4. If I couldn't find any LODs, then just take the first original 4. Maybe print some messages saying, "found 8 lods, using the first 4." or "Meshes don't end with LOD#, assuming first 4 are LOD0-LOD3".

* MeshAsset set_scene_file - You could also fix a bug where if they provide a scene with a MeshInstance3D as the root node, it won't use it. Probably because the find_child starts at the scene root, and ignores the scene root.

doc/doc_classes/Terrain3DMeshAsset.xml
```diff
@@ -69,8 +69,38 @@
...
+		<member name="lod_3_visibility_range" type="float" setter="set_lod_3_visibility_range" getter="get_lod_3_visibility_range" default="0.0">
+			See [lod_0_visibility_range]. Applies to the fourth and last LOD level.
+		</member>
+		<member name="visibility_margin" type="float" setter="set_visibility_margin" getter="get_visibility_margin" default="0.0">
+			Sets the [code skip-lint]GeometryInstance3D.set_visibility_range_begin_margin[/code] and [code skip-lint]GeometryInstance3D.set_visibility_range_end_margin[/code] on all multimeshes used by this mesh.
+			We use Godot's default `Disabled` Fade mode, which uses hysteresis to switch between LOD levels. This prevents situations where LOD levels are switched back and forth quickly when the player moves around a transition point.
+		</member>
 		<member name="visibility_range" type="float" setter="set_visibility_range" getter="get_visibility_range" default="100.0">
 			Sets [code skip-lint]GeometryInstance3D.visibility_range_end[/code] on all MultiMeshInstances used by this mesh. Allows the renderer to cull MMIs beyond this distance. Set to 0 to disable culling.
+			Deprecated, see [lod_0_visibility_range].
```
laurentsenta, line 103: (adding a note here to have a discussion thread)

@TokisanGames, regarding your comment on naming,

> Playing with the options and thinking about the names, I think they can use some adjustments to be clearer. I'd prefer to move away from "higher, lower, minimum, maximum" because lods numbers are a reverse scale, where higher/better is a lower number. The docs say "lods above this...", does "above" mean "larger number / lower detail"? Or "higher detail / lower number"? Does "maximum lod" mean "maximum distance / lowest detail" or "maximum detail / lowest distance"? I think "last" is not only shorter, but more universal in meaning lowest detail, farthest away.
> maximum_lod means "don't show lods lower than this lod". I think last_lod is a little better. It's also lod_count-1.
> maximum_shadow_lod means "show shadows for this lod and higher(0) - ie don't show shadows for lods lower than this value". This would be better named last_shadow_lod.
> minimum_shadow_lod means "use this shadow impostor lod or lower(3) - ie when showing a lod higher(0) than this lod, use this one as its shadow impostor". This would be better named max_shadow_imposter_lod, or max_shadow_lod.

Agreed; based on your notes, what if we used distances-related naming?
 We wouldn't mix the lexical field of ordering (last/first) and scale (min/max) but use something like "closest" and "furthest"?

Does `closest_lod`, `furthest_lod`, and `closest_shadow_lod` sound self-explanatory?

Xtarsia: perhaps near / far are more concise I think: `near_lod` , `far_lod` , `near_shadow_lod` ?

TokisanGames: I don't recall seeing where we are referring to lod0 by a descriptor, only the final/last/farthest lod. If we need to refer to lod0 and the farthest lod by descriptors then near_lod and far_lod are good.

However, my thoughts on the main variables:
* maximum_shadow_lod should be cut, and let the renderer cull shadows on the farthest objects. And it simplifies our code. 

* The final lod used in lod0, 1, 2, n. Without a corresponding near, I would use last_lod instead of far. "last_lod=X means of the 7 lods in the file, lod X is the last one used." vs "far_lod=X means of the 7 lods in the file, lod X is the farthest one used." The former seems a bit clearer. Far suggests there's only one far, and the other is a near. Whereas last also says there's only one, but doesn't suggest the others are limited in number. 

* minimum_shadow_lod (max highest resolution shadow impostor) could be near_shadow_lod, max_shadow_lod, first_shadow_lod. Actually, I like think first is clearest, and contrasts with last above.


7 lods, 0-6
Last_lod=3
First_shadow_lod=2
Without thinking too hard I know lods 4-6 won't be used at all, and only lods 2-3 will be used for shadow impostors. With all the other naming schemes my mind has to twist around to think about what the dev thought the word means when he named it.

src/terrain_3d_mesh_asset.h
```diff
@@ -88,6 +121,7 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
 
 	Ref<Mesh> get_mesh(const int p_index = 0);
```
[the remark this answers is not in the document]
laurentsenta: That's done in `get_lod_mesh`, I kept the lod-picking related code to an explicit function

src/terrain_3d_mesh_asset.h
```diff
@@ -67,11 +79,32 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
...
+
+	void set_lod_visibility_range(const int p_lod, const real_t p_distance);
+	void set_lod_0_visibility_range(const real_t p_distance);
+	real_t get_lod_0_visibility_range() const { return _lod_visibility_ranges[0]; }
+	void set_lod_1_visibility_range(const real_t p_distance);
+	real_t get_lod_1_visibility_range() const { return _lod_visibility_ranges[1]; }
+	void set_lod_2_visibility_range(const real_t p_distance);
+	real_t get_lod_2_visibility_range() const { return _lod_visibility_ranges[2]; }
+	void set_lod_3_visibility_range(const real_t p_distance);
+	real_t get_lod_3_visibility_range() const { return _lod_visibility_ranges[3]; }
```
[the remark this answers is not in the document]
laurentsenta: Could you confirm you meant to use a double underscore? so we'd use `get_lod0__range` and NOT `get_lod0_range`?

src/terrain_3d_mesh_asset.cpp
laurentsenta: When I tried `ADD_GROUP("", "")`, the fields disappeared.

Proposal with a group for the mesh settings, which disappears when empty / unset:





If we rename all the lod-related fields to `lod_....` I can implement a similar filter, that would appear / disappear based on the generated type and the number of lods in the scene.

src/terrain_3d_mesh_asset.cpp
laurentsenta: > Use internal names, not functions.

Do you mean writing code like `_maximum_lod = value` instead of using getters/setters?

The code is relying on the setter's validation code, like `_maximum_lod = CLAMP(p_lod, 0, get_valid_lod());`
that was a way to avoid duplicating it.

doc/doc_classes/Terrain3DMeshAsset.xml
laurentsenta: first and last it is 👍 

since we're on naming, see my comment in:
https://github.com/TokisanGames/Terrain3D/pull/604#discussion_r1942420841

We could use `lod_` as a prefix to filter properties and show / hide them when necessary,

```
lod_visibility_margin
lod_0_visibility_range
lod_1_visibility_range
...
lod_3_visibility_range
lod_last
lod_first_shadow
lod_last_shadow (should we keep this tbd)
```

I wouldn't implement the field filtering in that PR (it has grown too much already), but getting the naming right is free since I'll have to do a renaming pass already.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -359,18 +515,36 @@ void Terrain3DMeshAsset::_bind_methods() {
...
+
+	//DEPRECATED 1.0 - Remove 1.1
+	ClassDB::bind_method(D_METHOD("set_visibility_range", "distance"), &Terrain3DMeshAsset::set_visibility_range);
+	ClassDB::bind_method(D_METHOD("get_visibility_range"), &Terrain3DMeshAsset::get_lod_0_visibility_range);
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "visibility_range", PROPERTY_HINT_RANGE, "0.,4096.0,.05,or_greater", PROPERTY_USAGE_NO_EDITOR), "set_visibility_range", "get_visibility_range");
+}
+
+//DEPRECATED 1.0 - Remove 1.1
+void Terrain3DMeshAsset::set_visibility_range(const real_t p_visibility_range) {
+	set_lod_0_visibility_range(p_visibility_range);
```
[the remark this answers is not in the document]
TokisanGames: I would say the current behavior is to show _the_ mesh up to visibility range. Beyond that visibility range it hides the mesh. Changing this to lod0 range, when lod1-3 exist, and presumably lod1-3 ranges are higher, then upgrading this value to lod0 range won't hide the mesh beyond that range. Setting the last lod visibility range will most closely match the old behavior of keeping the object visible until that range. 

A more accurate change would be to set the range of the last lod, rather than lod3 or lod0. However when this loads, the lods may not yet be loaded, so you don't necessarily know which lod is last. Ultimately this is an unimportant setting to retain, so even resetting it to default is fine.

get_visibility_range() can go away, as the property is hidden. Only the setter will be called if loaded in the scene file.

src/terrain_3d_mesh_asset.h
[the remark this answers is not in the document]
TokisanGames: The double was a typo.

src/terrain_3d_mesh_asset.cpp
TokisanGames: Here in the validation function it's fine, but elsewhere it should be internal names. There are other uses where it's calling a getter function unnecessarily.

doc/doc_classes/Terrain3DMeshAsset.xml
TokisanGames: Filtering inspector fields is currently implemented with `begins_with("generated_")`, but you can see we control the filter. They could also be hidden with `contains("lod")`, or is in an array of names. Because of the confusing nature of the names as described elsewhere, I'd rather these names be clear.

We can keep last_shadow_lod.

doc/docs/instancer.md
```diff
@@ -75,7 +76,7 @@ There are some caveats and limitations built into the engine that you should be
 
 ### Simple Objects
 
-Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes support only one mesh. Our system scans the provided scene file and uses the first mesh it finds. If you give it a complex object with separate trunk and leaves, it will use only the trunk.
+Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes support only one mesh per LOD. Our system scans the provided scene file and uses the first mesh it finds for LOD0, the second mesh as LOD1, and so on up to LOD3. If you give it a complex object with separate trunk and leaves, it won't work as expected.
```
laurentsenta, line 79:
```suggestion
Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes supports only one mesh per LOD. If you give it a complex object with separate trunk and leaves, it won't work as expected.

Our system scans the provided scene file for _reasonable_ LOD structures:

1. If it finds meshes with the `LOD#` suffix in their names (`MyMeshInstanceLOD0`, `MyMeshInstanceLOD1`, etc), it'll order by LOD number and pick the first ones.
2. Fallback to sorting all the meshes in the scene by name,
3. Finally, fallback to using the root node as `LOD0` if that's a Mesh Instance.

Terrain3D supports four levels of detail.
```

doc/docs/instancer.md
```diff
@@ -113,6 +116,50 @@ Godot currently has no protection against filling up your VRAM. You could do so
...
+
+#### Artist Created LODs
+
+We also provide a way to insert artist created LODs, stored as separate meshes within your scene file. We generate a multimesh for each LOD, from LOD0 up to LOD3. The visiliby range of each LOD level can be set in the instancer options. The engine will automatically switch between LOD levels based on the distance from the camera.
+
+<figure class="video_container">
+ <video width="600px" controls="true" allowfullscreen="true">
+ <source src="../_static/video/lod-001-base.mp4" type="video/mp4">
+ </video>
+</figure>
```
laurentsenta, line 133:
```suggestion
</figure>

Terrain3D supports four levels of detail.

Our system scans the provided scene file for _reasonable_ LOD structures:

1. If it finds meshes with the `LOD#` suffix in their names (`MyMeshInstanceLOD0`, `MyMeshInstanceLOD1`, etc), it'll pick the first ones, ordered by their LOD number.
2. Else, it fallbacks to pick the first meshes in the scene, ordered by name,
3. Finally, we fallback to using the root node as `LOD0` if that's a Mesh Instance.
```

doc/docs/instancer.md
```diff
@@ -75,7 +76,15 @@ There are some caveats and limitations built into the engine that you should be
...
-Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes support only one mesh. Our system scans the provided scene file and uses the first mesh it finds. If you give it a complex object with separate trunk and leaves, it will use only the trunk.
+Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes supports only one mesh per LOD. If you give it a complex object with separate trunk and leaves, it won't work as expected.
+
+Our system scans the provided scene file for _reasonable_ LOD structures:
+
+1. If it finds meshes with the `LOD#` suffix in their names (`MyMeshInstanceLOD0`, `MyMeshInstanceLOD1`, etc), it'll order by LOD number and pick the first ones.
+2. Fallback to sorting all the meshes in the scene by name,
+3. Finally, fallback to using the root node as `LOD0` if that's a Mesh Instance.
+
+Terrain3D supports four levels of detail.
```
laurentsenta, line 87:
```suggestion
Some 3D assets are complex objects with multiple, separate meshes, such as a tree trunk and leaves, a chest and lid, or a door frame and door. MultiMeshes supports only one mesh per LOD. If you give it a complex object with separate trunk and leaves, it won't work as expected.
```

doc/docs/instancer.md
```diff
@@ -113,6 +116,50 @@ Godot currently has no protection against filling up your VRAM. You could do so
...
+
+In the following video, note how shadows are using the (simpler) cube mesh for LODs 0 and 1 when we set the `Minimum Shadow LOD` to 2.
+
+<figure class="video_container">
+ <video width="600px" controls="true" allowfullscreen="true">
+ <source src="../_static/video/lod-003-shadow-minimum-lod.mp4" type="video/mp4">
+ </video>
+</figure>
+
+Be wary of double checking your LODs configuration in the editor, as we don't validate nonsensical settings for now (e.g. LOD0 range is greater than LOD1 range, or `Maximum Shadow LOD` is less than `Minimum Shadow LOD`). In the future we may provide warning once [Configuration Warnings for Resources](https://github.com/godotengine/godot/pull/90049) makes its way into the engine.
```
laurentsenta, line 159:
```suggestion
When using artist-created LODs, we provide various options for shadow LODs.

The `Last Shadow LOD` option lets you set the last LOD level that will cast shadows. Objects with higher LOD levels (further away) will not cast shadows. This is used to improve performance in cases where shadows are not visible in the distance (blades of grass, for example).

In the following video, note how the shadows disappear as we lower the `last_shadow_lod` value.

<figure class="video_container">
 <video width="600px" controls="true" allowfullscreen="true">
 <source src="../_static/video/lod-002-shadow-maximum-lod.mp4" type="video/mp4">
 </video>
</figure>

The `First Shadow LOD` option lets you set the first LOD level that will cast shadows. Objects with lower LOD levels (closer) will appear as if they are casting a shadow of this LOD level. This is used to improve performance in cases where shadows don't need as much detail as the object itself.

In the following video, note how shadows use the (simpler) cube mesh for LODs 0 and 1 when we set the `first_shadow_lod` to 2.

<figure class="video_container">
 <video width="600px" controls="true" allowfullscreen="true">
 <source src="../_static/video/lod-003-shadow-minimum-lod.mp4" type="video/mp4">
 </video>
</figure>

Be wary of double-checking your LODs configuration in the editor, as we don't validate nonsensical settings for now (e.g., the LOD0 range is greater than the LOD1 range, or `last_shadow_lod` is less than `first_shadow_lod`). In the future, we may provide a warning once [Configuration Warnings for Resources](https://github.com/godotengine/godot/pull/90049) makes its way into the engine.
```

TokisanGames:
I'll take this over from here. Thanks for all of your efforts.

TokisanGames:
This PR is ready for final testing and code review. After that I'm going to reorder the code and merge it in.

TokisanGames:
Alright this is finished.

@laurentsenta Thank you very much for the first implementation and excellent effort on the docs. Both gave us a solid base to improve on.

@Xtarsia thanks for the reviews and testing.

PR #610, opened 2025-01-30
Correct projection derivative calculations

Xtarsia:
Fixes some minor artifacts with projection enabled.

before:
image

after:
image


TokisanGames:
Thanks!

PR #611, opened 2025-02-01
[4.4] Fix Physics Interpolation

### LINKED ISSUE #470: Physics Interpolation Breaks Terrain
### Terrain3D version

v0.9.2

### System information

4.4 dev1

Not applicable

### Issue description

Turning on Physics Interpolation Breaks the Terrain at runtime while moving, causing 2 issues:
1) Textures smear
2) Clipmaps get detached causing holes

They happen because the position is getting Interpolated. Disabling Interpolation on Terrain3D in question doesn't do anything.

This May be an Issue with how Godot is Handling it's Interpolation, or something internally setup wrong on Terrain3D's end regarding how the Meshes are handled, but I'm not sure Which is the case.

If the Clipmaps are using MultiMesh see https://github.com/godotengine/godot/pull/91818#issuecomment-2295310816 to resolve this issue

### Logs

_No response_

TokisanGames:
* Moves all snapping to the physics frame (no apparent effect, but handling it in process isn't necessary)
* Fixes physics interpolation by resetting it on the meshes

This fix requires a version of Terrain3D built against godot-cpp 4.4 and Godot Engine 4.4. The artifact in this build will work in 4.4 while its available. However, once this PR is merged into main, which is still running on godot-cpp 4.3, devs will have to build Terrain3D from source against godot-cpp 4.4 in order to use Physics Interpolation in 4.4. This will remain the case until we stop supporting 4.3.

Fixes #470

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -103,7 +103,7 @@ void Terrain3D::_initialize() {
  * This is a proxy for _process(delta) called by _notification() due to
  * https://github.com/godotengine/godot-cpp/issues/1022
  */
-void Terrain3D::__process(const double p_delta) {
+void Terrain3D::__physics_process(const double p_delta) {
 	if (!_initialized)
 		return;
 
@@ -140,7 +140,7 @@ void Terrain3D::_grab_camera() {
 		_camera_instance_id = _camera->get_instance_id();
 	} else {
 		_camera_instance_id = 0;
-		set_process(false); // disable snapping
+		set_physics_process(false); // disable snapping
 		LOG(ERROR, "Cannot find the active camera. Set it manually with Terrain3D.set_camera(). Stopping _process()");
 	}
 }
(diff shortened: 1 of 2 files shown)


mrjustaguy:
I confirm the issue is Fixed 👍

TokisanGames:
There's also a `main-godot4.4` branch for CI builds that will loosely track main but has 4.4 godot-cpp.

FeralBytes:
I have tried to fix this by downloading the latest builds, but for me the terrain continues to glitch if Physics Interpolation is enabled. I tried all 3 settings for physics interpolation on the terrain 3D object it's self. I am not sure what I am doing wrong.
I downloaded from here: https://github.com/TokisanGames/Terrain3D/actions/runs/13100790404
and Here: https://github.com/TokisanGames/Terrain3D/actions/runs/13100929282
and I tried from here: https://github.com/TokisanGames/Terrain3D/actions/runs/13312359979

TokisanGames:
None of those builds have both this fix merged in and godot-cpp 4.4. The comment right above yours says you can use the `main-godot-4.4` branch. Try this https://github.com/TokisanGames/Terrain3D/actions/runs/13342595702

F3der1co:
now that Godot 4.4 is out it would be amazing to have a "proper" release that includes this.

TokisanGames:
"Proper" releases will come when appropriate. You can use a nightly build from the main-godot4.4 branch. There's no difference in how they are built from main releases, they only differ in commit. You can watch the roadmap and PRs for progress.

PR #618, opened 2025-02-07
Color map fix region edges

Xtarsia:
Before:
image

After:
image

~~Also dropping mipmaps for the color maps, as texelfetch wont access them.~~

~~Also, this was the last use of uv2 - it is no longer used. making get_region_uv2 redundant as well...~~

### DIFF
--- a/project/addons/terrain_3d/extras/lowpoly_colormap.gdshader
+++ b/project/addons/terrain_3d/extras/lowpoly_colormap.gdshader
@@ -62,8 +62,7 @@ ivec3 get_region_uv(const vec2 uv, const int search) {
 // XY: (0 to 1) coordinates within a region
 // Z: layer index used for texturearrays, -1 if not in a region
 vec3 get_region_uv2(const vec2 uv2) {
-	// Remove texel offset to ensure correct region index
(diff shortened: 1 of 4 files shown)


TokisanGames:
Thank you for fixing these artifacts! 

Please also update lowpoly_colormap.gdshader.

Colormap mipmaps are still needed. Try inserting a satellite image on the colormap. It's noisy, especially when moving. If this doesn't apply to the regular, then the colormap debug shader should be adjusted to not use mipmaps.

{46997426-1A38-4229-B8E9-3CD0817641E2}

If uv2 and get_region_uv2 are truly not needed, they can be stripped out.

Xtarsia:
uv2 gets to live another day! Though i had to remove that 0.5 offset to get things pixel perfect.

new method calculates a region_mip value directly from the derivatives, removing the need for textureQueryLod(), and the bilerp skip is now based off that value instead (log2 of fractional parts is negative, so checking less than 0. for that now.

added bonus, the color map now blends at regions edges with flat/world noise, side effect of the new method it seems:
image

TokisanGames:
Thank you for this update. 

Looks like it will blend into -x, -z, but not +x (Left), +z (Up).  No regions in bottom left, top right.  🤷‍♂️ I'll merge this in anyway. Major improvement.

image

PR #620, opened 2025-02-10
Fix region data reading for linux dedicated server export

Zennii:
Fixes an issue with dedicated server export, where the `.res` terrain3d files end up becoming `.res.remap` files. This only seems to happen on some exports. Removing .remap from the file name lets it load properly

Tested with: Dedicated Server export, Ubuntu 22.04


TokisanGames:
Excellent, thank you!

Issue #191, opened 2023-08-21
Adjust LOD level with viewport resolution to reduce visible LOD transitions at high resolutions


Calinou:
I suggest changing the LOD level according to viewport resolution (using viewport height if using **Keep Height** camera FOV Mode, width otherwise). The `Viewport.scaling_3d_scale` multiplier should also be taken into account to multiply this resolution.

In this example, pop-in is reasonable in 1152×648:

https://github.com/outobugi/Terrain3D/assets/180032/027fdfd0-8ae5-473a-b1c9-9b9ccb25d6a5

However, it's much more jarring in 3840×2160 (watch in fullscreen):

https://github.com/outobugi/Terrain3D/assets/180032/4ffb0c88-d8c1-49c4-8720-630495081e63

There should still be a setting to set a LOD bias to perform futher adjustments, but this adjustment should ensure that visible pop-in is minimized at high resolutions. Godot's own mesh LOD system also adjusts itself to viewport resolution, including `Viewport.scaling_3d_scale`.

Issue #421, opened 2024-07-16
[Feature Request] Alternating mesh quad diagonals for symmetrical grid


Radivarig:
### Description

Alternating diagonals would allow symmetrical deformations and reduce the zigzag triangle spikes.

![terrain_alternating_quad_diagonals](https://github.com/user-attachments/assets/ec52a3f4-d66b-4ab1-8d00-6874706bf1ae)

```gdscript
@tool
extends MeshInstance3D

@export_group("Click to generate")
@export var generate: bool:
    set(value):
        if value:
            _generate_grid()

@export var size = Vector2i(10, 10)

func _ready():
    _generate_grid()

func _generate_grid():
    if mesh:
        mesh = null

    var immediate_mesh = ImmediateMesh.new()
    immediate_mesh.surface_begin(Mesh.PRIMITIVE_TRIANGLES)

    for x in range(size.x):
        for y in range(size.y):
            var top_left = Vector3(x, 0, y + 1)
            var top_right = Vector3(x + 1, 0, y + 1)
            var bottom_left = Vector3(x, 0, y)
            var bottom_right = Vector3(x + 1, 0, y)

            if (x + y) % 2 == 0:
                immediate_mesh.surface_add_vertex(bottom_left)
                immediate_mesh.surface_add_vertex(top_right)
                immediate_mesh.surface_add_vertex(top_left)
                
                immediate_mesh.surface_add_vertex(bottom_left)
                immediate_mesh.surface_add_vertex(bottom_right)
                immediate_mesh.surface_add_vertex(top_right)
            else:
                immediate_mesh.surface_add_vertex(bottom_left)
                immediate_mesh.surface_add_vertex(bottom_right)
                immediate_mesh.surface_add_vertex(top_left)
                
                immediate_mesh.surface_add_vertex(top_left)
                immediate_mesh.surface_add_vertex(bottom_right)
                immediate_mesh.surface_add_vertex(top_right)

    immediate_mesh.surface_end()
    mesh = immediate_mesh

```

TokisanGames:
Collision works using a HeightMapShape. The mesh is perfectly aligned with the collision shape. Your suggestion, as is, will make the collision shape and mesh be out of alignment pretty much everywhere between vertices except on flat areas. We would have to generate our own collision shape. If we did it would be after #278. 

Also I intend to experiment with triangle strips and a meshless terrain. I don't believe that pattern can be made with them.

[...]

PR #622, opened 2025-02-15
Implement Geomorphing, Circular LODs, Symmetric grid

### LINKED ISSUE #158: Implement Geomorphing between LODs
As the LODs move across the terrain there's a slight shift. Mike Savage said this could be resolved with geomorphing, which will create a smooth transition from a high resolution mesh to a low resolution one.

Links

MIT white paper
https://dspace.mit.edu/handle/1721.1/45990

https://archive.gamedev.net/archive/columns/hardcore/geomorph/index.html
https://www.gamedev.net/articles/programming/graphics/terrain-geomorphing-in-the-vertex-shader-r1936/

https://www.gamedev.net/forums/topic/219815-geomorphing/

### LINKED ISSUE #209: Last row of mesh backwards
Investigate if this is a problem. The odd row is not always on the bottom. Even in this view as the camera moves around the odd row flipped up to the top for a few frames. 

image

It certainly doesn't line up with collision, but neither does LOD2. The user will never see that as the camera is always far away from the edge. 

image

This may be exacerbating LOD transitions highlighted in #158 #191
Related #421

### LINKED ISSUE #383: Subtle artifact on autoshader on LOD > 0 down -Z or -X
### Terrain3D version

0.9.1

### System information

win10 vulkan 4.2.2

Yes

### Issue description

Godot_v4 2 2-stable_win64_MUHUTVES8R
Godot_v4 2 2-stable_win64_xvRcYd93wl
 very subtle and only in the distance. kind of low priority but noting it here since it is technically an error.

### Logs

_No response_

### LINKED ISSUE #553: Occluded terrain continues to render strips of itself
### Terrain3D version

v0.9.3a

### System information

Godot v4.4.dev3.mono - macOS 15.1.0 - Multi-window, 1 monitor - Metal (Forward+) - integrated Apple M1 Pro (Apple7) - Apple M1 Pro (10 threads)

Yes

### Issue description

Strips of the terrain are still rendered when they should be occluded.

Screenshot 2024-11-15 at 03 14 54

1. Open demo
2. Enable occlusion culling (`Project` > `Project Settings...` > `Occlusion Culling` > Toggle `Use Occlusion Culling`
3. Add `OccluderInstance3D` to the root node of `Demo.tscn`
4. Select the new node and in the `3D` view press `Bake Occluders`
5. Run the scene and venture to the tunnel
6. Use F10 to switch to wireframe mode and observe landscape through the tunnel walls

### Logs

_No response_

Xtarsia:
closes #158
closes #209
closes #383
partially addresses #421 
closes #553

Clipmap method re-written. now exclusivly built from rectangular meshes, with a minimum 2 unit width/length.

renamed geoclipmap to Terrain3DGeoMesh class object, and moved related functions from Terrain3D into this class.

main shader, and extra shaders updated to implement geomorphing.

Geomorph is required as the clipmap does not have any seam meshes.

Minimum mesh size raised to 16 (things break down less than that, even with the old clipmap)

Vertex spacing minimum stepping adjusted to 1/16th (0.0625) prevents a few artifacts with the new methods.

v_camera_pos has been replaced by a uniform _camera_pos, and all relevent shader inserts updated.
Required as the shadow pass would set VIEW_MATRIX[3] to the light source position, resulting in a missmatch between the vertex positions between shadow/standard render passes

The entire mesh is now an alternating grid. ~~Still to update collision to match (should hopefully be fairly straight forwards)~~
Edit: Collision grid now matches.

Current issues:
~~painting tools dont work untill swapping away / back to the scene on first load.~~
~~pretty sure ive not correctly implemented multiple things..~~
~~some crashes when closing / reloading.~~

Im not very familiar with c++ in general and have stumbled my way through this so far.. its "working" but Im a bit in the dark in regards to polishing this up.. an early review would be greatly appreciated!

https://github.com/user-attachments/assets/ae2056a3-d8c9-413d-ad50-ab765cc42910


TokisanGames:
Does this resolve #191?

Xtarsia:
> Does this resolve #191?

It doesnt implement a veiwport size based adjustment to mesh size, but it does very much reduce how noticable transitions are between LODs since they smoothly interpolate now.

TokisanGames:
* Wow, geomorphing looks great. Very nice work.

* The symmetric mesh looks great and is so much better at handling cliff areas at certain angles.

* The circular lods makes the lowpoly shaders look much better without the lod artifacts.

* Collision generation is quite slow.
 
Switching to Full / Editor is 10x slower. I thought it was going to crash at first. It's an extraordinary amount of time for only 3 regions in the demo.

Dynamic Editor at max size/radius is also quite slow to generate. The engine halts for a half second every update. Strangely CPU/GPU time don't report it, even with view gizmos turned off.

Actually, even at normal size, dynamic editor is noticeably sluggish. Turn off gizmos. Then compare with Dynamic Game and move the camera top down, to the side. FPS is high, but the display noticeably stutters.

No console messages. Haven't looked at the code yet.


* Sculpting height reports this error: `Terrain3DInstancer#1310:update_transforms: Instancer isn't initialized.` And instances don't work either.

Xtarsia:
Sculpting height is fixed for me after [1dc804f](https://github.com/TokisanGames/Terrain3D/pull/622/commits/1dc804ff3c0d6828ed91d3f91957bbb8847a10a1)

It seems the generate face loop in _get_shape_data takes an entire milisecond 😬

Xtarsia:
After some testing, and minor shuffling about, I can shave off a bit of time for a 65x65 size shapes from about 1.5ms to 1.2ms per _get_shape_data()

the generate faces loop creates the faces needed for a trimesh (concave) collision shape, so is required as far as I can see.

Testing on main, for a 65x65 shape size, it takes around 0.6ms.

So whilst this is slower, its not the main culprit. maybe set_shape_data() for concave collision shapes is just terribly slow?

Tho I am not sure why it would be so much slower than heightmapshape.

Xtarsia:
Godot does a whole load of processing to the faces data, that is handled entirley different to the heightmapshape methods.

It may be better to write a PR for Godot to enable alternating quad diagonals as an option directly in the heightmap shape.

TokisanGames:
Yes, but that could take 6 months. We should copy the heightmapshape class and make it our own SymmetricHeightMapShape3D and include it here in this PR. That's how I made Terrain3DMaterial. ShaderMaterial wasn't designed to be extended through godot-cpp, so I ripped all the code and built on top of it.

Xtarsia:
height_map_shape_3d.cpp makes calls to the Physics server which is either godot_physics_server_3d.cpp or jolt_physics_server_3d.cpp

Which then pass the data to the respective modules.

I was able to modify https://github.com/godotengine/godot/blob/394508d26dcf1b7a9362453f9009c07d969f1a7e/modules/godot_physics_3d/godot_shape_3d.cpp#L2061C1-L2081C2

That worked for the collision (with GodotPhysics), but I dont think gdextension can get that deep?

TokisanGames:
Which function call is the one causing the slowdown? Is it set_faces? 🤔 I don't see much complexity there though.

> height_map_shape_3d.cpp makes calls to the Physics server 

The only calls I see are to shape_create(), which just memnews an object and returns an RID. Or set_shape_data, which sets data inside the shape. We can do both of those. 

I looked through the code a bit and thought about options like making our own heightmap shape using the custom shape type. Or that GodotHeightMapShap3D is a GodotoncaveShape3D. Perhaps we could do our own processing and setting the data directly in the physics server to avoid setting faces.

Ultimately, however I think any shape we make needs to be a GodotShape3D and implement all of its virtual functions. And we don't have access to it via godot-cpp. So this option is out.

We can start on the Godot PR process, but I think our immediate options are either LOD0 not symmetric and fast or symmetric and slow. It's not just that it's slow, but it makes the engine halt and stutter, which is a big problem. Maybe we can put it in a thread?

Xtarsia:
LOD0 can be not symetric, its the best compromise at this point.

TokisanGames:
Ok. That still allows us to geomorph and have circular lods?
Too bad we didn't get the PR in during the betas. Now we're looking at maybe 4.4.1 and 3.4.1.

Xtarsia:
Collision is fast again now! I can say after looking a lot more into the inner workings, that heightmap shape is handled very differently than concave. I can probably get a PR together for the engine in due time, tho it would mean touching both godotphysics and joltphysics modules, and the scene shapes for the editor debug mesh generation.

Ive actually got godot building so I can chip away at that - and potentially things like dx12 mipmaps 🤔 .

Ultimatley I should be able to boil it down to a check box, and an extra field in the data dict "alternating_grid".

Ready for a good review!

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,456 @@
...
+void Terrain3DGeoMesh::update_clipmap_instances() {
+	if (!_initialized) {
+		return;
+	}
+	RID scenario = _terrain->get_world_3d()->get_scenario();
+	uint32_t render_layers = _terrain->get_render_layers();
+	RenderingServer::ShadowCastingSetting cast_shadows = _terrain->get_cast_shadows();
+	bool baked_light;
+	bool dynamic_gi;
+	bool v = _terrain->is_visible_in_tree();
```
TokisanGames, line 421: Move right before the forloop where it's used. vis or visible is better than v.

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,456 @@
...
+	_clear_mesh_types();
+	_tile_pos_lod_0.clear();
+	_trim_a_pos.clear();
+	_trim_b_pos.clear();
+	_edge_pos.clear();
+	_fill_a_pos.clear();
+	_fill_b_pos.clear();
+}
+
+void Terrain3DGeoMesh::snap(const Vector3 tracked_pos, const real_t mesh_density) {
```
TokisanGames, line 316: Vector3 &

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,456 @@
...
+	//# 7 STANDARD_TILE - mesh_size x mesh_size tiles
+	_mesh_rids.push_back(_generate_mesh(Vector2i(size, size), true));
+	//# 8 STANDARD_EDGE_A - 2 by (mesh_size * 4 + 8) strips to bridge LOD transitions along +-Z axis
+	_mesh_rids.push_back(_generate_mesh(Vector2i(2, size * 4 + 8), true));
+	//# 9 STANDARD_EDGE_B - (mesh_size * 4 + 4) by 2 strips to bridge LOD transitions along +-X axis
+	_mesh_rids.push_back(_generate_mesh(Vector2i(size * 4 + 4, 2), true));
+	return;
+}
+
+RID Terrain3DGeoMesh::_generate_mesh(const Vector2i size, const bool standard_grid) {
```
TokisanGames, line 131: Vector2i &

TokisanGames (changes requested):
Looking good. Thanks.
Does this replace Mike Savage's clipmap code or is it still a close extension of it?
Do the mesh components still move at different speeds?

src/terrain_3d.cpp
```diff
@@ -750,7 +567,9 @@ void Terrain3D::set_vertex_spacing(const real_t p_spacing) {
 void Terrain3D::set_render_layers(const uint32_t p_layers) {
 	LOG(INFO, "Setting terrain render layers to: ", p_layers);
 	_render_layers = p_layers;
-	_update_mesh_instances();
+	if (_initialized && _is_inside_world && is_inside_tree()) {
```
TokisanGames, line 570: Doing this check 3 times seems like it should be done in the update function.
```
if(_geomesh) {
  _geomesh->update()
}
```

The function can verify _terrain->get_world_3d().is_valid() and _terrain->is_inside_tree().
_is_inside_world can go away
get_world_3d().is_valid() can be used in terrain3d::initialize as well.

src/terrain_3d_geomesh.h
```diff
@@ -0,0 +1,81 @@
...
+
+#include <godot_cpp/templates/vector.hpp>
+
+#include "constants.h"
+
+using namespace godot;
+
+class Terrain3D;
+
+class Terrain3DGeoMesh : public Object {
```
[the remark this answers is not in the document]
Xtarsia: "Geo" as in Geomtry, or Earth / Land / Ground / Terrain - "Mesh" was my resoning towards "GeoMesh"

Your call.

src/terrain_3d.cpp
```diff
@@ -711,36 +526,38 @@ void Terrain3D::update_region_labels() {
...
 		_mesh_size = p_size;
 		_initialize();
+		_material->_update_maps();
+		_geomesh->regenerate();
 	}
 }
 
 void Terrain3D::set_vertex_spacing(const real_t p_spacing) {
-	real_t spacing = CLAMP(p_spacing, 0.25f, 100.0f);
+	real_t spacing = round(CLAMP(p_spacing, 0.25f, 100.0f) * 16.f) / 16.f;
```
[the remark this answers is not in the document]
Xtarsia: ive resolved issues that vertex spacing was giving me, so removing this.

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,456 @@
...
+	_generate_clipmap(size, lods, _terrain->get_world_3d()->get_scenario());
+	_initialized = true;
+	update_aabbs();
+	update_clipmap_instances();
+	snap(_terrain->get_snapped_position(), _terrain->get_vertex_spacing());
+}
+
+void Terrain3DGeoMesh::regenerate() {
+	if (!_initialized) {
+		return;
```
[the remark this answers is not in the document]
Xtarsia: I added the initialized bool because some crashes were occuring, however it was just down to my failure to protect _terrain calls.

src/terrain_3d.cpp
```diff
@@ -583,7 +398,7 @@ void Terrain3D::set_debug_level(const int p_level) {
 void Terrain3D::set_data_directory(String p_dir) {
 	LOG(INFO, "Setting data directory to ", p_dir);
 	if (_data_directory != p_dir) {
-		_clear_meshes();
+		//_clear_meshes();
```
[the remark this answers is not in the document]
Xtarsia: Ah I left them commented incase other things broke due to expecting that function to do something, that now wasnt happening.

Turns out it was what was setting _initialized = false; So ive replaced with that now so that setting some settings, properly updates other things (eg, changing vertex spacing, updates colision correctly as it did before)

Xtarsia:
> Looking good. Thanks. Does this replace Mike Savage's clipmap code or is it still a close extension of it? Do the mesh components still move at different speeds?

Its a replacement, It ended up closer to the Hoppes layout, tho instead of 1 L shape mesh, I have 4 long 2 unit wide strips that move in their axis pairs to 2 of 3 positions for their respective axis, enabling each LOD level to move at its own speed.

``ALBEDO = fract(MODEL_MATRIX[3].xyz * 0.1);`` will reveal each segment, tho the colors wont be stable as the camera moves about.

image


the extra pad between LODs was needed due to a minimum 2 unit size for any given mesh, required for the gemorph vertex shader shift. So whilst slightly more complicated, it removed the need for the degenerate seam meshes. I also used pre-calculated offsets (which Mike mentioned in his blog, but never implemented), which simplifies the snap logic a fair bit.

src/terrain_3d.cpp
```diff
@@ -54,6 +53,10 @@ void Terrain3D::_initialize() {
 		LOG(DEBUG, "Creating instancer");
 		_instancer = memnew(Terrain3DInstancer);
 	}
+	if (_geomesh == nullptr) {
```
TokisanGames, line 56: This can now be `!_geomesh`

As of [C++11](https://cplusplus.github.io/CWG/issues/1781.html) nullptr can be directly contextually converted to bool as false and isn't implicit or non-standard.

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,432 @@
...
+}
+
+///////////////////////////
+// Public Functions
+///////////////////////////
+
+void Terrain3DGeoMesh::initialize(Terrain3D *p_terrain) {
+	if (p_terrain) {
+		_terrain = p_terrain;
+	}
```
TokisanGames, line 276: Needs `else { return; }` or those _terrain-> calls will have a problem.

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,434 @@
...
+		AABB aabb = RS->mesh_get_custom_aabb(mesh);
+		aabb.position.y = height_range.x - cull_margin;
+		aabb.size.y = height_range.y + cull_margin * 2.f;
+		RS->mesh_set_custom_aabb(mesh, aabb);
+	}
+	return;
+}
+
+void Terrain3DGeoMesh::_bind_methods() {
+}
```
TokisanGames, line 434: Do we need to expose anything? Otherwise, there's no reason for us to register this class, or have it show up in the docs.

Is there a reason for the user to use update, update_aabbs, snap? Maybe.

src/terrain_3d_geomesh.h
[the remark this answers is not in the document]
TokisanGames: I thought more about it. Let's go with Terrain3DMesher. Thanks.

src/terrain_3d_geomesh.h
```diff
@@ -0,0 +1,78 @@
...
+	GDCLASS(Terrain3DGeoMesh, Object);
+	CLASS_NAME();
+	friend Terrain3D;
```
TokisanGames, line 17: I don't think they need to be friends. I don't see Terrain3D accessing any private functions.

src/terrain_3d_geomesh.cpp
```diff
@@ -0,0 +1,434 @@
...
+	_clear_mesh_types();
+	_tile_pos_lod_0.clear();
+	_trim_a_pos.clear();
+	_trim_b_pos.clear();
+	_edge_pos.clear();
+	_fill_a_pos.clear();
+	_fill_b_pos.clear();
+}
+
+void Terrain3DGeoMesh::snap(const Vector3 &p_tracked_pos, const real_t p_mesh_density) {
```
TokisanGames, line 300: I think you can remove p_mesh_density and just have it look it up from terrain, also checking IS_INIT. Calling this function, especially as a user makes sense with the vector, but not vertex spacing.

TokisanGames:
* Please sort the functions in the CPP and the H file so they are the same order.

src/terrain_3d_geomesh.h
```diff
@@ -0,0 +1,78 @@
...
+private:
+	Terrain3D *_terrain = nullptr;
+
+	RID _generate_mesh(const Vector2i &p_size, const bool p_standard_grid = false);
+	RID _instantiate_mesh(const PackedVector3Array &p_vertices, const PackedInt32Array &p_indices, const AABB &p_aabb);
+	void _generate_offset_data(const int p_mesh_size);
+	void _generate_mesh_types(const int p_mesh_size);
+	void _clear_mesh_types();
+	void _generate_clipmap(const int p_size, const int p_lods, const RID &scenario);
+	void _clear_clipmap();
```
TokisanGames, line 28: Please sort the functions in the CPP and the H file so they are the same order.

src/terrain_3d_geomesh.h
```diff
@@ -0,0 +1,78 @@
...
+	~Terrain3DGeoMesh() { destroy(); }
+
+	void initialize(Terrain3D *p_terrain);
+	void destroy();
+
+	void snap(const Vector3 &p_tracked_pos, const real_t p_mesh_density);
+	void update();
+	void update_aabbs();
+
+	enum MeshType {
```
TokisanGames, line 60: This can go in a `public: // Constants` section above `private:`

src/terrain_3d.cpp
```diff
@@ -118,9 +121,16 @@ void Terrain3D::__physics_process(const double p_delta) {
 	if (is_instance_valid(_camera_instance_id) && _camera->is_inside_tree()) {
 		Vector3 cam_pos = _camera->get_global_position();
 		Vector2 cam_pos_2d = Vector2(cam_pos.x, cam_pos.z);
+		RS->material_set_param(_material->get_material_rid(), "_camera_pos", cam_pos);
```
TokisanGames, line 124: If Terrain3D is going to insert the camera pos into the shader, why does it call update_maps() to insert mesh_size? Do we need to update the maps for the mesh size?

src/terrain_3d.cpp
```diff
@@ -1426,8 +1116,8 @@ void Terrain3D::_bind_methods() {
 
 	ADD_GROUP("Mesh", "");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "mesh_lods", PROPERTY_HINT_RANGE, "1,10,1"), "set_mesh_lods", "get_mesh_lods");
-	ADD_PROPERTY(PropertyInfo(Variant::INT, "mesh_size", PROPERTY_HINT_RANGE, "8,64,1"), "set_mesh_size", "get_mesh_size");
-	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "vertex_spacing", PROPERTY_HINT_RANGE, "0.25,10.0,0.05,or_greater"), "set_vertex_spacing", "get_vertex_spacing");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "mesh_size", PROPERTY_HINT_RANGE, "16,64,2"), "set_mesh_size", "get_mesh_size");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "vertex_spacing", PROPERTY_HINT_RANGE, "0.25,10.0,0.01,or_greater"), "set_vertex_spacing", "get_vertex_spacing");
```
TokisanGames, line 1120: Why did these two change? 

For vertex_spacing, as it is here, it doesn't show more fine grained than 0.05 in the inspector, but this might call it 5x as much when sliding in the inspector.

TokisanGames: I didn't see an issue, but Grok seems to think _mesh_size should be limited to a power of 2.

<details>
<summary>
How would _mesh_size or LOD levels need to change to cause test_x or test_z to be out of range or misplace edge strips?
</summary>
In Terrain3DGeoMesh::snap(), the variables test_x and test_z are calculated as:

```cpp
int test_x = CLAMP(int(round((pos.x - next_x) / snap_step)) + 1, 0, 2);
int test_z = CLAMP(int(round((pos.z - next_z) / snap_step)) + 1, 0, 2);
```
These determine edge strip placements by comparing the current LOD’s snap position (pos) to the next LOD’s snap position (next_x, next_z) relative to snap_step. The CLAMP ensures values stay between 0 and 2, but changes to _mesh_size or LOD levels could still lead to miscalculations or misplacements:

Non-Power-of-Two _mesh_size:

The clipmap system assumes _mesh_size (e.g., 16, 32, 64) is a power of two for even division into snap steps. If _mesh_size = 24 (not a power of two), snap_step might not align with the mesh grid, causing (pos.x - next_x) / snap_step to produce values that, before clamping, are outside 0 to 2. After clamping, this could place edge strips incorrectly, leading to gaps or overlaps.

Very Small _mesh_size (e.g., 2 or 4):

A tiny mesh size reduces the number of tiles, potentially making snap_step too large relative to the mesh. This could cause test_x and test_z to frequently clamp to 0 or 2, misplacing edge strips and leaving parts of the terrain uncovered or misaligned with higher LODs.

Very Large _mesh_size (e.g., 1024):

An excessively large mesh might introduce floating-point precision errors in pos.x - next_x, though this is less common. More critically, if snap_step doesn’t scale proportionally, edge strips might not span the intended areas, causing visual discontinuities.

Increased LOD Levels:

Adding more LOD levels beyond the current design (e.g., exceeding precomputed offsets in _generate_offset_data()) could misalign next_x and next_z. If the clipmap layout or offset data doesn’t account for these extra levels, test_x and test_z might clamp correctly but point to incorrect edge strip positions for higher LODs, disrupting the terrain’s continuity.

Clipmap Layout Changes:

If the tile, trim, or edge arrangement changes (e.g., different spacing or additional mesh types), the assumptions in snap() might not hold. For example, a denser clipmap could shift next_x unexpectedly, pushing test_x outside the expected range before clamping, resulting in misplaced strips.
Example: With _mesh_size = 16 and snap_step = 32 * _vertex_spacing, if _mesh_size changes to 20, the uneven division might make (pos.x - next_x) / snap_step yield a value like -0.8 or 2.3 before the +1 and clamp, leading to clamped values that don’t match the intended edge strip positions. This could create gaps or overlaps in the terrain.

Mitigation: Keep _mesh_size as a power of two and dynamically adjust offset data based on _mesh_size and LOD levels to ensure accurate snap calculations.</details>

src/terrain_3d.cpp
Xtarsia: update_maps() already handled vertex_spacing, so it seemed suitable to add mesh_size update there as well.

The shader needs to know the mesh size for the geomorph blending (vertices are de-scaled, and then the vertex distance from the camera determine the blending value, with mesh_size being a mid-point. So that closer is not shifted at all, and further way is shifted fully.) world_space is used to determine odd/even for each vertices shift direction only.

If the camera travel speed is far in excess of the snapping, then updating the position at true FPS would break the blend, and lead to seems showing up. so it should actually happen at the same rate as the actual vertices move.

there is no reason using the camera matrix value as for before for dual scaling couldn't be the case.

src/terrain_3d.cpp
Xtarsia: I increased mesh size limit as less than 16 was breaking things, even with the original clipmap, gaps between LODs would appear.

vertex spacing being more granular than before was a result of me making sure the clipmap movement was actually robust, I noticed before some tiny gaps when some values were set, fixed now. (missing rounds, either in snap(), or in the shader)

It can be increased back to 0.05, tho for users with imported real world data, maybe a more specific scaling would be appreciated?

As for Grok, precision errors would accumulate without the round, where some values might be 0.9999 instead of 1.0, and then set the wrong index offset as the int cast would floor that to 0. but round() is there to catch that. 

test_x and test_z determine the current LOD position reletive to the next LOD on that axis, and return either "nearside-edge, in the middle, or offside-edge" any float precision error would need to be extreme to cause problems, that 0.9999 would have to be 0.4999 to cause an incorrect positioning.

src/terrain_3d_geomesh.cpp
Xtarsia: Agree.

src/terrain_3d.cpp
Xtarsia: vertex spacing throws all this right out the window, non po2 spacing values, combined with non po2 mesh size can cause firefly artifacts even with the old clipmap past 100km away from origin.

Ive set spacing back to 0.05 increments again.

src/terrain_3d_geomesh.cpp
Xtarsia: no longer registered.

TokisanGames:
Please let me know on discord when you're ready for another review, or if you have questions. Otherwise, I'm ignoring notifications on this one until then.

TokisanGames:
Even after a rebase to pick up the new compat change, the minimal shaders don't work in compatibility mode. They report fma() isn't supported, and in the dump the shader defines aren't present, though I see them in the shader editor.

src/terrain_3d.cpp
```diff
@@ -184,209 +194,20 @@ void Terrain3D::_destroy_collision(const bool p_final) {
...
-
-	for (const RID rid : _mesh_data.seams) {
-		RS->instance_set_visible(rid, v);
-		RS->instance_set_scenario(rid, _scenario);
-		RS->instance_set_layer_mask(rid, _render_layers);
-		RS->instance_geometry_set_cast_shadows_setting(rid, _cast_shadows);
-		RS->instance_geometry_set_flag(rid, RenderingServer::INSTANCE_FLAG_USE_BAKED_LIGHT, baked_light);
-		RS->instance_geometry_set_flag(rid, RenderingServer::INSTANCE_FLAG_USE_DYNAMIC_GI, dynamic_gi);
+	if (p_final) {
+		delete _mesher;
```
TokisanGames, line 203: This whole function can be:
```
    if (_mesher) {
        _mesher->destroy();
        if (p_final) {
            delete _mesher;
            _mesher = nullptr;
        }
    }
```

src/terrain_3d_mesher.cpp
```diff
@@ -0,0 +1,453 @@
...
+	if (p_terrain) {
+		_terrain = p_terrain;
+	} else {
+		return;
+	}
+	IS_DATA_INIT_MESG("Terrain3D not initialized yet", VOID);
+	LOG(INFO, "Initializing GeoMesh");
+	int size = _terrain->get_mesh_size();
+	int lods = _terrain->get_mesh_lods();
+	_generate_clipmap(size, lods, _terrain->get_world_3d()->get_scenario());
```
TokisanGames, line 294: Like below in update, you should 	`if (_terrain->get_world_3d().is_null()) {`
I had a crash because of this. But then after doing some things, couldn't reproduce it.

src/terrain_3d.cpp
```diff
@@ -72,9 +75,9 @@ void Terrain3D::_initialize() {
 		_data->connect("maps_changed", callable_mp(_material.ptr(), &Terrain3DMaterial::_update_maps));
 	}
 	// Height map was regenerated, update aabbs
-	if (!_data->is_connected("height_maps_changed", callable_mp(this, &Terrain3D::update_aabbs))) {
+	if (!_data->is_connected("height_maps_changed", callable_mp(this, &Terrain3D::_update_mesher_aabbs))) {
 		LOG(DEBUG, "Connecting _data::height_maps_changed signal to update_aabbs()");
-		_data->connect("height_maps_changed", callable_mp(this, &Terrain3D::update_aabbs));
+		_data->connect("height_maps_changed", callable_mp(this, &Terrain3D::_update_mesher_aabbs));
```
TokisanGames, line 80: These can connect directly to the public `_mesher, &Terrain3Dmesher::update_aabbs`, and `_update_mesher_aabbs` can go away.

Xtarsia: That works if Terrain3DMesher is registered as a Godot class.

as I see it, the options are:

1. Re-register the class, remove the alias.
2. Keep the class native, and keep the current alias.

unless there is a 3rd option?

TokisanGames: Ok, I put it inline in the header.

Issue #351, opened 2024-04-15
Improve Height blending options


TokisanGames:
### Issues
* Provide artists with more knobs for adjusting height blending 
* Blend more interestingly when not using height textures

### How it currently works

The terrain editor is a vertex painter, not a pixel painter. The shader bilinearly blends the 4 surrounding vertices. Linear looks like a gradient. Bilinear looks like a blended square, as shown below when there are no height textures:
 
![image](https://github.com/TokisanGames/Terrain3D/assets/632766/a9aad89a-111a-48b7-bade-1de0f268a55c)

What varies the blend is either the height texture for use in height blending, or the application of the noise texture to the weighting. If height blending is disabled, or there is no height texture, the blend is a straight bilinear blend, aka gradient aka alpha blend.

### Suggestions for blending improvements

-------
### Scale height during weighting
This scales down heights to allow for greater noise in clamping the weights by @Xtarsia :

even with height map packed, it can still not be great depending on the heightmap, uv scale etc. The shader can be tweaked a little, to give a lot more weight to the noise3 value when blending.

currently the default blend weighting is setup like this:
```
    weights.x = blend_weights(weights0.x * weights0.y, clamp(mat[0].alb_ht.a + noise3, 0., 1.));
    weights.y = blend_weights(weights0.x * weights1.y, clamp(mat[1].alb_ht.a + noise3, 0., 1.));
    weights.z = blend_weights(weights1.x * weights0.y, clamp(mat[2].alb_ht.a + noise3, 0., 1.));
    weights.w = blend_weights(weights1.x * weights1.y, clamp(mat[3].alb_ht.a + noise3, 0., 1.));
```

but if you have no height data, this yields a value of 1 in alpha giving noise no room at all. 

however even with height data, due to being limited to 1m blend square, unless the height map is highly variant, and high contrast the "square" blending is going to show through in a basetexture <> basetexture case.

a very quick change is to modify the shader at this section:
```
    float height_bias = 0.5;
    weights.x = blend_weights(weights0.x * weights0.y, clamp(mat[0].alb_ht.a*height_bias + noise3, 0., 1.));
    weights.y = blend_weights(weights0.x * weights1.y, clamp(mat[1].alb_ht.a*height_bias + noise3, 0., 1.));
    weights.z = blend_weights(weights1.x * weights0.y, clamp(mat[2].alb_ht.a*height_bias + noise3, 0., 1.));
[...]

PR #626, opened 2025-02-20
implement normal depth and align projection with up axis

### LINKED ISSUE #171: Add Per Channel AO (height), Normal strength, Roughness mod
We can use height as AO for free. We can use per channel AO and strength. I'm currently experimenting with it on our many textures and so far it's something we'll definitely be using on most of our textures.

I'm also finding the need for per channel normal strength, and roughness mod.

### LINKED ISSUE #318: Misleading naming in shader

Xtarsia:
update:

- Implements normalmap depth property for texture assets.
- Implements AO estimation from height and normals.
- Implements Roughness modifier (+- 1.0 linear shift) per texture.
- closes #171 
****
- Implements world space texture normal blend mode.
- partially addressing #351 
****
- Implements additional detiling lateral shift, along side rotation (useful for textures with a clear orientation)
- Added a lightweight shader to extras
- Aligns projection with y axis, as shown below.
- Fixes height blending of normals, and normalizes during re-packing.
****  
- Renamed index lookup functions
- closes #318 
****
Todo:

- [x] update all extra shaders with new function naming
- [x] Implement roughness mod per texture asset to fully close 171

****

image
``ALBEDO = vec3(normal_map_depth);``
Setting 1 texture to 0.0, and the left as the default 1.0
image


Xtarsia:
world space normal blending, takes the base (default) normal map, and reduces the overlay height value where the base layer would have a high slant.

This means grass grows on top, not sides etc.

Before:
image

After:
image

this is per-pixel not per-index so gives quite detailed results.

blend values of 0 or 1 still result in pute base or overlay.

Xtarsia:
Implemented per texture AO strength, with a derivative based approimation in the shader (almost free, no textures / extra lookups required!)

no AO:
image
image


with apprimated AO:
image
image

src/shaders/main.glsl
```diff
@@ -168,19 +171,13 @@ void vertex() {
...
 		ivec3 uv_b = get_region_uv(end_pos, VERTEX_PASS);
 		float h = mix(texelFetch(_height_maps, uv_a, 0).r,texelFetch(_height_maps, uv_b, 0).r,vertex_lerp);
-		float u = texelFetch(_height_maps, get_region_uv(UV + vec2(1,0), VERTEX_PASS), 0).r;
-		float v = texelFetch(_height_maps, get_region_uv(UV + vec2(0,1), VERTEX_PASS), 0).r;
+		//float u = texelFetch(_height_maps, get_region_uv(UV + vec2(1,0), VERTEX_PASS), 0).r;
+		//float v = texelFetch(_height_maps, get_region_uv(UV + vec2(0,1), VERTEX_PASS), 0).r;
 //INSERT: WORLD_NOISE2
 		v_vertex.y = h;
-		v_normal = vec3(h - u, _vertex_spacing, h - v);
+		//v_normal = vec3(h - u, _vertex_spacing, h - v);
```
TokisanGames, line 178: Why are all of these commented out?

src/shaders/main.glsl
```diff
@@ -249,17 +263,39 @@ vec2 rotate_plane(vec2 plane, float angle) {
...
+	
+	if (region < 0 || i_normal.y >= projection_threshold || !enable_projection) {
+		base_uv = v_vertex.xz;
+	} else { // Project UVs and determine surface normal angle
+		// Quantize the normal otherwise textures lose continuity across domains
+		// Avoid potential singularitys
+		#define SQRT3 1.73205080757
+		vec3 p_normal = normalize(round(i_normal * SQRT3 * projection_angular_division));
+		vec3 p_tangent = normalize(cross(p_normal, vec3(1e-6, 1.0, 1e-6)));
+	    p_angle = atan(-i_normal.x, -i_normal.z);
```
TokisanGames, line 286: This has spaces in front of it, while all the other lines have tabs

src/shaders/world_noise.glsl
```diff
@@ -116,8 +116,8 @@ float get_noise_height(const vec2 uv) {
 				get_noise_height(nuv_b + vec2(0.0, _region_texel_size)),vertex_lerp);
 			world_noise_ddxy = vec2(nh - nu, nh - nv);
 			h += nh;
-			u += nu;
-			v += nv;
+			//u += nu;
+			//v += nv;
```
TokisanGames, line 120: Why are these commented?

src/terrain_3d_assets.h
```diff
@@ -34,8 +34,10 @@ class Terrain3DAssets : public Resource {
 	GeneratedTexture _generated_albedo_textures;
 	GeneratedTexture _generated_normal_textures;
 	PackedColorArray _texture_colors;
+	PackedFloat32Array _texture_normal_depths;
```
TokisanGames, line 37: Normal depth, not normal strength as godot calls it?
Is there a difference?
Does this calculation use non-unit normals?

TokisanGames:
I like the new blending, ao and normal strength a lot.

After seeing the horizontal rock pattern, I think it is vastly superior to diagonal textures like the current one.

How did you achieve the first picture? I put in the same texture, but none of the projection settings get the horizontal movement as consistent as in your picture. Mine looks chopped up. Yours looks clean and consistent. Also, on mine the vertical lines instead of horiztonal lines outside of regions in the background is odd. I don't see that on yours.



{633CCDE8-FEEE-41A5-BEE9-988CE1DCF465}

src/shaders/main.glsl
Xtarsia: I did that as they are un-used, tho vertex normal may get used for displacement later one.

Can remove entirely for now.

src/shaders/main.glsl
Xtarsia: Good spot!

src/shaders/world_noise.glsl
Xtarsia: same as the vertex normal, can remove entirley.

src/terrain_3d_assets.h
Xtarsia: Ah.. I had depth in my mind as thats how I see it behave. Can rename to match godot naming.

Xtarsia:
That area is a lot more varied in surface angle than where my screenshot was, so each angle will break things up, dropping projection divisions to 1.0 will make things more consistent, at the cost of less accuracy to actual surface angle.

The background has projection enabled, tho it shows unblended transitions, its simple to enable it for background as default, its probably fine as its far away.

src/terrain_3d_assets.h
Xtarsia: Oh, the actual shader built in is NORMAL_MAP_DEPTH too.

TokisanGames: Actually on the material in 4.3 it's called normal_scale. Maybe strength was 3.x. But if glsl is depth, that's fine.

TokisanGames:
Thanks for enabling projection on the world noise. It looks more consistent with the other. However the blend between projection seems to be missing.

angle 2.0 in regions

{5384D3C6-72FC-47D1-8B06-512B176F5ADF}

Has hard lines Outside of regions

{B7C4691C-D0F6-4248-A2EC-FCDDCB51F0F1}

Xtarsia:
Ah, to fix the blend outside of regions, world noise would need calculating in fragment 8x.

Can add it to the fragment normals option.

TokisanGames:
Maybe it's not worth having if it's going to sap so much performance. I don't want to give users bad options they shouldn't use. WorldNoise is already borderline where my official stance is don't use it for games, it's just for demos. On 1.0 Angular it's not really noticable unless you're looking for it, so it's fine.

TokisanGames:
* Looks amazing. What you've done with 3D projection is magical. People are going to love it. My team is very excited to use layered rock textures like this.
* The normals and AO look incredible.
* The roughness is already useful.
* The 3D projection greatly improves the previous rock texture, but I don't like it anymore. So many rock textures have diagonal lines. I didn't realize it was a problem before this, but now I see it's no good.
* The code looks good.
* Mobile and forward work fine on all shaders. 

Issues
* Compatibility Only default & lightweight work. Minimum, low_poly*: shader error - fma. 
* AO bleeds into the debug shaders**, eg. checkerboard. Please reset AO on all but nav and the grids.

TokisanGames:
Great, thank you!

PR #629, opened 2025-02-26
Strip comments before editor shader injection

Xtarsia:
Admin update:
* Strips comments from shader before injection to fix commented braces
* Verifies shader has been compiled before decal use in UI.gd
* Reduce shader recompilation on tool changes; only navigation now
* Implement region decal, partially working on #100 
* Fix default texture card offset

-----
Fixes issues with comments being incorrectly picked up as places to inject code.

Expand decal to include region tool

### DIFF
--- a/project/addons/terrain_3d/src/ui.gd
+++ b/project/addons/terrain_3d/src/ui.gd
@@ -33,7 +33,12 @@ const OP_NEGATIVE_ONLY: int = 0x02
 
 const RING1: String = "res://addons/terrain_3d/brushes/ring1.exr"
 @onready var ring_texture := ImageTexture.create_from_image(Terrain3DUtil.black_to_alpha(Image.load_from_file(RING1)))
-
+@onready var region_texture := ImageTexture.new() :
+	set(value):
+		var image: Image = Image.create_empty(1, 1, false, Image.FORMAT_R8)
+		image.fill(Color.WHITE)
+		value.create_from_image(image)
+		region_texture = value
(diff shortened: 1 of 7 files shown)


TokisanGames:
Please review these changes. Should be a faithful rendition of:

https://github.com/godotengine/godot/blob/master/servers/rendering/shader_preprocessor.cpp#L235-L312
and
https://github.com/godotengine/godot/blob/master/servers/rendering/shader_preprocessor.cpp#L356-L367

The functions were verbatim copied with these changes:
* Replaced `LocalVector` with `Vector`
* Replaced `.size()` with `.length()` (which the former includes the zero string termination)
* Converted the supporting functions into embedded lambdas.

Xtarsia:
Yes, this now covers all valid cases. Its was still possible to induce spam with a purposefully broken override shader, as the ui.gd would attempt to set uniforms that were not present.

Checking if if the shader is valid before setting anything via RenderingServer avoids un-needed spam as well.

I noticed that swapping tools causes the shader to re-compile every single time, tho I guess that is due to navigaion debug view?

TokisanGames:
Yes, I've noticed the rebuilding for a while now and find it annoying. It fills up the logs. Even pressing shift regenerates it. Terrain3DEditor::set_tool causes the shader to recompile and maps to update. We could limit that to the navigation tool, if it's even still needed.

TokisanGames:
You mean have the nav code in the editor function shader always? That sounds fine. I'd be happy to drop the rebuild for all tool changes. You want to add it here?

TokisanGames:
Great, thanks for your help with these updates.

PR #631, opened 2025-03-02
fix region decal errors

Xtarsia:
Prevents this error:
image

the region at map index 1022, in none BG mode is dark only when temporarily showing a region like this:
image

Normally the add region tool looks like this:
image

If there is a clean way to get an editor insert to modify the holes code this can be avoided, but for now this is tollerable.


Xtarsia:
Everything should be fixed now, including the single region being weird for none BG mode.

TokisanGames:
Excellent, thank you!

Issue #630, opened 2025-02-28
Meshes cause memory leaks on engine exit


pewcworrell:
### Terrain3D version

v0.9.3a

### System information

Godot v4.4.rc1 - Windows 11 (build 22631) - Multi-window, 3 monitors - OpenGL 3 (Compatibility) - AMD Radeon RX 7900 XT (Advanced Micro Devices, Inc.; 32.0.11037.4004) - AMD Ryzen Threadripper 7970X 32-Cores (64 threads)

### Is the issue reproducable in the demo?

Yes

### Issue description

It appears as though the `_packed_scene` and/or the `_meshes` within `terrain_3d_mesh_asset.cpp` are not cleaned up properly on engine exit. I've uploaded an MRP here:
[terrain_leak.zip](https://github.com/user-attachments/files/19033996/terrain_leak.zip)
To see the leak:
- Download and extract the project
- Copy in the terrain addons folder into the project
- Either set your GODOT_EXEC environment variable, or modify the `builds/BuildGame.bat` file to hardcode the path
- Run `builds\BuildGame.bat -windows`
- See leaks in the output (I've left my full output log in the project if you want to compare)

This doesn't really impair the usage of the this plugin, but the leaks in the output can mask leaks made elsewhere in users projects. Clearing the "Scene File" makes these leaks go away.

### Logs

```
ERROR: 1 RID allocations of type 'N16RendererViewport8ViewportE' were leaked at exit.
ERROR: Pages in use exist at exit in PagedAllocator: N20RasterizerSceneDummy21GeometryInstanceDummyE
   at: ~PagedAllocator (./core/templates/paged_allocator.h:170)
ERROR: 1 RID allocations of type 'N13RendererDummy9DummyMeshE' were leaked at exit.
WARNING: Leaked instance dependency: Bug - did not call instance_notify_deleted when freeing.
   at: ~Dependency (servers/rendering/storage/utilities.cpp:56)
ERROR: 1 RID allocations of type 'N13RendererDummy15MaterialStorage13DummyMaterialE' were leaked at exit.
ERROR: 4 RID allocations of type 'N17RendererSceneCull8InstanceE' were leaked at exit.
ERROR: 1 RID allocations of type 'N17RendererSceneCull8ScenarioE' were leaked at exit.
ERROR: 1 RID allocations of type 'N17RendererSceneCull6CameraE' were leaked at exit.
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
[...]

PR #632, opened 2025-03-04
Free instantiated node in set_scene_file after done processing meshes

pew-jfrohlander:
Fix for issue #630


TokisanGames:
Thanks!

PR #637, opened 2025-03-06
Improve region decal

Xtarsia:
@TokisanGames :
> It shouldn't light up on existing regions unless it's going to remove them. 
> Also in NONE mode, when the mouse is out of the viewport, the decal fades away but the land stays there

Both issues resolved by these changes.


TokisanGames:
Excellent, thank you!

PR #644, opened 2025-03-15
Update compatibility code

Xtarsia:
Addresses issue with strip comments removeing the text that was being checked against, causeing duplicate insert.

Also removed unused define overrides, and made fma override more robust.


TokisanGames:
Excellent, thank you.

PR #647, opened 2025-03-17
Free editor textures in game mode

### LINKED ISSUE #566: Free TextureAssets in game, or ensure API exists
### Description

Textures are copied into texture arrays. Provide an option to free them in game (enabled by default), and an API (probably in Terrain3DAssets) to do so on demand. Some games might have a level editor so will want to keep the original assets.

TokisanGames:
Frees editor textures in game mode, enabled by default

Fixes #566 

@Xtarsia Would you review this please?
You can use the vram debugger, but one texture is used by borders, so delete that node.
You can test the remote scene tree and edit texture settings with this setting enabled and not.


Xtarsia:
Looks good, does what it should.

PR #649, opened 2025-03-18
Fix set_control_overlay_id

k1r4n8:
Fixes the GDScript binding for setting an overlay id. Without it, the base id is being set instead.


TokisanGames:
Great, thank you!

PR #658, opened 2025-03-26
Consolidate force_update*, update_* and expose more options

TokisanGames:
Removes usages of `force`, which we're using improperly. It should be used to bypass warnings put in place to protect against major mistakes like formatting your disk. 

**Texture Arrays**

Consolidates force_update_maps, update_maps into one function and expose all of the options for the user.

By default, update_maps() does the slowest, most complete option, and the parameters reduce what it does.

The second line is the same behavior as the former:
```c++
-data.force_update_maps()
+data.update_maps(Terrain3DRegion.TYPE_MAX, true, false)
```

A user can add this to Player.gd:_physics_process() to paint the terrain it walks over, updating only the region the player is in, and regenerating the mipmap for that region since it's editing the color. Update_maps also doesn't even need to be called every frame or physics frame; it could be run at 10fps and look fine.
```gdscript
    $"../Terrain3D".data.set_color(global_position, Color.DARK_RED)
    var region:Terrain3DRegion = $"../Terrain3D".data.get_regionp(global_position)
    region.set_edited(true)
    $"../Terrain3D".data.update_maps(Terrain3DRegion.TYPE_COLOR, false, true)
    region.set_edited(false)
```

**Collision**
* renames update(force=false) -> update(rebuild=false), no logic change

**Instancer**
* renames force_update_mmis() -> update_mmis(rebuild=true)
* offers a new option to not rebuild everything

@Xtarsia Will you review this please?

Should we always default the updates to the slowest/complete choice and specify parameters to limit, or the opposite, or do it individually for each (current)?

### DIFF
--- a/doc/api/class_terrain3dcollision.rst
+++ b/doc/api/class_terrain3dcollision.rst
@@ -337,7 +337,7 @@ Returns true if :ref:`mode<class_Terrain3DCollision_property_mode>` is not ``Dis
 
 - If :ref:`mode<class_Terrain3DCollision_property_mode>` is Full, updates the existing collision shapes. If regions have been added or removed, set ``force`` to true or call :ref:`build()<class_Terrain3DCollision_method_build>` instead. Can be slow.
 
-- If :ref:`mode<class_Terrain3DCollision_property_mode>` is Dynamic, repositions collision shapes around the camera and recalculated ones not already in place, skipping those that are. Set ``force`` to true to recalculate all shapes. This is very fast, and can be updated at 60fps for little cost.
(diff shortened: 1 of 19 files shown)


TokisanGames:
* I'll change instancer to match collision, not destroying first unless requested.
* I'll set update_maps generate_mipmaps to false by default, since editing color will be more rare.
* And I'll default update_maps all_regions to true, since otherwise the user needs to set the edited flag, and should turn it off after. A function that needs three lines of code to work shouldn't be the default over a single line that works. I'll note this in the docs.
So update_maps() means all maps, all regions, no mipmaps.

PR #659, opened 2025-03-27
4.4 Compatibility updates

Xtarsia:
4.4 comes with some much needed features that removes the need for any of the hoop jumping that was required before.

Tested with full texture arrays on 4.4.1

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -256,7 +256,6 @@ void Terrain3D::_setup_mouse_picking() {
 	Ref<ShaderMaterial> shader_material;
 	shader_material.instantiate();
 	shader_material->set_shader(shader);
-	shader_material->set_shader_parameter("compatibility", _compatibility);
 	_mouse_quad->set_surface_override_material(0, shader_material);
 	_mouse_quad->set_position(Vector3(0.f, 0.f, -0.5f));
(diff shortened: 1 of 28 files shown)


src/terrain_3d.cpp
```diff
@@ -715,7 +706,7 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
 		Color screen_depth = vp_img->get_pixel(0, 0);
 
 		// Get position from depth packed in RGB - unpack back to float.
-		// Forward+ is 16bit, mobile is 10bit and compatibility is 8bit.
+		// Forward+ is 16bit, mobile and compatibility is 10bit.
 		// Compatibility also has precision loss for values below 0.5, so
 		// we use only the top half of the range, for 21bit depth encoded.
```
TokisanGames, line 711: Compatibility now has a 10-bit buffer like mobile? And yet still suffers precision loss below 0.5, while mobile doesn't?

Xtarsia: Compatibility buffer got upgraded to 10 bit, however the issue lies in the srgb conversion / tonemapping.

There is a PR adressing it: https://github.com/godotengine/godot/pull/103141

PR #669, opened 2025-04-06
fixes collision with double precision builds

xanhast:
add 64 bit float support to map data when double precision is used.

tested demo with both precisions


TokisanGames:
Thanks for the PR. I looked back at the old code before separating the collision class. We used to use this, which already has the define included. So let's do this instead:
```
		PackedRealArray map_data = PackedRealArray();
```

Also, please look at #671. Can you reproduce that either before or after your PR?

TokisanGames:
Ok great. Please squash your changes and I can merge it in.

`git rebase -i HEAD~2`

change second commit from pick to fixup

xanhast:
thanks for your patience and instruction

TokisanGames:
Thanks!

PR #675, opened 2025-04-12
feat region streaming

druidui:
add options in inspector under "Regions" to change render distance and to toggle region streaming
add region streaming implementation

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -122,6 +122,7 @@ void Terrain3D::__physics_process(const double p_delta) {
 		Vector3 cam_pos = _camera->get_global_position();
 		Vector2 cam_pos_2d = Vector2(cam_pos.x, cam_pos.z);
 		RS->material_set_param(_material->get_material_rid(), "_camera_pos", cam_pos);
+
 		if (_camera_last_position.distance_to(cam_pos_2d) > 0.2f) {
 			if (_mesher) {
 				_mesher->snap(cam_pos);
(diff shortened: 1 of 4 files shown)


druidui:
Some issues with performance here, only tested at region size 2048 - if there's anything you can spot which may immediately increase performance, please share!

druidui:
- put region loading on it's own thread

TokisanGames (changes requested):
Thanks for a solid start on this. There's some good stuff here. There's data loss and performance issues, but a good start.

* When running it with WorldBackground = none, 1-4 regions that don't exist appear under the camera, many more if rings > 1. They are real regions that generate collision. They can't be removed.

* Disabling streaming, all of the extra regions still exist. And existing regions that were hidden are insivible/unloaded.

* I made 56 regions with data (7x8), and saved. I then enabled region streaming and ran. Several regions had been flattened. I then quit, reverted the scene so streaming was no longer enabled and reloaded the scene. This then loaded all regions from disk that existed when I ran the scene. Several regions were missing entirely. The files had been deleted. Likely set_deleted then saved.

* I looked at the smaller code change blocks. I skipped over the large blocks of code changes for now, and will look closer over time as it improves. Because of the extent of the modifications, I want to either remove unrelated changes, or at the very least put them in a separate commit for now. I also want to discuss the design and intention of some of them either here or on discord.

* I want the function code to be in the same order as in the header. Right now you have private functions mixed in with the public functions.

* What I don't see in here is the lifting of max bounds and letting it generate a smaller window of regions. Which that may be more complicated and better done in a subsequent PR. Why don't you make a tracker in the first post of the various tasks you intend to scope for this PR, marking those you've done and still plan to do.

* Each time you do an update, it would be a good idea to rebase on the latest.

* Performance is a bit choppy. This is a first pass, but we'll need more refinement before it is production ready. We need to do some profiling, probably with an engine with debug symbols as well. Or at the very least, you could print out the time for every section so we know what is actually slow

* I haven't looked closely, but please describe the region states you've added. We had:
  * modified (to be saved)
  * edited (for undo)
  * deleted (to have file removed)
It seems like you haven't changed this and regions are either completely unloaded or loaded and rendering.

Perhaps we also need separate states for:
  * on disk only / unloaded
  * loaded in RAM only
  * visible/invisible (compiled in the texture array in VRAM, but not rendered - vertices discarded)
  * rendering normally

This would give us more fine-grained control over performance. Combined with the performance metrics, we'd limit the most expensive operations, or space out when we do them, and allow us to freely do the cheap ones (eg changing visibility). Just changing visibility will improve performance.

@Xtarsia did you play with this?

PR #680, opened 2025-04-22
Add contour lines shader & Hotkeys

### LINKED ISSUE #605: More shortcuts
### Description

I find myself having to move my mouse away from the viewport a lot to switch between tools and change the brush size. I think that the following common actions would benefit from having shortcuts:
* selecting a tool
* changing the brush size
* changing the brush strength

I imagine the shortcuts to work something like this:
* when the terrain node is selected, each tool can be selected with a single key-press, similar to the QWER shortcuts in the Godot editor.
  * ideally, the keys can be configured
* the brush size/strength can be changed with either
  * clicking and dragging with the mouse while holding down multiple modifier keys (photoshop)
    * fast and accurate
  * the mouse wheel while holding down a modifier key (gimp)
    * slow for large changes and inaccurate for small changes
  * a shortcut like "[", "]", "{" and "}" (gimp)
    * requires taking my hand off the mouse

TokisanGames:
Partially resolves #605 

* Adds customizable contour lines (enable w/ `4` or Overlays, and customize in material)
* Adds hotkeys 
* Update docs

{8E0BB697-0CE6-44DC-A8CF-6DC0EDF93BFD}

Contributions to this PR by @Xtarsia 

Reference:
* https://github.com/TokisanGames/Terrain3D/discussions/358
* https://github.com/FishOfTheNorthStar/Terrain3D_altCustom/tree/master/project/addons/terrain_3d/shader/mods `(carto*)`

### DIFF
--- a/project/addons/terrain_3d/src/editor_plugin.gd
+++ b/project/addons/terrain_3d/src/editor_plugin.gd
@@ -151,11 +151,13 @@ func _clear() -> void:
 	region_gizmo.clear()
 
 
-func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> int:
+func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> AfterGUIInput:
 	if not is_terrain_valid():
 		return AFTER_GUI_INPUT_PASS
 
-	_read_input(p_event)
+	var continue_input: AfterGUIInput = _read_input(p_event)
+	if continue_input != AFTER_GUI_INPUT_CUSTOM:
+		return continue_input
 	ui.update_decal()
 	
 	## Setup active camera & viewport
@@ -249,7 +251,7 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> 
 	return AFTER_GUI_INPUT_PASS
 
 
-func _read_input(p_event: InputEvent = null) -> void:
+func _read_input(p_event: InputEvent = null) -> AfterGUIInput:
 	## Determine if user is moving camera or applying
 	if Input.is_mouse_button_pressed(MOUSE_BUTTON_LEFT) or \
 		p_event is InputEventMouseButton and p_event.is_released() and \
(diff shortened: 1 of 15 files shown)


project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -293,14 +295,21 @@ func _read_input(p_event: InputEvent = null) -> void:
 	# Return if modifiers haven't changed AND brush_data has them;
 	# modifiers disappear from brush_data when clicking asset_dock (Why?)
 	var current_mods: int = int(modifier_shift) | int(modifier_ctrl) << 1 | int(modifier_alt) << 2
-	if _last_modifiers == current_mods and ui.brush_data.has("modifier_shift"):
-		return
 	
-	_last_modifiers = current_mods
```
TokisanGames, line 299: Though I read the comment, I don't recall what this is for. So I don't know if it's necessary anymore.

Xtarsia (commented):
I quite like the seperation of some debug views into the overlays section, good call!

The line color setting was not immediatley apparent, especially since i initially enabled it via the alias.

Hot keys work well too.

Everything else looks good!

project/addons/terrain_3d/src/editor_plugin.gd
TokisanGames: Hmm, there's a godot bug. If holding down ctrl, clicking the detached asset dock and hovering over the main window, which triggers `window.grab_focus()`, `Input.is_key_pressed(KEY_CTRL)` returns false. We cannot get around that.

However, the commented code is referring to brush_data["modifier_*"] disappearing. I don't see that with the detached window. It does occur if the asset_dock is attached and a texture is clicked, or when changing tools. So changing the textures probably sets the paint tool.

It seems we no longer recompile the shader every time we change tools though. That's an improvement.

I will restore an updated version of the code I took out.

TokisanGames: Hmm, there's a godot bug. If holding down ctrl, clicking the detached asset dock and hovering over the main window, which triggers `window.grab_focus()`, `Input.is_key_pressed(KEY_CTRL)` returns false. We cannot get around that.

However, the commented code is referring to brush_data["modifier_*"] disappearing. I don't see that with the detached window. It does occur if the asset_dock is attached and a texture is clicked, or when changing tools. So changing the textures probably sets the paint tool.

It seems we no longer recompile the shader every time we change tools though. That's an improvement.

I will restore an updated version of the code I took out.

TokisanGames:
* Added alpha to color
* Revised modifier code
* Scoped shader views

I'm considering dropping the fixed contour mode. I think dynamic will be adequate.

TokisanGames:
I removed the fixed scale contour lines in another commit, so it is at least saved in the repo in case we want to add it again.

PR #688, opened 2025-05-04
Implements a texture filtering feature in project_on_terrain3d

Furqit:
image
image

### DIFF
--- a/project/addons/terrain_3d/extras/3rd_party/project_on_terrain3d.gd
+++ b/project/addons/terrain_3d/extras/3rd_party/project_on_terrain3d.gd
@@ -15,8 +15,12 @@
 #
 #
 #@export var terrain_node : NodePath
-#@export var align_with_collision_normal := false
-#@export_range(0.0, 90.0, 0.1) var max_slope = 90.0
+#@export var align_with_collision_normal : bool = false
+#@export_range(0.0, 90.0, 0.1) var max_slope : float = 90.0
+#@export var enable_texture_filtering : bool = false
+#@export_range(0, 31) var target_texture_id : int = 0
+#@export var not_target_texture : bool = false
+#@export_range(0.0, 1.0, 0.01) var texture_threshold : float = 0.8
 #
 #var _terrain: Terrain3D
 #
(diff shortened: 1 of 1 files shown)


project/addons/terrain_3d/extras/3rd_party/project_on_terrain3d.gd
```diff
@@ -74,23 +103,43 @@
 		#var height: float = _terrain.data.get_height(location)
```
TokisanGames, line 103: Ensuring that that the location is valid should be done first. 
```
if is_nan(height):
  continue
```

Then checking the normal and confirming the location is within a valid area.
```
if not abs(Vector3....: 
   continue
```

Then checking the texture.
```
if enable_texture_filter:
   if not in texture area:
      continue
   texture
```

Then applying the collision normal if requested.
```
if align_with_collision_normal:
  t.basis
```

Then storing
```
new_transforms.push_back(t)
```

project/addons/terrain_3d/extras/3rd_party/project_on_terrain3d.gd
```diff
@@ -74,23 +103,43 @@
...
 		#var normal: Vector3 = _terrain.data.get_normal(location)
 		#
-		#if align_with_collision_normal and not is_nan(normal.x):
-			#t.basis.y = normal
-			#t.basis.x = -t.basis.z.cross(normal)
-			#t.basis = t.basis.orthonormalized()
-#
-		#if abs(Vector3.UP.dot(normal)) >= (1.0 - remapped_max_slope):
-			#t.origin.y = gt.origin.y if is_nan(height) else height - gt.origin.y
```
TokisanGames, line 83: Along with the comment above that `continues` if height is nan, this can just be:
```
t.origin.y = height - gt.origin.y
```

This will remove transforms over holes.

TokisanGames (changes requested):
Thanks for adding this for scatter users. My notes are inline.

TokisanGames:
Per the contributing doc, you should never submit a PR from main. Always use a branch. I'm now going to invalidate your main branch because of it. Rename it, check out main from upstream again, and delete the old.

I rebased and made further adjustments.

Testing this today, I found the latest version of Scatter to be unusably slow. I wouldn't use it. With less code than was written here you could scatter your own meshes or populate our instancer with 100x performance.

Thanks for the PR.

PR #692, opened 2025-05-13
Add tablet support

sci-comp:
I would like to have a tablet for texture painting. However, the problem I'm having is that my tablet doesn't automatically move the system cursor.

This proposed addition feels hacky, but it does work in my case.

Does anyone else use tablets for drawing?


TokisanGames:
We already enabled tablet support many months ago.

> the problem I'm having is that my tablet doesn't automatically move the system cursor.

Maybe this is a problem with your system/tablet/driver. How do you use your tablet with other painting programs if you can't move the mouse? What OS and tablet do you have?

sci-comp:
Ahh, I needed to install a driver for the tablet. It has always worked for notes, so I never thought to check.

Thank you, sorry for the foolish mistake!

PR #700, opened 2025-05-24
Reconfigure free_editor_textures

TokisanGames:
* Disables `free_editor_textures` by default
* If enabled it:
  * Force reloads the asset list on `_enter_tree()` (Only the list; textures are loaded only if not already loaded)
  * Generates the texture array on `_enter_tree()`
  * Unreferences the editor textures on `_ready()` (Godot unloads textures if not used elsewhere)

The demo has two demonstrations. 
1. If `free_editor_texture` is disabled, it prints the vram used at start, then manually clears the editor texture list, then prints the difference. This is the standard use case for beginning devs with one terrain in one scene.
2. If `free_editor_textures` is enabled, it removes Terrain3D, then a second later, loads and adds it to the scene as a scene loader would. Previously this would cause textures to be missing since they are cleared on the first _ready.

Using Godot's performance metrics, both of these tests result in a 5.3mb savings (4 textures @ 1.33mb).
Using Windows Task manager, the VRAM usage may appear higher or unchanged. I assume this is because of OS allocations Godot is managing. I am debating whether to leave this option enabled or disabled by default. Thoughts?

Would like a review and testing before I remove the debugging stuff.

### DIFF
--- a/doc/api/class_terrain3d.rst
+++ b/doc/api/class_terrain3d.rst
@@ -530,7 +530,7 @@ The verbosity of debug messages printed to the console. Errors and warnings are 
 - |void| **set_free_editor_textures**\ (\ value\: ``bool``\ )
 - ``bool`` **get_free_editor_textures**\ (\ )
 
-Frees ground textures used for editing at the start of the game. These textures are used to generate the TextureArrays, so if you don't change any :ref:`Terrain3DTextureAsset<class_Terrain3DTextureAsset>` settings in game, this can be enabled. Calls :ref:`Terrain3DAssets.clear_textures()<class_Terrain3DAssets_method_clear_textures>`.
+Frees ground textures used for editing in _ready(). These textures are used to generate the TextureArrays, so if you don't change any :ref:`Terrain3DTextureAsset<class_Terrain3DTextureAsset>` settings in game, this can be enabled. Also reloads the texture asset list in _enter_tree() in case you load scenes via code and need the textures again. Calls :ref:`Terrain3DAssets.clear_textures()<class_Terrain3DAssets_method_clear_textures>`.
 
 .. rst-class:: classref-item-separator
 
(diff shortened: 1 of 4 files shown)


Xtarsia:
I modified the demo script a bit to cylce back and forth between on / off

gives very repropducible results.
```
Current VRAM: 190982476
Freeing Terrain3D
VRAM after new scene loaded: 196583756
free_editor_textures is false, VRAM difference from last: 5601280

Current VRAM: 196583756
Freeing Terrain3D
Terrain3D#6000:_notification:855: Reloading Assets path: res://demo/data/assets.tres
Terrain3D#6000:_notification:857: Texture count: 2
Terrain3DAssets#6995:initialize:395: Assets path: res://demo/data/assets.tres
Terrain3D#6000:_notification:870: Clearing Textures, count: 0
VRAM after new scene loaded: 190982476
free_editor_textures is true, VRAM difference from last: -5601280

Current VRAM: 190982476
Freeing Terrain3D
VRAM after new scene loaded: 196583756
free_editor_textures is false, VRAM difference from last: 5601280

Current VRAM: 196583756
Freeing Terrain3D
Terrain3D#4933:_notification:855: Reloading Assets path: res://demo/data/assets.tres
Terrain3D#4933:_notification:857: Texture count: 2
Terrain3DAssets#5968:initialize:395: Assets path: res://demo/data/assets.tres
Terrain3D#4933:_notification:870: Clearing Textures, count: 0
VRAM after new scene loaded: 190982476
free_editor_textures is true, VRAM difference from last: -5601280

Current VRAM: 190982476
Freeing Terrain3D
VRAM after new scene loaded: 196583756
free_editor_textures is false, VRAM difference from last: 5601280
```

Xtarsia:
> Now that it force reloads the asset list, what do you think about leaving it on by default, or not?

I think it can stay on by default. I tested with web export quickly as well, and it worked just fine there too.

Issue #303, opened 2024-01-22
Change roughness modifier to make puddles


TokisanGames:
### Description

Painting wetness introduces a roughness modifier that can increase or decrease roughness. Consider instead of allowing rougher values (-0.5, 0.5), move the range to (-1, 0). (-0.5, 0) still increases wetness as is current. But below -0.5 have it flatten the normal to make puddles.

FishOfTheNorthStar:
I started a discussion about this concept recently, not the exact same implementation but pretty similar.  Here's a link:
https://github.com/TokisanGames/Terrain3D/discussions/375

There's an animated rain drop texture there I made that might come in handy.

PR #706, opened 2025-06-03
Fix wetness

TokisanGames:
* Clamps roughness in the shader. Previously range was about -1 to 3. Sub zero broke the reflections in wetness.
* Explicitly quantizes roughness so that what is painted and what is picked are the same.
Grok picked this one up.

Still planning on #303 in the future, maybe in 1.1.

---
Grok's Explanation of Quantization Code
Both code snippets address the issue of 8-bit quantization when storing dest.a (alpha channel, 0.0 to 1.0) in terrain_3d_editor.cpp and retrieving it in tool_settings.gd. They ensure the stored value aligns with the 8-bit precision (0 to 255) used by Godot, minimizing precision errors.

1. C++: float(int(dest.a * 255.0)) / 255.0; // Quantize explicitly
* Purpose: Forces dest.a (0.0 to 1.0) to the nearest value representable in 8-bit precision (multiples of 1/255).
* How it works:
  * dest.a * 255.0: Scales dest.a (0.0 to 1.0) to 0 to 255.
  * int(...): Truncates to the nearest integer (e.g., 127.6 → 127).
  * float(int(...)) / 255.0: Converts back to 0.0 to 1.0 (e.g., 127 → 127/255 ≈ 0.498039).
  * This mimics Godot’s 8-bit storage, ensuring dest.a matches what’s actually stored.
* Why: Without this, floating-point dest.a values (e.g., 0.5) may not align with 8-bit steps, causing slight offsets when read back (e.g., 0.5 → 0.498).

2. GDScript: float(int(p_color.a * 255.0)) / 255.0 - 0.5 in settings["roughness"].value = round(200 * (...))
* Purpose: Converts the stored p_color.a (0.0 to 1.0) back to the UI roughness range (-100 to 100) while accounting for 8-bit quantization.
* How it works:
  * p_color.a * 255.0: Scales p_color.a to 0 to 255.
  * int(...): Truncates to the nearest integer (mimics 8-bit storage).
  * float(int(...)) / 255.0: Converts back to 0.0 to 1.0 (e.g., 127 → 0.498039).
  * -0.5: Shifts to -0.5 to 0.5.
  * 200 * (...): Scales to -100 to 100 for the UI slider.
  * round(...): Ensures the UI value is an integer, avoiding fractional UI values.
* Why: Ensures the UI slider reflects the quantized value stored in p_color.a, preventing discrepancies (e.g., 0.498 → -0.4 instead of 0).

### DIFF
--- a/project/addons/terrain_3d/extras/shaders/lightweight.gdshader
+++ b/project/addons/terrain_3d/extras/shaders/lightweight.gdshader
@@ -349,8 +349,8 @@ void fragment() {
 		macrov = mix(vec3(1.0), macrov, clamp(v_normal.y + macro_variation_slope, 0., 1.));
 	}
 
-	// Wetness/roughness modifier, converting 0 - 1 range to -1 to 1 range
-	float roughness = fma(color_map.a - 0.5, 2.0, normal_rough.a);
+	// Wetness/roughness modifier, converting 0 - 1 range to -1 to 1 range, clamped to Godot roughness values 
+	float roughness = clamp(fma(color_map.a - 0.5, 2.0, normal_rough.a), 0., 1.);
 
 	// Apply PBR
 	ALBEDO = albedo_height.rgb * color_map.rgb * macrov;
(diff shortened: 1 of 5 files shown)


Xtarsia (approved):
Looks good.

Issue #690, opened 2025-05-07
Optimize get_intersection in non-gpu mode


TokisanGames:
### Description

* `get_intersection` iterates over `get_height`, which interpolates over 4 calls, unless rounded to the nearest vertex, in which case it only does one call. So snap to the nearest vertex before every call. I considered doing it once before the loop, but if it's off axis, it will always need to be snapped.

* We iterate one meter at a time, which is inaccurate if vertex_spacing != 1. `direction` should be multipled by vertex_spacing after being normalized, as should 4000. However if vs is sub 1, you'll need to not snap to the nearest vertex if the distance is less than 1.

Bonus points: 
* Rather than iterating linearly, iterate like quick sort does, which will change the algorithm from O(n) to O(n log n). Most checks are going to be sub 1000m, likely sub 500m and even 100m. So I'd start looking at 100m, then:
  * If passed the terrain, check at 50m then continue to divide the remaining in half.
  * If it hasn't passed, check 1000m, 10,000m, 100,000m. Then divide the remaining section in half. Somewhat like the following picture.

![image](https://upload.wikimedia.org/wikipedia/commons/thumb/7/72/Visualization_of_SDF_ray_marching_algorithm.png/1920px-Visualization_of_SDF_ray_marching_algorithm.png)

aidandavey:
I'd like to shoot for the bonus points, but I can't see how this could work. 

If we do the first check at 100m, how do we know we didn't miss something?

In the example below, the first check would register a miss, and start the next partition along.

![Image](https://github.com/user-attachments/assets/46d25463-a3b7-4986-9f6b-8360814d6628)

To catch this we'd need to start from the origin, as it does now.

TokisanGames:
You can check at 0 initially. You first want to establish outside bounds. So, 0 and 100 are good initial checks based on practical use. If 100 is above the terrain, increase 10x each time until 100,000.

Once you have outside bounds, check at the midpoint of your range.  You'll be testing points above and below the true terrain height, getting closer and closer to it until within tolerance. (eg <0.001). This is sphere assisted ray marching, though it's not mentioned here. 

* Look at 0, if below the terrain, quit w/ MAX
* Look at 100, if within -.001 quit w/ height
* If above the terrain at 100, look at 1000
  * if above at 1000, look at 10,000, then 100,000 (maximum)
  * if below at 1000, look at 500
* If below at 100, look at 50
  * if above at 50, look at 75
  * if below at 50, look at 25

[...]

PR #712, opened 2025-06-09
Optimise get_intersection()

aidandavey:
This PR addresses issue #690

Run up to 800 primary steps at a step size of 5 to find the upper and lower bounds.

Run up to 5 secondary steps (in practice less than 4 are usually required) to find an intersection point with a height difference of less than 0.001

The secondary steps uses a plane intersection calculation to contract the U and L bounds accordingly.

At short ranges, we get a moderate speed (~2x) improvement and at long ranges it is much faster. To detect no intersection is ~10x faster. 

In addition, we get a more accurate result since the secondary steps are not snapped to 1m or even vertex spacing. 

Typically it returns a result between 0.01 and 0.001 of the result from a physics ray cast. Original implementation was >0.5

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -687,37 +687,70 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
 		}
 
 	} else if (!p_gpu_mode) {
-// Else if not gpu mode, use raymarching mode
-
+		int time = Time::get_singleton()->get_ticks_usec();
+		// Else if not gpu mode, use raymarching mode
 		direction *= _vertex_spacing;
-		real_t height;
-		Vector3 snapped_point;
-		Vector3 last_point;
+		real_t height = FLT_MAX;
+		Vector3 snapped_point = V3_MAX;
+		Vector3 last_point = p_src_pos;
+		const int PRIMARY_STEP_SIZE = 5;
+		const int MAX_SECONDARY_STEPS = 5;
+		const real_t height_difference_tolerance = 0.001f;
 
-		for (int i = 0; i < 2000; i++) {
-			point = p_src_pos + (direction * i * 2);
-			if (i) {
-				last_point = p_src_pos + (direction * (i - 1) * 2);
-			} else {
-				last_point = point;
-			}
+		for (int i = 0; i < 4000; i += PRIMARY_STEP_SIZE) {
+			point = p_src_pos + (direction * i);
 
 			snapped_point = point.snapped(Vector3(_vertex_spacing, point.y, _vertex_spacing));
 			height = _data->get_height(snapped_point);
 
+			// Early return if outside of regions
 			if (isnan(height)) {
+				LOG(EXTREME, "Determined no intersection in : ", Time::get_singleton()->get_ticks_usec() - time, " us");
 				return V3_MAX;
 			}
 
 			if (snapped_point.y - height <= 0.0f) {
-				height = _data->get_height(last_point);
-				if (last_point.y - height <= 0.0f) {
-					return last_point;
-				}
-				return point;
-			}
-		}
+				real_t height_difference = FLT_MAX;
 
+				Vector3 u_bound = point;
+				Vector3 l_bound = last_point;
+				Vector3 u_terrain = Vector3(point.x, _data->get_height(point), point.z);
+				Vector3 l_terrain = Vector3(last_point.x, _data->get_height(last_point), last_point.z);
+
+				Vector3 estimated_intersection = V3_MAX;
+				
+				for (int j = 0; j < MAX_SECONDARY_STEPS; j++) {		
+					
+					// Define 4-vertex plane from any three vertices
+					Plane plane(l_terrain, u_terrain, l_terrain + l_terrain.direction_to(p_src_pos).cross(l_terrain.direction_to(u_terrain)));
+
+					// Compute ray-plane intersection
+					float t = -plane.distance_to(p_src_pos) / plane.get_normal().dot(p_direction);
+
+					estimated_intersection = p_src_pos + p_direction * t;
+			
+					height = _data->get_height(estimated_intersection);
+					height_difference = estimated_intersection.y - height;
+
+					if (abs(height_difference) <= height_difference_tolerance) {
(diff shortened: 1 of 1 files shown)


src/terrain_3d.cpp
```diff
@@ -687,15 +687,70 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
...
+			snapped_point = point.snapped(Vector3(_vertex_spacing, point.y, _vertex_spacing));
+			height = _data->get_height(snapped_point);
+
+			// Early return if outside of regions
+			if (isnan(height)) {
+				LOG(EXTREME, "Determined no intersection in : ", Time::get_singleton()->get_ticks_usec() - time, " us");
+				return V3_MAX;
+			}
+
+			if (snapped_point.y - height <= 0.0f) {
```
TokisanGames: This new function has two steps: find bounds, then refine. But structurally you have them nested. Once it comes into this conditional, it never leaves, so it doesn't need to be nested. I'd like to see this check break out of the loop, then process the second loop on the previous indent level for ease of reading and comprehension.

I'd also like more comments defining the two broad steps, and some details within so future contributors understand the logic.

src/terrain_3d.cpp
```diff
@@ -687,15 +687,70 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
...
+				real_t height_difference = FLT_MAX;
+
+				Vector3 u_bound = point;
+				Vector3 l_bound = last_point;
+				Vector3 u_terrain = Vector3(point.x, _data->get_height(point), point.z);
+				Vector3 l_terrain = Vector3(last_point.x, _data->get_height(last_point), last_point.z);
```
TokisanGames: I'd like you to spell out the variables and comment what these are for. upper_bound is more clear than u_bound. u_terrain is unclear. 

upper/lower_bound are within the cast ray, u/l_terrain are the vertices of the face the ray has hit. This isn't clear from the names, so needs to be explained in comment.

src/terrain_3d.cpp
```diff
@@ -687,15 +687,70 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
...
+
+				Vector3 estimated_intersection = V3_MAX;
+				
+				for (int j = 0; j < MAX_SECONDARY_STEPS; j++) {		
+					
+					// Define 4-vertex plane from any three vertices
+					Plane plane(l_terrain, u_terrain, l_terrain + l_terrain.direction_to(p_src_pos).cross(l_terrain.direction_to(u_terrain)));
+
+					// Compute ray-plane intersection
+					float t = -plane.distance_to(p_src_pos) / plane.get_normal().dot(p_direction);
```
TokisanGames: All of these functions return real_t, so you should use it too.

TokisanGames (changes requested):
Thanks for the start on this. I edited editor_plugin.gd and changed the get_intersection line to not use GPU mode for testing and it works well in most cases.

* If the camera is not above a region, it won't work. I don't know if this worked before, but it's something we should account for. I don't see any early exit for is_nan in the old code.

* I made a mistake in assumptions. Plane assumes 4 vertices (ABCD) on the same plane, and that the two triangles forming the plane (ABC and ACD) have the same normal. But that is not correct. Each vertex has its own height, so each triangle has its own normal. So, we need to do a ray/triangle intersection instead, and we need the height of the third vertex.

I had a conversation with grok about it who suggested a different algorithm, Möller-Trumbore ray-triangle intersection.  [Peruse it here](https://grok.com/share/bGVnYWN5_e6a6efcd-fe08-4db8-86d1-c8d5486206b4).

aidandavey:
Thanks for reviewing, I'll work through the restructuring and naming and update soon.

You are right about the early return on NaN. I didn't consider the ray might start outside a region, now I think about it we may want to cross an empty region too, islands for example.

Grok really wants us to snap y! glad it saw sense in the end.

I think you were right first time and planes work... there is a constructor to create a plane from 3 points. The normal is determined by the winding order. 

There are no triangles in a plane, it's infinite and only has a normal and distance from the origin.

I am not using the vertices of the terrain grid, at the secondary stage we are using interpolated heights from the bounds which could easily be between vertices.

Still maybe triangles are faster or something so happy to experiment.

src/terrain_3d.cpp
aidandavey: Loops have been restructured as recommended and comments added

src/terrain_3d.cpp
aidandavey: Added explanation and renamed these variables. 
upper/lower_ray_bound
upper/lower_triangle_point

Added a new variable to make clear where we create the third point of the triangle.

src/terrain_3d.cpp
aidandavey: Done

aidandavey:
> I had a conversation with grok about it who suggested a different algorithm, Möller-Trumbore ray-triangle intersection. [Peruse it here](https://grok.com/share/bGVnYWN5_e6a6efcd-fe08-4db8-86d1-c8d5486206b4).

I read your conversation and experimented with the ray-triangle intersection process Grok suggested. Used as shown, it takes ~40-100 us (depending on the distance) to find a valid or approximated intersection since it takes four cheap height checks and one expensive one per secondary step rather than a single expensive one. It also has to check two triangles rather than one plane. It also suffers when the intersection point is close to the vertex.  There is no difference in the primary step section, so a fail takes the same length of time as the version I submitted.

When refining it is exceeding max steps more often than not (~25% results within 0.001 tolerance). 

I then swapped out the plane intersection method in my version for the triangle intersection method and there was practically no speed difference (17-28 us within a similar distance range).  Typically it takes max 3 steps to get a result within tolerance. 

On balance, using Godots built in plane methods seems like the better option to me.

aidandavey:
I wonder whether it's worth including a physics raycast before attempting raymarching. If we get a hit on the terrain we're done, if not proceed to ray marching.

For mouse picking, this is slower for shorter results but stable and much faster for longer distance checks. Maybe not worthwhile, since you'll probably often be painting at relatively short distances. 

Though I think a physics based check could be useful for other issues, such as foliage instancing on collision objects.

Like this:

image

PR #713, opened 2025-06-09
Shader performance pass

### LINKED ISSUE #704: Performance regression in mobile
### Terrain3D version

Build from  #1397 interpolate blend value. 

### System information

Godot 4.4 stable sansung s24 ultra 12 Gb Ram

Yes

### Issue description

It does look like when going from the 1.0.0 stable release to this build with the improved blend, the performance gets much worst , meanwhile using the lightweight shader gives most of those frames backs

Some screenshots of performance between stable and the build: and also with the build lightweight shader

Image
Image
Image

### Logs

Later.

Xtarsia:
closes #704 

Vectorised a lot of math where possible.
Removed the control data struct.
Reduced operation count by a decent amount, including more fma() use, as well as reusing / avoiding additonal sin/cos calls, and intermediate values.

note: In compatibility, the fma() override requires all nested fma()'s to be on a single line.

Overall faster, and brings mobile back to 95-98% on my Samsung S23 Ultra.

### DIFF
--- a/doc/docs/shader_design.md
+++ b/doc/docs/shader_design.md
@@ -84,7 +84,7 @@ Generating normals in the shader works fine, and modern GPUs can easily handle t
 
 The control maps are queried for each of the 4 adjacent grid points (aka vertices and indices) and stored in `control[0]`-`control[3]`.
 
-The control map bits, as defined in [Controlmap Format](controlmap_format.md), are parsed into an array of control_data structs, which include the decoded `texture_id`, `texture_weight`, `blend`, `scale`, and `angle` values.
+The control map bits, are decoded when needed, as defined in [Controlmap Format](controlmap_format.md).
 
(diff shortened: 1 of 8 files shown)


TokisanGames:
@Saul2022, please test this.

Saul2022:
Yep think seems to be well for the most part.
Notes this test were done under a bit of overheating so fps will normally be much higher 

Screenshot_20250611_004903
Screenshot_20250611_004859

Overall it recovers almost all performance.  Pretty well. And hopefully when rvt texture arrives it might improve fps to 50% or 60% but that's just speculation.

Xtarsia:
Force pushed some comment changes, and 1 more fma into the lieghtweight shader. Glad that Performance is mostly back now! 

Also blend sharpness is not a linear value anymore, so something like 0.2 is closer to the pre-blend-gate value of 0.87.

This should be ready, pending review.

TokisanGames:
Looks great, thank you!

PR #715, opened 2025-06-10
Fix invalid identifiers used as parameter names

Ryan-000:


TokisanGames:
Thanks. Please change the parameters to `save_16_bit` which matches `Terrain3D.save_16_bit`.
Also change the doc_classes XML files for Data and Region. You can just edit them by hand. I'll regenerate the RSTs later.

TokisanGames:
Thanks.

PR #717, opened 2025-06-11
Dynamic collision and clipmap target overrides

### LINKED ISSUE #703: Custom reference point for dynamic collision generation
### Description

Currently, dynamic collision is always generated around the camera - which does not make a whole lot of sense for third-person games, especially for the extreme case of 2.5D where camera could have a narrow FoV while being really far away. What I propose is adding a `dynamic_collision_reference` property to `Terrain3D`/`Terrain3DCollision` that can point to a node (most likely the player caracter) that'll act as the center around which collisions are generated - if unset, it can fall back to the current behavior of using the camera.

P.S.: the link in the feature request issue template leads to a 404 error page.

aidandavey:
Fixes #703

Added the ability to select any Node3D to act as the clipmap and dynamic collision targets.

If no node is assigned, falls back to using the camera.

image

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -112,26 +112,16 @@ void Terrain3D::_initialize() {
 void Terrain3D::__physics_process(const double p_delta) {
 	if (!_initialized)
 		return;
-
 	if (!_camera.is_valid()) {
 		LOG(DEBUG, "Camera is null, getting the current one");
 		_grab_camera();
 	}
-
-	// If clipmap target has moved enough, re-center the terrain on it.
-	Vector3 target_pos = get_clipmap_target_position();
-	Vector2 target_pos_2d = Vector2(target_pos.x, target_pos.z);
-	RS->material_set_param(_material->get_material_rid(), "_camera_pos", target_pos);
-	if (_camera_last_position.distance_to(target_pos_2d) > 0.2f) {
-		if (_mesher) {
-			_mesher->snap(target_pos);
-		}
-		_snapped_position = (target_pos / _vertex_spacing).floor() * _vertex_spacing;
-		_camera_last_position = target_pos_2d;
+	if (_mesher) {
+		_mesher->snap();
 	}
-
-	// For collision, distance moved and re-centering is handled internally.
-	_collision->update();
+	if (_collision) {
+		_collision->update();
+	}
 }
 
 /**
@@ -146,9 +136,9 @@ void Terrain3D::_grab_camera() {
 		_camera.set_target(get_viewport()->get_camera_3d());
 		LOG(DEBUG, "Grabbing the in-game viewport camera: ", _camera.get_target());
 	}
-	if (!_camera.is_valid() && !_clipmap_target_override.is_valid()) {
-		set_physics_process(false); // disable snapping
-		LOG(ERROR, "Cannot find clipmap target or the active camera. Set it manually with Terrain3D.set_camera(). Stopping _physics_process()");
+	if (!_camera.is_valid() && !_clipmap_target.is_valid()) {
+		set_physics_process(false); // No target to follow, disable snapping until one set
+		LOG(ERROR, "Cannot find clipmap target or active camera. LODs won't be updated. Set manually with set_clipmap_target() or set_camera()");
 	}
 }
 
@@ -461,65 +451,73 @@ void Terrain3D::set_assets(const Ref<Terrain3DAssets> &p_assets) {
 }
 
 void Terrain3D::set_editor(Terrain3DEditor *p_editor) {
+	if (p_editor && p_editor->is_queued_for_deletion()) {
+		LOG(ERROR, "Attempted to set a node queued for deletion");
+		return;
+	}
+	LOG(INFO, "Set Terrain3DEditor: ", p_editor);
 	_editor = p_editor;
 	if (_material.is_valid()) {
 		_material->update();
 	}
-	LOG(DEBUG, "Received Terrain3DEditor: ", p_editor);
 }
 
 void Terrain3D::set_plugin(EditorPlugin *p_plugin) {
+	if (p_plugin && p_plugin->is_queued_for_deletion()) {
+		LOG(ERROR, "Attempted to set a node queued for deletion");
+		return;
+	}
+	LOG(INFO, "Set EditorPlugin: ", p_plugin);
 	_plugin = p_plugin;
-	LOG(DEBUG, "Received editor plugin: ", p_plugin);
+}
+
+void Terrain3D::set_clipmap_target(Node3D *p_node) {
+	if (p_node && p_node->is_queued_for_deletion()) {
+		LOG(ERROR, "Attempted to set a node queued for deletion");
+		_clipmap_target.clear();
+		return;
+	}
+	LOG(INFO, "Setting clipmap target: ", p_node);
+	_clipmap_target.set_target(p_node);
+	set_physics_process(true);
+}
+
+Vector3 Terrain3D::get_clipmap_target_position() const {
+	if (_clipmap_target.is_inside_tree()) {
+		return _clipmap_target.ptr()->get_global_position();
+	}
+	if (_camera.is_inside_tree()) {
+		return _camera.ptr()->get_global_position();
+	}
+	return V3_ZERO;
+}
+
+void Terrain3D::set_collision_target(Node3D *p_node) {
+	if (p_node && p_node->is_queued_for_deletion()) {
+		LOG(ERROR, "Attempted to set a node queued for deletion");
+		_collision_target.clear();
+		return;
+	}
+	LOG(INFO, "Setting collision target: ", p_node);
+	_collision_target.set_target(p_node);
+}
+
+Vector3 Terrain3D::get_collision_target_position() const {
+	if (_collision_target.is_inside_tree()) {
+		return _collision_target.ptr()->get_global_position();
+	}
+	return get_clipmap_target_position();
 }
 
 void Terrain3D::set_camera(Camera3D *p_camera) {
-	LOG(INFO, "Setting camera");
+	if (p_camera && p_camera->is_queued_for_deletion()) {
+		LOG(ERROR, "Attempted to set a node queued for deletion");
+		_camera.clear();
+		return;
+	}
+	LOG(INFO, "Setting camera: ", p_camera);
 	_camera.set_target(p_camera);
-}
-
-Node3D *Terrain3D::get_collision_target_override() {
-	return _collision_target_override.get_target();
-}
-
-void Terrain3D::set_collision_target_override(Node3D *p_node) {
-	LOG(INFO, "Setting collision target override");
-	_collision_target_override.set_target(p_node);
-}
-
-Vector3 Terrain3D::get_collision_target_position() {
-	if (_collision_target_override.is_valid()) {
-		return _collision_target_override.get_global_position();
-	}
-	if (_camera.is_valid()) {
-		return _camera.get_global_position();
-	}
-	if (_clipmap_target_override.is_valid()) {
-		return _clipmap_target_override.get_global_position();
-	}
-
-	return Vector3(NAN, NAN, NAN);
-}
-
-Node3D *Terrain3D::get_clipmap_target_override() {
-	return _clipmap_target_override.get_target();
-}
-
-void Terrain3D::set_clipmap_target_override(Node3D *p_node) {
-	LOG(INFO, "Setting clipmap target override");
-	_clipmap_target_override.set_target(p_node);
-}
-
-Vector3 Terrain3D::get_clipmap_target_position() {
-	if (_clipmap_target_override.is_valid()) {
-		return _clipmap_target_override.get_global_position();
-	}
-
-	if (_camera.is_valid()) {
-		return _camera.get_global_position();
-	}
-
-	return Vector3(NAN, NAN, NAN);
+	set_physics_process(true);
 }
 
 void Terrain3D::set_region_size(const RegionSize p_size) {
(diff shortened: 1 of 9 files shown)


aidandavey:
If the concept is sound I'll have a go at updating the documentation

src/terrain_3d_collision.h
```diff
@@ -39,6 +39,7 @@ class Terrain3DCollision : public Object {
 	uint32_t _mask = 1;
 	real_t _priority = 1.f;
 	Ref<PhysicsMaterial> _physics_material;
+	Node3D *_custom_collision_node3d = nullptr;
```
TokisanGames: 1. This and all references to it need to be a `Ref<Node3D>`. Anything passed through the gdscript interface should be reference counted. The camera is fetched internally so is handled differently. It is not the example to follow.
2. Use `_collision_target_override`.

src/terrain_3d_collision.cpp
```diff
@@ -554,6 +558,12 @@ RID Terrain3DCollision::get_rid() const {
 	return RID();
 }
 
+void Terrain3DCollision::set_custom_collision_node(Node3D *p_node) {
+	if (_custom_collision_node3d != p_node) {
```
TokisanGames: This needs a `LOG(INFO, text)`

src/terrain_3d.cpp
```diff
@@ -128,11 +128,19 @@ void Terrain3D::__physics_process(const double p_delta) {
...
+	// Update collision
+	if (!get_custom_collision_node() || !get_custom_collision_node()->is_inside_tree()) {
+		if (_camera) {
+			LOG(DEBUG, "Custom collision node is null, setting to camera");
+			set_custom_collision_node(_camera);
+		}
+	}
+	if (_collision && _collision->is_dynamic_mode()) {
+		_collision->update();
+	}
```
TokisanGames, line 124: * Define get_clipmap_target_position() and get_collision_target_position(). 
  * Move the checks here to there.
  * Both check the validity of their target_override, and fall back to the camera if invalid. If collision_target_override is invalid, fall back to the camera if it can grab it, otherwise use the clipmap_target_override.
* Move the mesh only if the clipmap target moves (current code)
* Update collision if the collision target moves

src/terrain_3d.cpp
```diff
@@ -1145,6 +1155,7 @@ void Terrain3D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "collision_mask", PROPERTY_HINT_LAYERS_3D_PHYSICS), "set_collision_mask", "get_collision_mask");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "collision_priority", PROPERTY_HINT_RANGE, "0.1,256,.1"), "set_collision_priority", "get_collision_priority");
 	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "physics_material", PROPERTY_HINT_RESOURCE_TYPE, "PhysicsMaterial"), "set_physics_material", "get_physics_material");
+	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "custom_collision_node", PROPERTY_HINT_NODE_TYPE, "Node3D"), "set_custom_collision_node", "get_custom_collision_node");
```
TokisanGames: I like setting the object via the API, but how did this work in the inspector? I see your screenshot showing the player attached. How did you set that in the inspector? A NodePath would bring up a Scene Tree selector. Did this configuration also bring that up?

src/terrain_3d_collision.cpp
```diff
@@ -295,8 +295,12 @@ void Terrain3DCollision::update(const bool p_rebuild) {
 	real_t spacing = _terrain->get_vertex_spacing();
 
 	if (is_dynamic_mode()) {
+		if (!_custom_collision_node3d) {
+			LOG(WARN, "Collision reference node was null. Could not update collision");
+			return;
+		}
 		// Snap descaled position to a _shape_size grid (eg. multiples of 16)
-		Vector2i snapped_pos = _snap_to_grid(_terrain->get_snapped_position() / spacing);
+		Vector2i snapped_pos = _snap_to_grid(_custom_collision_node3d->get_global_position() / spacing);
```
TokisanGames: This will crash if the node has been freed.

Instead of both of these changes, have this query `_terrain->get_collision_target_position()`.

src/terrain_3d.cpp
```diff
@@ -128,11 +128,19 @@ void Terrain3D::__physics_process(const double p_delta) {
...
-				_collision->update();
-			}
 		}
 	}
+
+	// Update collision
+	if (!get_custom_collision_node() || !get_custom_collision_node()->is_inside_tree()) {
+		if (_camera) {
+			LOG(DEBUG, "Custom collision node is null, setting to camera");
+			set_custom_collision_node(_camera);
```
TokisanGames: This may add the camera to the inspector, and potentially the scene file upon save, neither of which is desirable. It should remain empty unless they set it. Use the camera as an internal target rather than setting the public variable.

src/terrain_3d.cpp
```diff
@@ -128,11 +128,19 @@ void Terrain3D::__physics_process(const double p_delta) {
...
 			_snapped_position = (cam_pos / _vertex_spacing).floor() * _vertex_spacing;
 			_camera_last_position = cam_pos_2d;
-			if (_collision && _collision->is_dynamic_mode()) {
-				_collision->update();
-			}
 		}
 	}
+
+	// Update collision
+	if (!get_custom_collision_node() || !get_custom_collision_node()->is_inside_tree()) {
```
TokisanGames: This will crash if the node is freed, due to a dangling pointer.

TokisanGames:
Thanks for this. 

We discussed this on discord. Please extend this to provide a custom clipmap target as well to override the camera. Call them `clipmap_target_override` and `collision_target_override`. The first can be at the top of Mesh, the second right below collision_mode.

If either the camera or clipmap target override are set, avoid the `can't find camera` warning.

This won't go out in the pending 1.0.1 release, but will target 1.1.

src/terrain_3d.cpp
aidandavey: Yes it brings up the selector which is filtered to only show Node3D or derived

src/terrain_3d_collision.h
aidandavey: I can't get this to compile, seems Node3D does not inherit from Ref if that makes a difference. This is the console log:

> 
> godot-cpp\include\godot_cpp/classes/ref.hpp(204): error C2039: 'unreference': is not a member of 'godot::Node3D'
> godot-cpp\gen\include\godot_cpp/classes/node3d.hpp(55): note: see declaration of 'godot::Node3D'
> godot-cpp\include\godot_cpp/classes/ref.hpp(204): note: the template instantiation context (the oldest one first) is
> C:\Users\adave\Documents\Terrain3D-Collision\src\terrain_3d.h(72): note: see reference to class template instantiation 'godot::Ref<godot::Node3D>' being compiled
> godot-cpp\include\godot_cpp/classes/ref.hpp(203): note: while compiling class template member function 'void godot::Ref<godot::Node3D>::unref(void)'
> godot-cpp\include\godot_cpp/classes/ref.hpp(217): note: see the first reference to 'godot::Ref<godot::Node3D>::unref' in 'godot::Ref<godot::Node3D>::~Ref'
> C:\Users\adave\Documents\Terrain3D-Collision\src\terrain_3d.h(128): note: see the first reference to 'godot::Ref<godot::Node3D>::~Ref' in 'Terrain3D::~Terrain3D'
> scons: *** [src\generated_texture.windows.template_debug.x86_64.obj] Error 2
> 

```
Ref<Node3D> _collision_target_override;
```

TokisanGames: You're right. My mistake. Ref<> is for passing Resources. In that case, you need to do the opposite and fully follow the camera example. Since the node can be freed at any time, and not by you, you need to test if the instance is valid. However, the godot-cpp function is not reliable so you have to track the instance id.

You could make a struct UserNode3D { int instance_id, Node3D *node} and include is_instance_valid() and other relevant functions for setting, getting, and checking so it's a self-contained object. Then have the _camera and the two override nodes all have this same type. This will also differentiate between pointers of objects we're creating and need to destroy, and user provided nodes that we have no responsibility over.

aidandavey: Any thoughts on where this struct should be defined? ATM I have it as the top of Terrain3D.h

TokisanGames: That sounds fine for now.

src/terrain_3d.h
```diff
@@ -23,6 +23,38 @@
 
 using namespace godot;
 
+// Does this struct belong here?
+struct TargetNode3D {
```
v1993: Doesn't this somewhat reimplement Godot's built-in `WeakRef` class?

TokisanGames: We're reimplementing [is_instance_valid()](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-method-is-instance-valid), which was removed from godot-cpp because it didn't work reliably.

`WeakRef.get_ref()` provides a `Variant`. The question is, can we test the validity of a freed reference with it in godot-cpp, without using is_instance_valid? I see it does have get_validated_object():
```
Object *Variant::get_validated_object() const {
	if (type == OBJECT) {
		return ObjectDB::get_instance(_get_obj().id);
	} else {
		return nullptr;
	}
}

```

So WeakRef may work. But it needs to be tested against an object that has been freed. If it properly changes the variant to type Nil, or returns nullptr from get_validated_object() then yes we can use it and drop our struct.

aidandavey: Using the struct allows us to extend it and hold the last_snapped_position, the snap radius, possibly the fallback target, define should_snap() etc...

The struct specifically defines and handles the logic of a Node3D which will be used as a target, rather than being a WeakRef to any object (maybe this should be a class rather than struct, though the logic is not too complex)

Either way I think this will result in cleaner code within the classes which use the target. 

WeakRef might still be useful within the struct/class assuming it does return nullptr if the target is freed (which I expect it will).

I'll try it out and report back here.

TokisanGames: You could also extend WeakRef.

aidandavey: 🤯

v1993: > `WeakRef.get_ref()` provides a `Variant`. The question is, can we test the validity of a freed reference with it in godot-cpp, without using is_instance_valid?

The whole idea of `WeakRef` is that it validates the reference it stores before handing it out (and has special handling for refcounted objects, but that's not relevant for nodes); if it has been destroyed, an empty/NIL `Variant` is returned. So yes, `get_validated_object` should work.

FWIW, `WeakRef` [internally](https://github.com/godotengine/godot/blob/46c495ca21f40f57a7fb9c7cde6143738f1652d4/core/object/ref_counted.cpp#L102-L129) just stores an instance ID and uses `ObjectDB::get_instance(id)` to get the object or null back. So maybe just do that instead of bringing the entire class in and dealing with `Variant`... or stick with the existing method. Idk.

TokisanGames: > FWIW, WeakRef internally just stores an instance ID and uses ObjectDB::get_instance(id) to get the object or null back. So maybe just do that instead of bringing the entire class in and dealing with Variant...  

That is what we were doing before you brought up weakref. I thought you had a particular idea about efficiency or code reduction. I don't particularly care for Variants. Without it we can ensure everything is a Node3D. 

@aidandavey perhaps what you've done here originally is fine. Is_instance_valid can be moved from Util, as it's only used for object references the user provided and controls like the camera.

src/terrain_3d.cpp
```diff
@@ -111,28 +111,25 @@ void Terrain3D::__physics_process(const double p_delta) {
...
-				_mesher->snap(cam_pos);
-			}
-			_snapped_position = (cam_pos / _vertex_spacing).floor() * _vertex_spacing;
-			_camera_last_position = cam_pos_2d;
-			if (_collision && _collision->is_dynamic_mode()) {
-				_collision->update();
-			}
+	// If clipmap target has moved enough, re-center the terrain on it.
+	Vector3 cam_pos = get_clipmap_target_position();
+	Vector2 cam_pos_2d = Vector2(cam_pos.x, cam_pos.z);
```
Xtarsia: For clarity, its worth renaming these to target_pos etc

Xtarsia (commented):
Works well!

src/terrain_3d.cpp
aidandavey: Done, thanks!

TokisanGames:
Closing in on merging.

API changes: 
* get_snapped_position() ->
  * get_clipmap_target_position()
  * get_collision_target_position()

Tested:
* setting separate targets
* freeing targets mid game
* setting targets mid game
* setting null targets mid game
* removing targets from tree
* adding targets back to tree
* load scene via script
* after physics process stopped, resume w/ set_camera, set*target
* pass null to set_camera, set_*target
* pass freed instance
  * This will crash, but I don't see any way around it. We have no way to determine if the instance is a dangling pointer or not.

TokisanGames:
Alright, I have addressed this case:

```
var node = Node3D.new()
node.queue_free()
terrain.set_clipmap_target(node)
```

This also doesn't crash:
```
var node = Node3D.new()
node.free()
terrain.set_clipmap_target(node)
```

But both of these crash and there's nothing we can really do about it:
```
var node = Node3D.new()
node.free()
await get_tree().create_timer(1).timeout
terrain.set_clipmap_target(node)

var node = Node3D.new()
node.queue_free()
await get_tree().create_timer(1).timeout
terrain.set_clipmap_target(node)
```

However setting the target as null or freeing it while set work fine:

```
var node := Node3D.new()
add_child(node)
node.global_position = Vector3(1, 1, 1)
$Terrain3D.set_clipmap_target(node)
await get_tree().create_timer(5).timeout
node.free()
```

TokisanGames:
Thanks @aidandavey !

PR #719, opened 2025-06-13
per id projection setting

### LINKED ISSUE #718: Limit Projection to Cliff Faces
### Description

Consider limiting 3D Projection to cliff faces.

What if we had a single uint32 uniform that had bits for all textures that are marked for cliff faces and projection and dual scaling only applies to those? I might want to paint different rocks on different cliff faces. But I don't really need grass or dirt vertically projected. They will be painted in for variation only, not large faces unless on gentle slopes.

I'd like to avoid artifacts on non-cliff faces. Currently projection is a little difficult to use because you have to balance getting it high enough to cover the cliff faces, but low enough that it doesn't cause your grass to warp.

Image

Xtarsia:
Fairly straight forwards, however the current implementation in the shader is costing a bit more fps than i'd hope - around a 2-5% drop.

closes #718

### DIFF
--- a/project/demo/data/assets.tres
+++ b/project/demo/data/assets.tres
@@ -49,7 +49,7 @@ albedo_texture = ExtResource("3_g8f2m")
 normal_texture = ExtResource("4_aw5y1")
 normal_depth = 1.0
 ao_strength = 2.0
-uv_scale = 0.2
+uv_projection = false
 detiling_rotation = 0.161
 
 [resource]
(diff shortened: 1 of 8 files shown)


Saul2022:
A simple fix might just be a toggle to turn that on or off, like the dual scaling. Anyways if not RVT will most likely make the shader don’t cost that much (atleast on mobile if it bakes  textures).

Edit after testing it seems that when opening the project using this pr artficact the shader breaks on mobile  resulting as if it had no shader.

Xtarsia:
> Edit after testing it seems that when opening the project using this pr artficact the shader breaks on mobile resulting as if it had no shader.

Many thanks for catching this! It was actually causing godot to crash on my phone, it should be fixed now.

heavy use of the ternary operator is a no-go it appears!

TokisanGames:
Very nice, much easier to work with. Thank you.

Btw, did you know that on your merged PRs you have a button that says "delete branch". Below? That will automatically delete the branch off of your server fork.

PR #720, opened 2025-06-15
Update docs

TokisanGames:
@Xtarsia Would you review [this doc](https://github.com/TokisanGames/Terrain3D/blob/29b31dc5c34e0cfede620bfabc30842be106fca3/doc/docs/tips_environment.md) please? You can do the whole thing if you like, but I'm specifically interested in a review of the foliage material and shader section, and any other tips you have.


doc/docs/tips_environment.md
```diff
@@ -0,0 +1,290 @@
...
+```glsl
+uniform float alpha_scissor_threshold : hint_range(0,1) = 0.5;
+uniform float alpha_antialiasing_edge : hint_range(0,1) = 0.3;
+
+void fragment() {
+	ALPHA = clamp(albedo_tex.a * max(log(-VERTEX.z), 1.0), 0.0, 1.0);
+	ALPHA_SCISSOR_THRESHOLD = alpha_scissor_threshold;
+	ALPHA_ANTIALIASING_EDGE = alpha_antialiasing_edge;
+}
+```
```
Xtarsia, line 250: It could also be explained that this is partially due to mipmap averaged alpha values, where as the mip level increases, the alpha can decrease due to more and more 0 alpha pixel being included, eg 51% of a leaf texture is alpha 0, so with a clip threshold of 0.5, mip0 will look as desired, but as mip level increases, edges get blended under the threshold, giving the shrinking effect we usually see. (disabling mipmaps for alpha clip textures will stop this happening, but.. results garbage noisy pixels)

A lot of games use MAX comparison when generating alpha clip texture mipmaps to help with this, though godot doesnt have this as an option for now.

doc/docs/tips_environment.md
```diff
@@ -0,0 +1,290 @@
...
+uniform float specular_distance_fade : hint_range(0,1000,.1) = 15.;
+
+void fragment() {
+	vec3 camera_pos = INV_VIEW_MATRIX[3].xyz;
+	vec3 pixel_pos = (INV_VIEW_MATRIX * vec4(VERTEX,1.0)).xyz;
+	SPECULAR = mix(specular, 0., clamp(length(pixel_pos - camera_pos) / specular_distance_fade, 0., 1.));
+}
+```
+
+* Some foliage renders with a white edge even though it's not visible on the texture. Check if it's an issue with roughness or specular and adjust as above. However if the issue is in ALBEDO, this line placed *after* assigning the texture to ALPHA will allow you to clip off white pixels.
```
Xtarsia: Another possible more likley cause of edge artifacting here, is textures not having the `repeat disable` flag, when they should.

doc/docs/tips_environment.md
```diff
@@ -0,0 +1,290 @@
...
+
+```glsl
+render_mode skip_vertex_transform;
+
+void vertex() {
+	VERTEX = (MODELVIEW_MATRIX * vec4(VERTEX, 1.0)).xyz;
+	NORMAL = normalize(VERTEX); // Re calculate vertex normals from origin
+	BINORMAL = normalize((MODELVIEW_MATRIX * vec4(BINORMAL, 0.0)).xyz);
+    TANGENT = normalize((MODELVIEW_MATRIX * vec4(TANGENT, 0.0)).xyz);
+}
```
Xtarsia, line 212:
```glsl
	NORMAL = normalize(VERTEX); // Re calculate vertex normals from model origin
	TANGENT = normalize(cross(NORMAL, vec3(0., 0., 1.)));
	BINORMAL = normalize(cross(NORMAL, TANGENT));
```

Shouldn't need skip_vertex_transform.

I would say this is really only a "have a quick look" suggestion, as it might lead to other issues with flipped tangents and other lighting errors etc. Setting up propper adjusted normals in blender or other software is really the course of action that should be taken here.

the articles linked are very informative too.

Xtarsia (commented):
Decent comprehensive tips section here!

TokisanGames:
Thanks for the help.

PR #722, opened 2025-06-16
Convert instancer MMIs from nodes to RS only

aidandavey:
This PR updates the instancer to use the RenderingServer directly instead of adding nodes. 

It should have the same functionality as the previous version, but works much faster (~2x) in the test I did. 

---
Admin edit:

Also 
* Override material is now only replaced when changing scene file / generated if it is one we generated (via set_meta())
* Incoming meshes are loaded into _pending_meshes, and only after the instancer has cleared, are they commited to _meshes.
* Adds a piece of #835, which is to replace show_instances with instancer_mode - disabled, which doesn't generate, or normal. Future options are changing visibility, placeholder materials/meshes

### DIFF
--- a/project/demo/assets/models/LOD10Example.tscn
+++ b/project/demo/assets/models/LOD10Example.tscn
@@ -87,50 +87,50 @@ bottom_radius = 0.575
 height = 1.185
 radial_segments = 4
 
-[node name="TestMultimesh" type="Node3D"]
+[node name="Lod10Example" type="Node3D"]
 
 [node name="Node3D" type="Node3D" parent="."]
 
-[node name="MeshInstance3D0" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD0" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("SphereMesh_u4ac2")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D1" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD1" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("TorusMesh_tjrnq")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D2" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD2" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("BoxMesh_xyuxq")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D3" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD3" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.118133, 0)
 mesh = SubResource("PrismMesh_xnm4h")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D4" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD4" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.246009, 0)
 mesh = SubResource("CylinderMesh_i1yhp")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D5" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD5" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("CylinderMesh_bvmel")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D6" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD6" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("CylinderMesh_x8oip")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D7" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD7" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.246009, 0)
 mesh = SubResource("CylinderMesh_dp7xj")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D8" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD8" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("CylinderMesh_8ctts")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D9" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD9" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.126, 0)
 mesh = SubResource("CylinderMesh_w2aj8")
 skeleton = NodePath("../..")
--- a/project/demo/assets/models/LOD5Example.tscn
+++ b/project/demo/assets/models/LOD5Example.tscn
@@ -36,32 +36,32 @@ radial_segments = 4
 cull_mode = 2
 albedo_color = Color(0.6205, 0.31, 1, 1)
 
-[node name="TestMultimesh" type="Node3D"]
+[node name="Lod5Example" type="Node3D"]
 
 [node name="Node3D" type="Node3D" parent="."]
 
-[node name="MeshInstance3D1" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD0" type="MeshInstance3D" parent="Node3D"]
 mesh = SubResource("SphereMesh_u4ac2")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D2" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD1" type="MeshInstance3D" parent="Node3D"]
 visible = false
 mesh = SubResource("BoxMesh_xyuxq")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D3" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD2" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.118133, 0)
 visible = false
 mesh = SubResource("PrismMesh_xnm4h")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D4" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD3" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.246009, 0)
 visible = false
 mesh = SubResource("CylinderMesh_dp7xj")
 skeleton = NodePath("../..")
 
-[node name="MeshInstance3D5" type="MeshInstance3D" parent="Node3D"]
+[node name="MeshInstance3D_LOD4" type="MeshInstance3D" parent="Node3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.246009, 0)
 visible = false
 mesh = SubResource("QuadMesh_264m5")
--- a/project/demo/data/assets.tres
+++ b/project/demo/data/assets.tres
@@ -1,6 +1,6 @@
 [gd_resource type="Terrain3DAssets" load_steps=11 format=3 uid="uid://dal3jhw6241qg"]
 
-[ext_resource type="PackedScene" uid="uid://bn5nf4esciwex" path="res://demo/assets/models/LOD4Example.tscn" id="1_4jrdu"]
+[ext_resource type="PackedScene" uid="uid://bn5nf4esciwex" path="res://demo/assets/models/LOD5Example.tscn" id="1_4jrdu"]
 [ext_resource type="Texture2D" uid="uid://c88j3oj0lf6om" path="res://demo/assets/textures/rock023_alb_ht.png" id="2_pog6b"]
 [ext_resource type="Texture2D" uid="uid://ddprscrpsofah" path="res://demo/assets/textures/ground037_alb_ht.png" id="3_g8f2m"]
 [ext_resource type="Texture2D" uid="uid://c307hdmos4gtm" path="res://demo/assets/textures/rock023_nrm_rgh.png" id="3_wncaf"]


aidandavey:
This PR needs to be updated to replicate the fix in https://github.com/TokisanGames/Terrain3D/pull/748

Saul2022:
Trying on mobile  on the demo and well  the mmis dissapear and only the last lod seems to appear for the sphere mesh. ..  Aside from this is didnt notice much framerate improbement though not sure 

https://github.com/user-attachments/assets/3e7f6176-8231-4747-b1d2-c31640c81ba7

aidandavey:
Thanks for testing, I'll investigate. Can you see console errors on mobile? If so, did you get any?

Saul2022:
> Thanks for testing, I'll investigate. Can you see console errors on mobile? If so, did you get any?

Sadly 0 errors while moving  the camera or adding more instance say it is with errors option or debug. A bit off topic though  at the first comment  you mean framerate improvement or something else?

aidandavey:
> > at the first comment  you mean framerate improvement or something else?

The 2x improvement is in the time it takes to create/update the mmis, which happens mainly while editing and when first loading the scene. 

RAM usage should be slightly lower both in editor and in game. Larger scenes and those with many mesh asset types/lods/cells in use will benefit more than simple scenes with few.

TokisanGames:
#788 has been merged. This can be rebased and looked at again. Also if we remove the MMI nodes, we remove the ability for users to access them in the tree to configure. So we need to expose their features such as visual instance layer, probably per meshAsset. Looking at the list I think that's the only thing.

aidandavey:
> #788 has been merged. This can be rebased and looked at again. Also if we remove the MMI nodes, we remove the ability for users to access them in the tree to configure. So we need to expose their features such as visual instance layer, probably per meshAsset. Looking at the list I think that's the only thing.

OK I've updated this and it's ready for review when you can get to it.

src/terrain_3d_mesh_asset.h
```diff
@@ -4,7 +4,7 @@
 #define TERRAIN3D_MESH_ASSET_CLASS_H
 
 #include <godot_cpp/classes/array_mesh.hpp>
-#include <godot_cpp/classes/geometry_instance3d.hpp>
+//#include <godot_cpp/classes/geometry_instance3d.hpp>
```
TokisanGames: Remove if not needed

src/terrain_3d_mesh_asset.h
```diff
@@ -41,6 +41,7 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
 	real_t _height_offset = 0.f;
 	real_t _density = 10.f;
 	ShadowCasting _cast_shadows = SHADOWS_ON;
+	int _visibility_layers = 1;
```
TokisanGames: Visibility layers should be a uint32_t, same with all the setters and getters.

src/terrain_3d_instancer.h
```diff
@@ -28,16 +28,12 @@ class Terrain3DInstancer : public Object {
 	// MM Resources stored in Terrain3DRegion::_instances as
 	// Region::_instances{mesh_id:int} -> cell{v2i} -> [ TypedArray<Transform3D>, PackedColorArray, modified:bool ]
 
-	// MMI Objects attached to tree, freed in destructor, stored as
-	// _mmi_nodes{region_loc} -> mesh{v2i(mesh_id,lod)} -> cell{v2i} -> MultiMeshInstance3D
-	using CellMMIDict = std::unordered_map<Vector2i, MultiMeshInstance3D *, Vector2iHash>;
+	// MMI Objects, freed in destructor, stored as
+	// _mmi_nodes{region_loc} -> mesh{v2i(mesh_id,lod)} -> cell{v2i} -> RID
+	using CellMMIDict = std::unordered_map<Vector2i, std::pair<RID, RID>, Vector2iHash>;
```
TokisanGames, line 40: The comment doesn't match the data structure. One says you get an RID. The other says you get an RID tuple. It's not clear what these RIDs are for, whereas before you knew you were getting a MultiMeshInstance3D. The first comment says I'll get an MMI Object (which I won't), and presumably they are no longer freed in the destructor (I haven't looked yet), but I don't know why I get two RIDs for an MMI.

src/terrain_3d_instancer.cpp
```diff
@@ -1,7 +1,7 @@
 // Copyright © 2025 Cory Petkovsek, Roope Palmroos, and Contributors.
 
 #include <godot_cpp/classes/resource_saver.hpp>
-
+#include <godot_cpp/classes/world3d.hpp>
```
TokisanGames, line 4: Add a space between godot headers and ours.

src/terrain_3d_instancer.cpp
```diff
@@ -1336,20 +1293,16 @@ void Terrain3DInstancer::update_mmis(const int p_mesh_id, const Vector2i &p_regi
...
 
 void Terrain3DInstancer::dump_mmis() {
-	LOG(WARN, "Dumping MMI tree and node containers");
-	LOG(MESG, "_mmi_containers size: ", int(_mmi_containers.size()));
-	for (auto &it : _mmi_containers) {
-		LOG(MESG, "_mmi_containers region: ", it.first, ", node ptr: ", ptr_to_str(it.second));
-	}
+	LOG(WARN, "Dumping MMI tree");
 	LOG(MESG, "_mmi tree: ");
 	_terrain->get_mmi_parent()->print_tree();
```
TokisanGames: If there are no MMI nodes, we don't need this whole function, nor do we need `Terrain3D::_mmi_parent`.

src/terrain_3d_instancer.cpp
```diff
@@ -441,37 +414,25 @@ void Terrain3DInstancer::_destroy_mmi_by_cell(const Vector2i &p_region_loc, cons
...
 		}
-		MultiMeshInstance3D *mmi = cell_mmi_dict[p_cell];
-		if (ma.is_valid() && mmi && mmi->get_multimesh().is_valid()) {
-			if (lod == 0) {
-				ma->update_instance_count(-mmi->get_multimesh()->get_instance_count());
+
+		RID &mmi = cell_mmi_dict[p_cell].first;
+		RID &mm = cell_mmi_dict[p_cell].second;
+		if (ma.is_valid() && mm.is_valid()) {
+			if (lod == ma->get_last_lod()) {
```
TokisanGames: I think what I had is more correct. I changed this to 0 in my last round of updates which fixed a lot of edge cases. If last_lod gets changed it screws up the count. But there will always be a lod0.

src/terrain_3d_instancer.cpp
```diff
@@ -441,37 +414,25 @@ void Terrain3DInstancer::_destroy_mmi_by_cell(const Vector2i &p_region_loc, cons
...
 			}
 		}
-		LOG(EXTREME, "Freeing ", ptr_to_str(mmi), " and erasing mmi cell ", p_cell);
-		remove_from_tree(mmi);
-		memdelete_safely(mmi);
+
+		LOG(EXTREME, "Freeing ", mmi, " and erasing mmi cell ", p_cell);
+		RS->free_rid(mmi);
+		RS->free_rid(mm);
+
```
TokisanGames: Drop the last blank line since the log refers to the cell.

src/terrain_3d_instancer.cpp
```diff
@@ -441,37 +414,25 @@ void Terrain3DInstancer::_destroy_mmi_by_cell(const Vector2i &p_region_loc, cons
...
-			Node *node = _mmi_containers[p_region_loc];
-			if (node && node->get_child_count() == 0) {
-				LOG(EXTREME, "Removing ", node->get_name());
-				_mmi_containers.erase(p_region_loc);
-				remove_from_tree(node);
-				memdelete_safely(node);
-			}
-		}
-		// This invalidates mesh_mmi_dict here and for calling functions
-		_mmi_nodes.erase(p_region_loc);
```
TokisanGames, line 473: _mmi_nodes still exists and the region needs to be removed from it.

src/terrain_3d_instancer.cpp
```diff
@@ -111,15 +111,8 @@ void Terrain3DInstancer::_update_mmi_by_region(const Terrain3DRegion *p_region,
...
-	String rname("Region" + Util::location_to_string(region_loc));
-	if (_mmi_containers.count(region_loc) == 0) {
-		LOG(DEBUG, "Creating new region MMI container Terrain3D/MMI/", rname);
-		Node3D *node = memnew(Node3D);
-		node->set_name(rname);
-		_mmi_containers[region_loc] = node;
-		_terrain->get_mmi_parent()->add_child(node, true);
-	}
+	// Get or create mesh dict
+	MeshMMIDict &mesh_mmi_dict = _mmi_nodes[region_loc];
```
TokisanGames: This will cause crashes. It is already defined below with a note about it.

src/terrain_3d_instancer.cpp
```diff
@@ -142,6 +135,7 @@ void Terrain3DInstancer::_update_mmi_by_region(const Terrain3DRegion *p_region,
 	// Process cells
 	Dictionary cell_inst_dict = mesh_inst_dict[p_mesh_id];
 	Array cell_locations = cell_inst_dict.keys();
+	RID shadow_impostor_source_mm;
```
TokisanGames: Since this isn't used until very far down in the loop, it's possible when it is used it could have a value from a previous cell loop iteration. If defined where it was, it gets reset every cell.

src/terrain_3d_instancer.cpp
```diff
@@ -235,40 +213,37 @@ void Terrain3DInstancer::_update_mmi_by_region(const Terrain3DRegion *p_region,
...
-			if (modified || mmi->get_multimesh().is_null() ||
+			if (modified || !mm.is_valid() ||
 					lod == Terrain3DMeshAsset::SHADOW_LOD_ID) {
-				// Subtract previous instance count of this MMI
-				if (lod == 0) {
-					if (mmi->get_multimesh().is_valid()) {
-						ma->update_instance_count(-mmi->get_multimesh()->get_instance_count());
-					}
+				// Subtract previous instance count for this cell
+				if (mm.is_valid() && lod == ma->get_last_lod()) {
```
TokisanGames: There's always a lod0. Last_lod is adjustable and can screws with the instance count.

src/terrain_3d_instancer.cpp
```diff
@@ -277,40 +252,43 @@ void Terrain3DInstancer::_update_mmi_by_region(const Terrain3DRegion *p_region,
...
 			Vector2i mesh_key(p_mesh_id, Terrain3DMeshAsset::SHADOW_LOD_ID);
 			CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
-			MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
-			if (mmi) {
-				mmi->set_custom_aabb(mmi_custom_aabb);
+			//RID &mmi = cell_mmi_dict[cell].first;
+			RID &mm = cell_mmi_dict[cell].second;
+			if (mm.is_valid()) {
+				RS->multimesh_set_custom_aabb(mm, mm_custom_aabb);
 			}
```
TokisanGames, line 296: Why don't you set the mmi aabb like you do above, and as removed?

src/terrain_3d_instancer.cpp
aidandavey: I'd like to rename _mmi_nodes, what do you think to _region_mmis ?

TokisanGames: MMIs are by cells, not regions. This data structure used to contain nodes. Now it contains rids. So I think _mmi_rids would be appropriate.

src/terrain_3d_instancer.h
aidandavey: For reference, the pair is an MMI Rid and it's MM RID. Both are freed in the destructor. 
I couldn't find a way to get the MultiMesh RID back from the Instance using the rendering server, so just store both RIDs together.  I have clarified the comment. 

```
	// A pair of MMI and MM RIDs, freed in destructor, stored as
	// _region_mmis{region_loc} -> mesh{v2i(mesh_id,lod)} -> cell{v2i} -> std::pair<mmi.RID, mm.RID>

	using CellMMIDict = std::unordered_map<Vector2i, std::pair<RID, RID>, Vector2iHash>;
	using MeshMMIDict = std::unordered_map<Vector2i, CellMMIDict, Vector2iHash>;
	std::unordered_map<Vector2i, MeshMMIDict, Vector2iHash> _region_mmis;
```

aidandavey:
I've updated following the review. 
I found I'm getting an error when clearing instances (middle mouse in asset dock)
```
ERROR: Parameter "mesh" is null.
   at: mesh_get_surface_count (servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp:667)
```

aidandavey:
Also need to test the new 'show instances' implementation

TokisanGames:
Be sure to test instancer with:
* Changing Vertex spacing
* Changing region size
* Changing MMI shadow casting, lod_count, shadow lod
* Changing between two open scenes w/ instances

> I found I'm getting an error when clearing instances (middle mouse in asset dock)
> ERROR: Parameter "mesh" is null.
>    at: mesh_get_surface_count (servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp:667)

I've not seen this; you're only seeing it after this PR?
Does it occur many times?
You are freeing the mmi RID before the MM RID. Perhaps that should be reversed?
Perhaps you are clearing the mesh resource before removing the mesh/mm/mmi RIDs from the Rendering Server.
You can step through the C++ code in a debugger and see when the console issues the error.

aidandavey:
> > I found I'm getting an error when clearing instances (middle mouse in asset dock)
> > ERROR: Parameter "mesh" is null.
> >    at: mesh_get_surface_count (servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp:667)
> 
> I've not seen this; you're only seeing it after this PR?

Yes it's just here in this PR.

> Does it occur many times?

Just a few times but it varies, a guess is one error per cell used * nb lods. 

> You are freeing the mmi RID before the MM RID. Perhaps that should be reversed?
> Perhaps you are clearing the mesh resource before removing the mesh/mm/mmi RIDs from the Rendering Server.
> You can step through the C++ code in a debugger and see when the console issues the error.
> 

Good ideas, I'll start here, thanks!

aidandavey:
> 
> > You are freeing the mmi RID before the MM RID. Perhaps that should be reversed?

This is correct, should free dependents first 

> > Perhaps you are clearing the mesh resource before removing the mesh/mm/mmi RIDs from the Rendering Server.

This was it, destruction order must be MMI, MM, MA to avoid the error

aidandavey:
Ok I'm happy with this and I tested it quite thoroughly including:

Changing Vertex spacing 
Changing region size (destroys instances - same behavior as main)
Changing MMI shadow casting, lod_count, shadow lod
Changing between two open scenes w/ instances

One minor difference in behavior is that when show_instances is false, the instance counter label reads as 0. This is because they have actually been destroyed and will be recounted when they are recreated.

src/terrain_3d.h
```diff
@@ -188,8 +186,8 @@ class Terrain3D : public Node3D {
 	real_t get_cull_margin() const { return _cull_margin; };
 	void set_free_editor_textures(const bool p_free_textures) { _free_editor_textures = p_free_textures; }
 	bool get_free_editor_textures() const { return _free_editor_textures; };
-	void set_show_instances(const bool p_visible) { _mmi_parent ? _mmi_parent->set_visible(p_visible) : void(); }
-	bool get_show_instances() const { return _mmi_parent ? _mmi_parent->is_visible() : false; }
+	void set_show_instances(const bool p_visible) { _instancer ? _instancer->set_show_instances(p_visible) : void(); }
+	bool get_show_instances() const { return _instancer ? _instancer->get_show_instances() : true; }
```
TokisanGames: This used to default to false. If the instancer doesn't exist, which will never happen unless we make the system more modular, we can be sure that instances are not showing.

aidandavey:
> * Upon clearing a scene file with instances on the ground, eg LodExample, takes quite a while before it finally shows the generated card. The console is dumping this error:
>   `ERROR: Parameter "mesh" is null.
>    at: mesh_get_surface_count_and_materials (./servers/rendering/renderer_rd/storage_rd/mesh_storage.h:405)`
>   Then later it reports:
>   `ERROR: 161826 RID allocations of type 'N10RendererRD11MeshStorage9MultiMeshE' were leaked at exit.`
> 
>   It's probably clearing the mesh RID before destroying the MMIs. That might have worked when it was nodes but now doesn't like it.
> 

Yes you are right, this destroys the meshes  immediately without ever destroying the mmi or mm. I have a fix for this.

> 
> 
> * > Changing region size (destroys instances - same behavior as main)
> 
>   Hmm, that shouldn't happen. It used to work fine. I tested all the way back to the start of 1.1 and it behaves the same. Reported in #855.
>

> * Sometimes after painting instances, changing the asset dock, then sculpting, it won't update the height of instances in every region. Also present on main. This is another bug that needs to be looked at in #856.
> 

I'll have a look at these

> * Can be squashed and rebased.
> 
> * I'm going to try and get #825 merged in before this. So you could rebase it off of that branch. To `Terrain3DMeshAsset::set_visibility_layers` you can add `SET_IF_DIFF(_visibility_layers, p_layers);` at the top and remove the if statement, and the assignment. Same with `Terrain3DInstancer::set_show_instances` `SET_IF_DIFF(_show_instances, p_visible);` and no conditional or assignment.

No problem

TokisanGames:

src/terrain_3d_mesh_asset.cpp
```diff
@@ -368,6 +369,13 @@ ShadowCasting Terrain3DMeshAsset::get_lod_cast_shadows(const int p_lod_id) const
 	return _cast_shadows;
 }
 
+inline void Terrain3DMeshAsset::set_visibility_layers(const int p_layers) {
+	LOG(INFO, _name, ": Setting visibility layers: ", p_layers);
+	_visibility_layers = p_layers;
+	LOG(DEBUG, "Emitting instancer_setting_changed, ID: ", _id);
+	emit_signal("instancer_setting_changed", _id);
```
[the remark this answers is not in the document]
TokisanGames: New form:
```
inline void Terrain3DMeshAsset::set_visibility_layers(const uint32_t p_layers) {
	SET_IF_DIFF(_visibility_layers, p_layers);
	LOG(INFO, _name, ": Setting visibility layers: ", p_layers);
	LOG(DEBUG, "Emitting instancer_setting_changed, ID: ", _id);
	emit_signal("instancer_setting_changed", _id);
}
```

src/terrain_3d.h
TokisanGames: I was wrong, it needs to be true to set the default, when Terrain3D is first instantiated (Godot instantiates once for defaults, then another time for real).

TokisanGames:
Nice work, very solid.

I rebased and adjusted for the for loops commit.

I can get an error under certain circumstances.
- Restart Godot.
- Drag Lod4Example to an existing texture card mesh asset in the asset dock.

```
ERROR: Parameter "mesh" is null.
   at: mesh_get_surface_count_and_materials (./servers/rendering/renderer_rd/storage_rd/mesh_storage.h:405)
```

It doesn't seem to have an issue after that.

If instead I restart, then edit the mesh asset, and drag the new packedscene to the scene file slot, it doesn't have an issue.

aidandavey:
> Nice work, very solid.
> 
> I can get an error under certain circumstances.
> - Restart Godot.
> - Drag Lod4Example to an existing texture card mesh asset in the asset dock.
> 
> ```
> ERROR: Parameter "mesh" is null.
>    at: mesh_get_surface_count_and_materials (./servers/rendering/renderer_rd/storage_rd/mesh_storage.h:405)
> ```
> 
> It doesn't seem to have an issue after that.
> 
> If instead I restart, then edit the mesh asset, and drag the new packedscene to the scene file slot, it doesn't have an issue.
> 
> 

Thanks!

I'll have a look at this again. My wild guess is that this error is coming from the thumbnail generation, since we are hovering the entry after dropping. Because we deferred the mesh parsing there is no mesh to put in the tmp scenario at that stage.

TokisanGames:
Good idea. The mouse would be right there triggering a thumbnail update.

aidandavey:
> Good idea. The mouse would be right there triggering a thumbnail update.

Turns out the issue was that dropping the data was creating a new MA, indirectly destroying the old meshes.

I pushed a change which will only create a new MA if the slot is empty, otherwise reuses the one referenced in the slot.

TokisanGames:
I discovered memory leaks after removing or updating transforms, and fixed it. This one is good to go. Nice work.

Later, we could look at these, added to #43 
* Reusing RIDs
* Pre-Allocating MultiMeshes w/ maximum amount of memory
* Dynamically "create and destroy" (enable/disable/reuse preallocated buffers) MMI/MMs as the camera moves
* Updating the RS Buffer more directly with RS->multimesh_set_buffer(rid, buffer)
* Use RS->multimesh_set_visible_instances(rid, count) to limit the instances visible and the amount of the buffer used if using a preallocated maximum. Empty cells can be 0

TokisanGames:
Alright this version appears to have resolved the new issues.

* Customized assets load all settings. Every setting in the two mesh assets that can be changed, has been changed, so everything should be non-default.
* Both should be paintable on terrain (one is marked shadow only)
* After painting, both assets should allow replacement via the asset dock drag, or replacing the scene file
* Thumbnails generate, and update

New changes:
* Override material is now only replaced when changing scene file / generated if it is one we generated (via set_meta())
* Incoming meshes are loaded into _pending_meshes, and only after the instancer has cleared, are they commited to _meshes.
* Adds a piece of #835, which is to replace show_instances with instancer_mode - disabled, which doesn't generate, or normal. Future options are changing visibility, placeholder materials/meshes


Pending:
* cleanup, testing

TokisanGames:
* Fixed Instance count alternated between 0 and # when adding, but not removing. Caused by an old node based block in append_regions. Removing that properly fixed this.
* Redesigned MeshAsset initialization, including reverting some of #859 as unnecessary
* Fixed shadows only asset not changing the instance counter.

Pending:
* Still needs a bit more testing, cleanup of messages and another review

TokisanGames:
Ok, found, fixed, and redesigned some things. I think it finally addresses all issues.
Thanks for the effort on this.

PR #724, opened 2025-06-19
Apply MeshInstance3D transformation to Terrain3DMeshAsset vertex data

aidandavey:
Addresses an ambition of #43, to apply mesh transforms for the foliage instancer. 

The vertex data is extracted from the original mesh and transformed using the MeshInstance3D transformation. 

Then we create an ArrayMesh from this data, apply the surface materials and store the ArrayMesh for use by the instancer.

### DIFF
--- a/project/demo/Demo.tscn
+++ b/project/demo/Demo.tscn
@@ -34,11 +34,11 @@ noise = SubResource("FastNoiseLite_d8lcj")
 _shader_parameters = {
 "_mouse_layer": 2147483648,
 "auto_base_texture": 0,
-"auto_height_reduction": 0.0,
+"auto_height_reduction": 0.05,
 "auto_overlay_texture": 1,
-"auto_slope": 1.0,
+"auto_slope": 0.45,
 "bias_distance": 512.0,
-"blend_sharpness": 0.87,
+"blend_sharpness": 0.5,
 "depth_blur": 0.0,
 "dual_scale_far": 170.0,
 "dual_scale_near": 100.0,
@@ -46,7 +46,7 @@ _shader_parameters = {
 "dual_scale_texture": 0,
 "enable_macro_variation": true,
 "enable_projection": true,
-"height_blending": true,
+&"flat_terrain_normals": false,
 "macro_variation1": Color(0.855, 0.8625, 0.9, 1),
 "macro_variation2": Color(0.9, 0.885, 0.81, 1),
 "macro_variation_slope": 0.333,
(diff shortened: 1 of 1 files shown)


TokisanGames:
On a cursory glance, it looks like you're modifying the mesh to adjust the origin point, right?

If you memnew something, then we need to memdelete it. What you've done here creates memory leaks. _meshes.clear() won't release this memory.

I appreciate the effort, but I was thinking this could be done much more simply. Why not just retain the MeshInstance3D transform, and apply it every time we place the mesh as we do with the temporary transforms on the tool settings bar?

aidandavey:
That's right and I should have probably explained my reasoning. 

I used the method you described in my instance_collision PR and we could do the same here no problem.

I figured it might be cheaper, since we apply the transform just once and then dont need to do casting from the typed arrays and so on in the instancer. The array just gets passed directly to the MM as it is now. 

Applying the transform on the vertex data could also make it possible to combine meshes from multiple MeshInstance3Ds, each with their own transforms, since we can parse the surfaces for each and add them to the ArrayMesh.

For example:

LOD_0 -> trunk -> leaves
LOD_1 -> trunk -> leaves
LOD_2 -> Billboard

Memory leak noted! Will fix or remove depending on how you want to proceed.

TokisanGames:
> Applying the transform on the vertex data could also make it possible to combine meshes from multiple MeshInstance3Ds, 

You mean allowing a user to give us a scene with separate Trunk and Leaf Mesh instances?

I don't want to encourage sloppy behavior of users. Users should fix their assets in blender and import them properly combined into one object with separate LODs, multiple materials, and applied transforms with the origin point at the base of the mesh and at 0,0,0. If they have poorly constructed objects, they're going to have problems with their assets in more places than our instancer.

I want to build in code that is convenient for tedious tasks, but not that fixes user slop.

Supporting transforms in the scene is a minor adjustment, so acceptable. Perhaps they have a tree that they want to sink lower or rotate a bit and it's easy change to do in the scene rather than reimport. 

Applying the scene transform occurs once during painting, then it's set. There's no on-going cost. Unlike with this PR, you need to regenerate the ArrayMesh for every asset (we have 60-80 so far), every scene load in the editor and game. Plus there's the minor cost of mesh duplication.

Combining meshes would be more interesting to me where we have bunch of rocks piled into a hill, or wooden boards building a house. I would like to bake those into an object. Or combine a bunch of collision shapes into one. 

So, save the code and we can use it in OOTA. But I think we don't need to do it here for these transforms.

Also, I hate to waste your effort. Not all items in the Issues list needs to be or should be implemented. Before doing major work it's best to run it by us in discord. Can also be helpful if looking for something to work on as some ideas may not have even made it to the Issues yet (eg converting between instances and instanced scenes).

aidandavey:
I see, no worries! I'll use your suggestion instead.

aidandavey:
Oh and I'd be happy to look at instanced scenes. I'll discuss it first though

aidandavey:
> Applying the scene transform occurs once during painting, then it's set.

I think this won't work if the user wants to transform one LOD mesh but not the others, for example. I think it might be better to do it in the instancer while creating the MMI if we still want it.

TokisanGames:
Hmm, offsetting different LODS? That gets back to slop. This whole feature is to help them with a little bit of slop, but not too much. If their asset is so bad that their lods are offset from each other, then they need to rework it in blender. Let's just retain the global transform of the LOD0 mesh instance and apply it to all instances, all lods. If that's not good enough for them, they need to fix their asset. That's more than enough accommodation on our end.

PR #726, opened 2025-06-19
Apply transform of first LOD to MeshAsset Instances

aidandavey:
As discussed on previous PR #724, applying transforms from the first LOD mesh in the MeshAsset file while painting instances.

### DIFF
--- a/src/terrain_3d_instancer.cpp
+++ b/src/terrain_3d_instancer.cpp
@@ -540,7 +540,8 @@ void Terrain3DInstancer::add_instances(const Vector3 &p_global_position, const D
 	TypedArray<Transform3D> xforms;
 	PackedColorArray colors;
 	for (int i = 0; i < count; i++) {
-		Transform3D t;
+		// Start with the mesh transform to correct the mesh to its visual state
+		Transform3D t = mesh_asset->get_mesh_transform();
 
 		// Get random XZ position and height in a circle
 		real_t r_radius = radius * sqrt(UtilityFunctions::randf());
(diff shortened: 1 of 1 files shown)


TokisanGames:
Thanks for the start on this. I've rebased it and started to review it, thinking about what we should be doing with this corrective feature.


### Background

The LODs in the example are setup how they often will be coming in from an artist with a bunch of LODs underneath a Blender `Empty` node. The root node might be setup as a static body. Below that will often be a parent node for all of the meshes, either a `Node3D` as here, or a Blender `Armature` or `Skeleton3D` node. Or the mesh lods might just be dumped under the root. Then a collision node will be under the root.
{BA59D943-2519-4F54-B301-611B61F52CB1}

I don't want to encourage poor behavior by accommodating and fixing every artist created problems, especially if it's a lot of code for us. I do want to facilitate convenience, so if it's a minor thing for great convenience, then fine.

So now the question is, where in this tree might there be transforms, and why? 
1. LOD misalignment where LOD0, LOD1, LOD2+ have different transforms. This is a messy asset that we should not accommodate by handling every LOD transform. This needs to be fixed in blender by the artist.
2. Import/export issues. Sometimes transferring assets between UE/Unity/Blender/DCC have different coordinate systems so the incoming asset is 90 degrees off on one axis. Sometimes transforms haven't been applied properly, or are baked into the animations where the root/mesh parent node is scaled to 0.01 but the meshes and baked animations are scaled to 100 (though we won't have animations on instanced nodes), and they can't be fixed in the DCC. These are very annoying, and easier to fix in the scene with a rotation/scale on the root/mesh parent. We can accommodate this.
3. The origin / scale is off on the mesh parent/root node. It's an artist issue, that should be fixed in Blender, but it can be provided for free by fixing the above issue.

### Review

In this PR, you've captured the LOD0 local transform. But we should capture the root / mesh parent transform, or get the LOD0 global transform (not available), or all of the transforms between LOD0 and the scene root.

I swapped example LOD0/1. Then transformed the mesh parent to affect all LODs simultaneously to rotate it as follows. However, placing this on the terrain it still appears flat because there are no transforms on LOD0:

{13A3F30B-958F-4164-A349-2C4C528FFDAE}

Accumulating the transforms from LOD0 to scene root allows this to work:

{D1722D35-D31F-402D-8573-0C3396EDEED0}

-----
Because the mesh asset transform is intended to be corrective (moving a transform to 0 or resetting rotation to neutral), I also think the mesh asset transform needs to be applied before the instancer applied transforms which are for variation. This is a bit more complicated, so I'll address it in another commit and comment.

aidandavey:
Thinking on this, baking the LOD transform into the instance data might not be what we want to do. It doesn't allow for modification afterwards, and causes an issue with my instance collision PR since I need the shape positions, not the LOD.

In fact baking the variation transforms might already be an issue there.

I'm thinking the instancer data could hold the instance_origin, and a variation transform.

The LOD transform, and CollisionShape transform, will remain in the MA data and can be applied in update_mmis and/or update_instance_collision, in the proper order region->instance->correction->variation

This also sets us up to support an instance transform brush, so you can paint alternative variation.

Issue #691, opened 2025-05-08
Billboard visibility gaps


TokisanGames:
### Description

MMI visibility ranges have a gap if the edges of ranges are perfectly aligned. They must be overlapped. We expand the far range of the previous lod to overlap and close the gap in terrain_3d_instancer.cpp. The gap gets wider the further out it is, so we multiplied the range by 1.0005 which covers most cases.

When planes are used as lods, their visibility gap is much further than with 3D meshes. It's likely due to aabbs, but setting a custom reportedly doesn't work. Extending the range further to say 0.3 for 128 (1.0024). 

* Follow the fix in #639
* On the 2nd to last lod, we can assume it's a billboard, even though it may not be. Apply 1.0024 instead of 1.0005f. However this may be a bit much so test for just over the minimum value at close and far ranges of several thousand meters.

aidandavey:
I don't see any gaps in v4.4.stable.official [4c311cbee]

In fact I get a visible overlap at far distances (~2000m)

![Image](https://github.com/user-attachments/assets/bbc0f2d0-edd8-4a42-9df4-b7fa3d82a1ec)

For me it seems better without the  * 1.0005f increase.

TokisanGames:
We want a visible overlap at all distances, since MMI fading doesn't work, and neither do perfect ranges. We want to ensure no gaps, even if there is a greater overlap. There were a couple people experiencing a gap with billboards on the last lod.

https://discord.com/channels/691957978680786944/1130291534802202735/1378112906427760741

https://discord.com/channels/691957978680786944/1130291534802202735/1369825896797569195

aidandavey:
Unfortunately these links are not working for me. 

Can anyone submit a MRP demonstrating the gap?

aidandavey:
Got it, I can apply the suggested fix if it helps but I can't verify if it fixed the issue, or recommend a more appropriate multiplier.

TokisanGames:
I'm also not able to reproduce it with the opaque or transparent pipelines. But two users had it and the given fix addressed it for one, and shouldn't impact anyone else so we'll go with it.

PR #728, opened 2025-06-19
Fix penultimate LOD Billboard visibility gaps addressing #691

aidandavey:
Fix penultimate LOD Billboard visibility gaps addressing #691

> MMI visibility ranges have a gap if the edges of ranges are perfectly aligned. They must be overlapped. We expand the far range of the previous lod to overlap and close the gap in terrain_3d_instancer.cpp. The gap gets wider the further out it is, so we multiplied the range by 1.0005 which covers most cases.
> 
> When planes are used as lods, their visibility gap is much further than with 3D meshes. It's likely due to aabbs, but setting a custom reportedly doesn't work. Extending the range further to say 0.3 for 128 (1.0024).


TokisanGames:
Great, thank you.

PR #732, opened 2025-06-23
Add instance on all collision

aidandavey:
This PR allows you to paint foliage instances on collision shapes. 

Added a checkbox to instancer tool settings to toggle paint on all collision 
Add physics raycast intersection functions
If checkbox is toggled on, check for physics intersection at the position and instance at that height. We check from 100m to - 100m in y only.
If physics intersection fails, we fall back to using _terrain->get_height() as normal.
Works with align to normal

As discussed, if a collision shape is moved after painting then we will have floating instances, which may not be what was intended! In this case the smooth tool can be used at low strength to snap the instances back to the terrain.

---
Partially implements foliage tracker #43

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -16,6 +16,8 @@
 #include <godot_cpp/classes/viewport.hpp>
 #include <godot_cpp/classes/viewport_texture.hpp>
 #include <godot_cpp/classes/world3d.hpp>
+#include <godot_cpp/classes/physics_direct_space_state3d.hpp>
+#include <godot_cpp/classes/physics_ray_query_parameters3d.hpp>
 
 #include "logger.h"
 #include "terrain_3d.h"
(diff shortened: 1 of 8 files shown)


aidandavey:
This could be improved to allow painting on the sides and underside of collision shapes. So painting hanging mosses or something like that. 
Currently it projects the ray downwards, we could instead project from the camera/mouse position.

TokisanGames:
Downward raycasts is sufficient. Things like hanging ivy require special meshes, which can then be manually placed with level design tools.

src/terrain_3d.cpp
```diff
@@ -16,6 +16,8 @@
 #include <godot_cpp/classes/viewport.hpp>
 #include <godot_cpp/classes/viewport_texture.hpp>
 #include <godot_cpp/classes/world3d.hpp>
+#include <godot_cpp/classes/physics_direct_space_state3d.hpp>
+#include <godot_cpp/classes/physics_ray_query_parameters3d.hpp>
```
TokisanGames: These should have been alphabetized above V and W automatically by clang-format. Also there was some extra white space it would have caught. Did you install the hooks in your repo per the contributor doc?

TokisanGames (approved):
Nice work, it will be very useful. Please check out my changes for edification.

src/terrain_3d.cpp
aidandavey: I think I missed this on my laptop. I'll fix it and check my other work.

PR #733, opened 2025-06-23
Region undo bugs

### LINKED ISSUE #731: Remove Region Undo Broken
### Description

Undoing removal of a region is broken. Redo/undo reports two errors:
```
ERROR: Terrain3DData#2275:set_region_deleted:202: Region not found at: (0, -2)
   at: push_error (core/variant/variant_utility.cpp:1098)
ERROR: Terrain3DData#2275:set_region_modified:184: Region not found at: (0, -2)
   at: push_error (core/variant/variant_utility.cpp:1098)

```

TokisanGames:
Changes:

* Remove region, undo resulted in errors and didn't work. Caused by a regression in 3e98700f13dca5f2a98e98452a6328cbad24d2d5. get_region() returns any region regardless of modified or deleted. get_region_ptr() only looked at active regions. Now it looks at all regions.

* Adding a new blank region did not get placed in the redo set, so when going through the sequences below it wouldn't replace the blank region and instead used the old region and unmarked it for delete.

    * Remove region, replace it with a blank, undo, redo -> Correctly have the new blank region in place.
    * Remove region, replace it with a blank, undo twice (original state), redo twice -> Should have the new blank region, but instead shows the original region. The blank region has disappeared. **(fixed)**
    * Sculpt 1, 2, 3, 4 and undo/redo -> Works fine going forward and backwards in history repeatedly.

* Edited regions were retaining a deleted flag in some cases and needed to be marked undeleted

Fixes #731

### DIFF
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -76,9 +76,9 @@ TypedArray<Terrain3DRegion> Terrain3DData::get_regions_active(const bool p_copy,
 	TypedArray<Terrain3DRegion> region_arr;
 	for (int i = 0; i < _region_locations.size(); i++) {
 		Vector2i region_loc = _region_locations[i];
-		Terrain3DRegion *region = get_region_ptr(region_loc);
-		if (region) {
-			region_arr.push_back((p_copy) ? region->duplicate(p_deep) : region);
+		Ref<Terrain3DRegion> region = get_region(region_loc);
+		if (region.is_valid()) {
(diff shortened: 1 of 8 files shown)


TokisanGames:
> add a region > sculpt > sculpt > subtract that region.

Going back and forth over the undo history, the 2nd sculpt disappears on like the 2nd round.

TokisanGames:
This PR seems pretty solid to me, testing various painting, sculpting, and instancer tools.

PR #734, opened 2025-06-23
Add mesh asset picking support to instancer tool

### LINKED ISSUE #775: Texture Picker
### Description

We need a picker to find and select textures.

aidandavey:
Implements the ability to pick mesh assets directly from the terrain using the instancer tool. Adds a mesh asset picker setting, updates the UI to show it for the instancer tool, and introduces a method to find the closest mesh asset at a given position.

Currently only finds the closest instance in the cell you picked. Could be expanded to check neighbouring cells.

---
Partial implementation #43 
Fixes #775

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -272,7 +272,7 @@ func update_thumbnails() -> void:
 			mesh_asset.queue_redraw()
 
 
-func set_selected_by_resource_id(p_id: int) -> void:
+func set_selected_by_asset_id(p_id: int) -> void:
 	search_box.text = ""
 	_on_search_text_changed()
 	current_list.set_selected_id(p_id)
@@ -393,6 +393,10 @@ func update_assets() -> void:
 func remove_all_highlights():
 	if not plugin.terrain:
 		return
+	for i: int in texture_list.entries.size():
+		var resource: Terrain3DTextureAsset = texture_list.entries[i].resource
+		if resource and resource.is_highlighted():
+			resource.set_highlighted(false)
 	for i: int in mesh_list.entries.size():
 		var resource: Terrain3DMeshAsset = mesh_list.entries[i].resource
 		if resource and resource.is_highlighted():
@@ -634,12 +638,12 @@ class ListContainer extends Container:
 		plugin.ui._on_setting_changed()
 
 
-	func get_selected_resource_id() -> int:
+	func get_selected_asset_id() -> int:
 		# "Add new" is the final entry only when search box is blank
 		var max_id: int = max(0, entries.size() - (1 if search_text else 2))
 		var id: int = clamp(selected_id, 0, max_id)
 		if plugin.debug:
-			print("Terrain3DListContainer: get_selected_resource_id: selected_id: ", selected_id, ", clamped: ", id, ", entries: ", entries.size())
+			print("Terrain3DListContainer: get_selected_asset_id: selected_id: ", selected_id, ", clamped: ", id, ", entries: ", entries.size())
 		if id >= entries.size():
 			return 0
 		var res: Resource = entries[id].resource
@@ -780,7 +784,7 @@ class ListEntry extends MarginContainer:
 		add_theme_constant_override("margin_left", 5)
 		add_theme_constant_override("margin_right", 5)
 
-		if resource is Terrain3DMeshAsset:
+		if resource:
 			is_highlighted = resource.is_highlighted()
 
 		setup_buttons()
@@ -809,19 +813,19 @@ class ListEntry extends MarginContainer:
 			button_enabled.toggle_mode = true
 			button_enabled.mouse_filter = Control.MOUSE_FILTER_PASS
 			button_enabled.mouse_default_cursor_shape = Control.CURSOR_POINTING_HAND
-			button_enabled.pressed.connect(enable)
+			button_enabled.pressed.connect(_on_enable)
 			button_row.add_child(button_enabled, true)
 			
 		button_highlight.set_texture_normal(highlight_icon)
 		button_highlight.set_custom_minimum_size(icon_size)
 		button_highlight.set_h_size_flags(Control.SIZE_SHRINK_END)
 		button_highlight.set_visible(resource != null)
-		button_highlight.tooltip_text = "Highlight Instances"
+		button_highlight.tooltip_text = "Highlight " + ( "Instances" if type == Terrain3DAssets.TYPE_MESH else "Texture" )
 		button_highlight.toggle_mode = true
 		button_highlight.mouse_filter = Control.MOUSE_FILTER_PASS
 		button_highlight.mouse_default_cursor_shape = Control.CURSOR_POINTING_HAND
 		button_highlight.set_pressed_no_signal(is_highlighted)
-		button_highlight.pressed.connect(highlight)
+		button_highlight.pressed.connect(_on_highlight)
 		button_row.add_child(button_highlight, true)
 		
 		button_edit.set_texture_normal(edit_icon)
@@ -831,7 +835,7 @@ class ListEntry extends MarginContainer:
 		button_edit.tooltip_text = "Edit Asset"
 		button_edit.mouse_filter = Control.MOUSE_FILTER_PASS
 		button_edit.mouse_default_cursor_shape = Control.CURSOR_POINTING_HAND
-		button_edit.pressed.connect(edit)
+		button_edit.pressed.connect(_on_edit)
 		button_row.add_child(button_edit, true)
 
 		spacer.size_flags_horizontal = Control.SIZE_EXPAND_FILL
@@ -845,7 +849,7 @@ class ListEntry extends MarginContainer:
 		button_clear.tooltip_text = "Clear Asset"
 		button_clear.mouse_filter = Control.MOUSE_FILTER_PASS
 		button_clear.mouse_default_cursor_shape = Control.CURSOR_POINTING_HAND
-		button_clear.pressed.connect(clear)
+		button_clear.pressed.connect(_on_clear)
 		button_row.add_child(button_clear, true)
 
 
@@ -907,7 +911,7 @@ class ListEntry extends MarginContainer:
 		if not mesh_resource:
 			count_label.text = str(0)
 		else:
-			count_label.text = format_number(mesh_resource.get_instance_count())
+			count_label.text = _format_number(mesh_resource.get_instance_count())
 
 
 	func _notification(p_what) -> void:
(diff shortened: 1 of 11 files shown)


Xtarsia:
Tested quickly and works well even with thousands of instances inside a cell.

As for searching across cells, using a small AABB centered on the picker position, and using the code from the update / remove instances function could work.

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+		return -1; // No region found
+	}
+
+	Vector3 region_global_pos = v2v3(region_loc) * _terrain->get_region_size() * _terrain->get_vertex_spacing();
+	Vector2i cell = _get_cell(p_picker_location, region->get_region_size());
+
+	Dictionary mesh_inst_dict = region->get_instances();
+
+	if (mesh_inst_dict.is_empty()) {
+		LOG(WARN, "No meshes found in region: ", region->get_location());
```
TokisanGames: I don't think we need to warn if they pick a blank area; just silent fail is fine.

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+	if (!region) {
+		LOG(WARN, "No region found at location: ", p_picker_location);
+		return -1; // No region found
+	}
+
+	Vector3 region_global_pos = v2v3(region_loc) * _terrain->get_region_size() * _terrain->get_vertex_spacing();
+	Vector2i cell = _get_cell(p_picker_location, region->get_region_size());
+
+	Dictionary mesh_inst_dict = region->get_instances();
+
```
TokisanGames: trim whitespace

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+		LOG(WARN, "No meshes found in region: ", region->get_location());
+		return -1; // No meshes found
+	}
+
+	if (mesh_inst_dict.size() == 1) {
+		// Only one mesh type in this region, return it
+		return mesh_inst_dict.keys()[0];
+	}
+
+	// Multiple meshes, find the closest one
```
TokisanGames: //Iterate through all mesh types looking for meshes in cell to find the closest transform

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+		Array triple = cell_inst_dict[cell];
+		TypedArray<Transform3D> xforms = triple[0];
+		if (xforms.is_empty()) {
+			continue; // No transforms in this cell
+		}
+		for (int i = 0; i < xforms.size(); i++) {
+			Transform3D instance_transform = xforms[i];
+			instance_transform.origin += region_global_pos; // Convert to global position
+
+			real_t distance = instance_transform.origin.distance_to(p_picker_location);
```
TokisanGames: You can use distance_squared_to, since we don't need the real distance, only the consistent check across all cases.

Also this picks in 3D, but thinking about that it's probably correct. You could have plantA that is closer X/Z than plantB, but up on a higher ledge so is actually farther away.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -377,6 +377,13 @@ func _on_window_input(event: InputEvent) -> void:
 		EditorInterface.save_scene()
 
 
+func on_picked_mesh_asset_from_terrain(mesh_asset_id: int):
+	if not mesh_asset_id >= 0:
+		return
+
```
TokisanGames: trim whitespace

return -> void

`if mesh_asset_id < 0`

project/addons/terrain_3d/src/tool_settings.gd
```diff
@@ -357,6 +359,9 @@ func _on_picked(p_type: Terrain3DEditor.Tool, p_color: Color, p_global_position:
 			settings["angle"].value = p_color.r
 		Terrain3DEditor.SCALE:
 			settings["scale"].value = p_color.r
+		Terrain3DEditor.INSTANCER:
+			plugin.asset_dock.on_picked_mesh_asset_from_terrain(p_color.r)
```
TokisanGames: Just insert the whole contents of on_picked_mesh_asset_from_terrain here.

project/addons/terrain_3d/src/tool_settings.gd
```diff
@@ -154,6 +154,8 @@ func _ready() -> void:
 								"list":color_list, "default":0, "unit":"°", "range":Vector3(0, 360, 1) })
 	add_setting({ "name":"random_darken", "type":SettingType.SLIDER, "list":color_list, "default":50, 
 								"unit":"%", "range":Vector3(0, 100, 1) })
+	add_setting({ "name":"mesh_asset_picker", "type":SettingType.PICKER, "list":main_list,
```
TokisanGames: Can be renamed mesh_picker

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+	if (mesh_inst_dict.size() == 1) {
+		// Only one mesh type in this region, return it
+		return mesh_inst_dict.keys()[0];
+	}
+
+	// Multiple meshes, find the closest one
+	int closest_id = -1;
+	real_t closest_distance = FLT_MAX;
+	Array mesh_instance_keys = mesh_inst_dict.keys();
+
```
TokisanGames: remove ws, or put it above L969

src/terrain_3d_instancer.cpp
```diff
@@ -938,6 +938,62 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+		if (!cell_inst_dict.has(cell)) {
+			continue; // No instances in this cell
+		}
+		Array triple = cell_inst_dict[cell];
+		TypedArray<Transform3D> xforms = triple[0];
+		if (xforms.is_empty()) {
+			continue; // No transforms in this cell
+		}
+		for (int i = 0; i < xforms.size(); i++) {
+			Transform3D instance_transform = xforms[i];
```
TokisanGames: I think this copies the Transform. You might be able to just use the pointer being a read only operation:
`const Transform3D &instance_transform = xforms[i];`

If so, we could use this optimization throughout the instancer when doing read only.

TokisanGames (changes requested):
Ok great. I haven't tested it yet but the code looks good pending some minor tweaks. You can do them, or I'll make them when I test it.

src/terrain_3d_instancer.cpp
aidandavey: I can do that, but I was translating the copied transform. I can do this instead I think:

```			
const Transform3D &instance_transform = xforms[i];			
Vector3 instance_origin = instance_transform.origin + region_global_pos; // Convert to global position

real_t distance = instance_origin.distance_squared_to(p_picker_location);
```

aidandavey:
Needs to account for picking assets which are not in the current search results

TokisanGames:
I think you could just clear the search text.
Also you could rebase this, and add a texture picker as well since you're touching the same areas.

src/terrain_3d_texture_asset.h
```diff
@@ -41,7 +43,7 @@ class Terrain3DTextureAsset : public Terrain3DAssetResource {
 	int get_id() const override { return _id; }
 
 	void set_albedo_color(const Color &p_color);
-	Color get_albedo_color() const { return _albedo_color; }
+	Color get_albedo_color() const { return _highlighted ? get_highlight_color() : _albedo_color; }
```
TokisanGames: I think this should always return albedo color.

src/terrain_3d_instancer.cpp
```diff
@@ -402,8 +402,8 @@ Ref<MultiMesh> Terrain3DInstancer::_create_multimesh(const int p_mesh_id, const
 	return mm;
 }
 
-Vector2i Terrain3DInstancer::_get_cell(const Vector3 &p_global_position, const int p_region_size) {
-	IS_INIT(V2I_ZERO);
+Vector2i Terrain3DInstancer::_get_cell(const Vector3 &p_global_position, const int p_region_size) const {
+	IS_INIT(Vector2i());
```
TokisanGames: These are effectively the same. However V2I_ZERO is defined as:
`static const Vector2i V2I_ZERO{ 0, 0 };`

Which means that on a failure condition, your version creates a new class, allocates space for 2x 32-bit ints, assigns 0, and runs the constructor, then returns it. The previous version just returns the already created static variable, which is similarly shared throughout the code base in constructors and comparisons.

src/terrain_3d_instancer.cpp
```diff
@@ -1032,6 +1032,57 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+			Vector3 instance_origin = instance_transform.origin + region_global_pos; // Convert to global position
+
+			real_t distance = instance_origin.distance_squared_to(p_global_position);
+			if (distance < closest_distance) {
+				closest_distance = distance;
+				closest_id = mesh_id;
+			}
+		}
+	}
+	LOG(EXTREME, "Closest Mesh Asset ID: ", closest_id, " at: ", closest_distance);
```
TokisanGames: DEBUG is fine for this; it's a one time operation. Extreme is typically used for continuous things like snapping in order to avoid overwhelming the console.
Typically INFO at the start of a function and DEBUG for the detail. But if too much output then EXTREME for the detail, possibly cutting out INFO.

src/terrain_3d_instancer.cpp
```diff
@@ -1032,6 +1032,57 @@ void Terrain3DInstancer::update_transforms(const AABB &p_aabb) {
...
+int Terrain3DInstancer::get_closest_mesh_id(const Vector3 &p_global_position) const {
+	LOG(INFO, "Finding mesh asset ID closest to specified position ", p_global_position);
+	Vector2i region_loc = _terrain->get_data()->get_region_location(p_global_position);
+	Terrain3DRegion *region = _terrain->get_data()->get_region_ptr(region_loc);
+	if (!region) {
+		LOG(WARN, "No region found at position: ", p_global_position);
+		return -1; // No region found
+	}
+	Vector3 region_global_pos = v2v3(region_loc) * _terrain->get_region_size() * _terrain->get_vertex_spacing();
+	Vector2i cell = _get_cell(p_global_position, region->get_region_size());
```
TokisanGames: As far as the instancer is concerned, region size and terrain region size should always be the same. The only time they might be different is if Data is loading a region. If loading one with a different size as the first region, Data will change to the region's size. If loading a second with a different size, Data will error.

TokisanGames (commented):
Great to have these functions. Thanks for working on it.

I made some changes:
* Now that the highlight functions are in both asset resource classes, I added them to the AssetResource interface.
* Renamed some functions in both C++/GDScript
* Changed/fixed the decal so it changes color when picking for any type. This used to be the case a long time ago but got lost along the way.
* Fixed highlight button turning crimson only on meshes
* Fixed texture highlight colors getting saved to the file. The albedo/highlight change needed to be applied in the Assets class where the material is constructed.

TokisanGames:
Great work, thank you.

PR #736, opened 2025-06-24
Highlight selected MMI on select in asset dock

aidandavey:
Adds a new icon to the asset dock mesh entry container. When checked the selected mesh asset will be highlighted on the ground in a random, highly saturated colour.

It temporarily swaps out the geometry instance overlay material (if it exists) and restores it when no longer highlighting this asset.

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -291,6 +291,8 @@ func _on_meshes_pressed() -> void:
 
 
 func _on_tool_changed(p_tool: Terrain3DEditor.Tool, p_operation: Terrain3DEditor.Operation) -> void:
+	remove_all_highlights()
+	
 	if p_tool == Terrain3DEditor.INSTANCER:
 		_on_meshes_pressed()
 	elif p_tool in [ Terrain3DEditor.TEXTURE, Terrain3DEditor.COLOR, Terrain3DEditor.ROUGHNESS ]:
@@ -313,6 +315,16 @@ func update_assets() -> void:
 
 	_current_list.update_asset_list()
 
+
+func remove_all_highlights():
+	if not plugin.terrain:
+		return
+		
+	for i: int in mesh_list.entries.size():
+		var resource: Terrain3DMeshAsset = mesh_list.entries[i].resource 
+		if resource:
+			resource.set_highlighted(false)
+	
 
 ## Window Management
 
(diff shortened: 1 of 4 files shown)


Xtarsia:
This works. Though im not sure about how it overides the material overlay in the inspector. Especially if wanting to use the overlay.

It should be possible to apply via the extensition side of things without showing it in the inspector.

Also im not sure that the tool settings is the correct place to have this check box. It should likley belong in the plugin settings as part of editor settings (which could do with some better dressing up, looking at it now)

aidandavey:
We could drill into overlay material -> next_pass recursively until we find an empty material slot (in the extension if that's better).

It should restore the old overlay as it is, but you can't view both.

aidandavey:
Could you screenshot where you think the box should go please?

Xtarsia:
here:
image

Maybe better, would be here actually, in the "advanced" menu:
image

Setting the next, next pass may well work, tho it may unintentionally modify the user material if not careful.

aidandavey:
I think if I stash the actual user material then duplicate it with the next free next pass being the highlight material that should be safe.

I prefer the checkbox in the advanced settings where it's relatively easy to access.

Thanks!

TokisanGames:
I intended the highlight would be another icon right on the asset thumbnail like visible, edit, delete, and now indicator count. Visibility is the current method for identifying instance types on the ground. [Yuri's theme explorer](https://github.com/YuriSizov/godot-editor-theme-explorer) makes it easy to explore the built in icons.

For the highlight, I was thinking it would retain the user override/overlay material and replace it (on geometry) with the temporary highlight. Then if they uncheck highlight, it would replace the material with the one the user specified. It shouldn't change the inspector at all, but be handled internally.

The highlight should be something that can be seen far away. Imagine looking for a piece of grass that is somewhere on a 4k or 16k terrain.

Also we probably want a random color (within a visible bright range) so that if we highlight 5 things, they're all different colors.

aidandavey:
Made the changes discussed on discord

TokisanGames (changes requested):
Thanks for working on this. I think the C++ can be simplified greatly by pushing all of the work to the MeshAsset instead of the Instancer. The Instancer doesn't need to know about any of this. 

Currently, if you add a material override, all instances change instantly. So, the MA could have an internal material override, and if highlighted it applies the internal to geometry override material; if not highlighted restore the user provided material override to the geometry override. This would automatically update all of the MMIs.

TokisanGames:
Ok, nice work. It's very clean. Thanks!

Minor issues I addressed:
* Overlay only half covered the generated texture billboard cards. Changed highlight to override.
* Normal overlay material on the assets didn't work. Adjusted instancer logic.
* I modulated the thumbnail same color as highlight, and consolidated the icons to one with modulation (which overwrites the thumbnail highlight).
* Changed the mouse cursor to the hand for all of the thumbnail buttons

PR #738, opened 2025-06-24
Add foliage instance counter to asset dock as noted in #43

### LINKED ISSUE #546: Minor feature: show amount of mesh instances
### Description

Show how many instances there are of a mesh on the corner of the icons in Meshes tab

aidandavey:
Admin edit: Closes #546

Adds a counter label to the asset dock.

image

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -79,10 +79,12 @@ func initialize(p_plugin: EditorPlugin) -> void:
 	asset_container = $Box/ScrollContainer
 
 	texture_list = ListContainer.new()
+	texture_list.name = "TextureList"
 	texture_list.plugin = plugin
 	texture_list.type = Terrain3DAssets.TYPE_TEXTURE
-	asset_container.add_child(texture_list)
+	asset_container.add_child(texture_list, true)
 	mesh_list = ListContainer.new()
+	mesh_list.name = "MeshList"
 	mesh_list.plugin = plugin
 	mesh_list.type = Terrain3DAssets.TYPE_MESH
 	mesh_list.visible = false
@@ -125,7 +127,7 @@ func _ready() -> void:
 
 	update_thumbnails()
 	confirm_dialog = ConfirmationDialog.new()
-	add_child(confirm_dialog)
+	add_child(confirm_dialog, true)
 	confirm_dialog.hide()
 	confirm_dialog.confirmed.connect(func(): _confirmed = true; \
 		emit_signal("confirmation_closed"); \
@@ -358,12 +360,12 @@ func create_window() -> void:
 	window.wrap_controls = true
 	var mc := MarginContainer.new()
 	mc.set_anchors_preset(PRESET_FULL_RECT, false)
-	mc.add_child(self)
-	window.add_child(mc)
+	mc.add_child(self, true)
+	window.add_child(mc, true)
 	window.set_transient(false)
 	window.set_size(plugin.get_setting(ES_DOCK_WINDOW_SIZE, Vector2i(512, 512)))
 	window.set_position(plugin.get_setting(ES_DOCK_WINDOW_POSITION, Vector2i(704, 284)))
-	plugin.add_child(window)
+	plugin.add_child(window, true)
 	window.show()
 
 
@@ -412,7 +414,7 @@ func _on_godot_focus_exited() -> void:
 func load_editor_settings() -> void:
 	floating_btn.button_pressed = plugin.get_setting(ES_DOCK_FLOATING, false)
 	pinned_btn.button_pressed = plugin.get_setting(ES_DOCK_PINNED, true)
-	size_slider.value = plugin.get_setting(ES_DOCK_TILE_SIZE, 83)
+	size_slider.value = plugin.get_setting(ES_DOCK_TILE_SIZE, 90)
 	_on_slider_changed(size_slider.value)
 	set_slot(plugin.get_setting(ES_DOCK_SLOT, POS_BOTTOM))
 	if floating_btn.button_pressed:
@@ -453,6 +455,7 @@ class ListContainer extends Container:
 
 	
 	func _ready() -> void:
+		name = "ListContainer"
 		set_v_size_flags(SIZE_EXPAND_FILL)
 		set_h_size_flags(SIZE_EXPAND_FILL)
 
@@ -509,7 +512,7 @@ class ListContainer extends Container:
 		entry.changed.connect(_on_resource_changed.bind(id))
 		entry.type = type
 		entry.asset_list = p_assets
-		add_child(entry)
+		add_child(entry, true)
 		entries.push_back(entry)
 		
 		if p_resource:
@@ -641,7 +644,7 @@ class ListContainer extends Container:
 ##############################################################
 
 
-class ListEntry extends VBoxContainer:
+class ListEntry extends MarginContainer:
 	signal hovered()
 	signal selected()
 	signal changed(resource: Resource)
@@ -656,7 +659,12 @@ class ListEntry extends VBoxContainer:
 	var is_highlighted: bool = false
 	var asset_list: Terrain3DAssets
 	
-	@onready var button_row := HBoxContainer.new()
+	@onready var focus_style: StyleBox = get_theme_stylebox("focus", "Button").duplicate()
+	@onready var background: StyleBox = get_theme_stylebox("pressed", "Button")
+	@onready var label_rows := VBoxContainer.new()
+	var name_label: Label
+	var count_label: Label
+	@onready var button_row := FlowContainer.new()
 	@onready var button_clear := TextureButton.new()
 	@onready var button_edit := TextureButton.new()
 	@onready var spacer := Control.new()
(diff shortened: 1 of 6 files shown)


project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -734,6 +748,24 @@ class ListEntry extends VBoxContainer:
...
+		add_child(count_label, true)
+		count_label.visible = false
+		count_label.horizontal_alignment = HORIZONTAL_ALIGNMENT_RIGHT
+		count_label.vertical_alignment = VERTICAL_ALIGNMENT_BOTTOM
+		count_label.size_flags_vertical = Control.SIZE_EXPAND_FILL
+		count_label.add_theme_color_override("font_color", Color.WHITE)
+		count_label.add_theme_color_override("font_shadow_color", Color.BLACK)
+		count_label.add_theme_constant_override("shadow_offset_x", 1.)
+		count_label.add_theme_constant_override("shadow_offset_y", 1.)
+		count_label.add_theme_font_size_override("font_size", 18)
```
Xtarsia: This size will likley want a`` * EditorInterface.get_editor_scale()`` tacking on, and maybe a little smaller default size like 14.

Xtarsia (commented):
Tested, works exactly as expected! Nice.

src/terrain_3d_instancer.cpp
```diff
@@ -370,6 +371,41 @@ Vector2i Terrain3DInstancer::_get_cell(const Vector3 &p_global_position, const i
 	return cell;
 }
 
+void Terrain3DInstancer::_update_instance_count_array() {
```
TokisanGames: It's good that you're only counting xforms.size(). I think could simplify this and reuse the loops in _update_mmis(). As it creates the MMIs and populates them, you could just accumulate the lod0 `mmi->get_instance_count()` or xforms.size() as you have it.

The signal instance_count_changed, could be renamed "mmis_updated" so it's more generic, and it will still go out at the same time at the end of the function.

src/terrain_3d_instancer.h
```diff
@@ -40,8 +40,7 @@ class Terrain3DInstancer : public Object {
 	std::unordered_map<Vector2i, Node3D *, Vector2iHash> _mmi_containers;
 
 	uint32_t _density_counter = 0;
-	uint32_t _get_density_count(const real_t p_density);
-
```
TokisanGames: Let's keep a space between functions and variables.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -133,6 +132,16 @@ func _ready() -> void:
 	confirm_dialog.canceled.connect(func(): _confirmed = false; \
 		emit_signal("confirmation_closed"); \
 		emit_signal("confirmation_canceled") )
+			
+
+func update_instance_count():
+
```
TokisanGames: Remove this blank line and the one below. And check if initialized, return if not. See update_assets()

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -133,6 +132,16 @@ func _ready() -> void:
...
+	var arr: Array = plugin.terrain.instancer.get_instance_count_array()
+
+	for i: int in mesh_list.entries.size() -1:
```
TokisanGames: Just a slight spacing adjustment `size() - 1:` 

But are you sure you want -1? `for n in 8:` goes from 0 to 7. https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#for

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -310,10 +320,12 @@ func update_assets() -> void:
...
 		if not plugin.terrain.assets.meshes_changed.is_connected(mesh_list.update_asset_list):
 			plugin.terrain.assets.meshes_changed.connect(mesh_list.update_asset_list)
-
+		if not plugin.terrain.assets.meshes_changed.is_connected(update_instance_count):
+			plugin.terrain.assets.meshes_changed.connect(update_instance_count)
+		if not plugin.terrain.instancer.instance_count_changed.is_connected(update_instance_count):
+			plugin.terrain.instancer.instance_count_changed.connect(update_instance_count)			
 	_current_list.update_asset_list()
 
-
```
TokisanGames: Should have two lines here

TokisanGames (changes requested):
Looks great! Thanks for setting this up. I think the C++ can be consolidated a bit. Notes below.

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: Thanks, yes the final entry is the 'Add Mesh' button which doesn't need a label.

src/terrain_3d_instancer.cpp
aidandavey: I actually started with this approach but _update_mmi()s can update single regions, while I want to accumulate across all regions. I started to track the accumulation in a dictionary then realized I was recreating the original data structure so I opted to go through the original data separately as it seemed simpler. 

I will rename and move the signal emission.

I also found I hadn't limited the counting to the mesh_id which is being updated, so I'll include that now.

TokisanGames: Hmm, good thinking to make sure we get everything, but how about this. If update_mmis doesn't need to update the MMI, then it also doesn't need to update the counter. Both will happen the first time. Every update only needs to do the ones that are modified.

aidandavey: Yes this makes sense, i will try on uodate add the differnce between the pre update and post update counts. This should correctly account for removals as well.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -766,6 +768,40 @@ class ListEntry extends VBoxContainer:
 			name_label.text = "Add Mesh"
 
 
+	func setup_count_label() -> void:
+		if type == Terrain3DAssets.TYPE_TEXTURE:
+			return			
+		count_label = Label.new()
+		add_child(count_label, true)
+		count_label.visible = true
+		count_label.horizontal_alignment = HORIZONTAL_ALIGNMENT_RIGHT
```
TokisanGames, line 775: {A13175C6-5CAD-46CB-AF56-230EC8DB9692}

This needs a bit of margin on the right side. You can probably acheive that by setting position.x = -5.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -766,6 +768,40 @@ class ListEntry extends VBoxContainer:
...
+
+
+	func update_count_label() -> void:
+		if not type == Terrain3DAssets.AssetType.TYPE_MESH:
+			return
+		var mesh_resource: Terrain3DMeshAsset = resource as Terrain3DMeshAsset
+		if not mesh_resource:
+			count_label.text = str(0)
+		else:
+			count_label.text = str(mesh_resource.get_instance_count())
```
TokisanGames: I think it would be nicer if we use comma delimited strings. We're certainly going to have instances up into the millions. You can add and use this.

```
func format_number(num: int) -> String:
	var is_negative = num < 0
	var str_num = str(abs(num))
	var result = ""
	var length = str_num.length()
	for i in range(length):
		result = str_num[length - 1 - i] + result
		if i < length - 1 and (i + 1) % 3 == 0:
			result = "," + result
	return "-" + result if is_negative else result
```

src/terrain_3d_instancer.cpp
```diff
@@ -137,6 +137,10 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
 
 					// Create MM and assign to MMI
 					mmi = cell_mmi_dict[cell];
+
+					// Update count : Subtract previous count for this cell
+					ma->update_instance_count(lod == ma->get_last_lod() && mmi->get_multimesh().is_valid() ? -mmi->get_multimesh()->get_instance_count() : 0);
```
TokisanGames: This is a bit long and awkward. A two part conditional, three groups of function calls, all as a parameter. Please break it up, with the update_instance_count() call within, if the condition is true.

src/terrain_3d_instancer.cpp
```diff
@@ -295,13 +304,18 @@ void Terrain3DInstancer::_destroy_mmi_by_cell(const Vector2i &p_region_loc, cons
...
+		}
+
 		LOG(EXTREME, "Freeing ", ptr_to_str(mmi), " and erasing mmi cell ", p_cell);
 		remove_from_tree(mmi);
 		memdelete_safely(mmi);
 		cell_mmi_dict.erase(p_cell);
 		if (cell_mmi_dict.empty()) {
 			LOG(EXTREME, "Removing mesh ", mesh_key, " from cell MMI dictionary");
 			mesh_mmi_dict.erase(mesh_key); // invalidates cell_mmi_dict
+			ma.is_valid() ? ma->set_instance_count(0) : void(); // Reset count for this mesh
```
TokisanGames: The bug is likely here. I don't get the logic of it anyway. This function only has visibility into a cell, not even a full region, let alone all regions. I'm surprised the bug triggers only after clearing a whole region, rather than just a cell.

src/terrain_3d_mesh_asset.h
```diff
@@ -58,6 +58,7 @@ class Terrain3DMeshAsset : public Terrain3DAssetResource {
 	Ref<Material> _highlight_mat;
 	TypedArray<Mesh> _meshes;
 	Ref<Texture2D> _thumbnail;
+	int _instance_count = 0;
```
TokisanGames: This should be a uint32_t, as should function returns.

src/terrain_3d_mesh_asset.cpp
```diff
@@ -163,6 +163,16 @@ void Terrain3DMeshAsset::set_enabled(const bool p_enabled) {
 	emit_signal("instancer_setting_changed");
 }
 
+void Terrain3DMeshAsset::update_instance_count(const int p_amount) {
+	_instance_count += p_amount;
```
TokisanGames: This should clamp the result between 0 and uint32_max, as should set().

src/terrain_3d_instancer.cpp
aidandavey: You're right and this doesn't belong here. I will move it to the end of clear_by_mesh()

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: I repositioned as suggested. There is still some oddness when the tiles are small, not just with this label but stretching of the thumbnail and overlapping tiles. 
image

aidandavey: I propose looking at this in a separate PR or setting a larger minimum tile size for a quick fix.

TokisanGames:
Great! Thanks for the new feature.

Issue #501, opened 2024-09-26
Improve AssetDock


TokisanGames:
### Description

AssetDock has a clunky design and needs to be rewritten.

* [x] Add text filter box #792 
* [ ] Add a sort id independent of texture_id/mesh_id #956 
* [ ] Allow drag to reorder
* [ ] Move delete button to inside the inspector when editing (https://github.com/godotengine/godot/pull/96290)
* [ ] Improve & Optimize thumbnail generation
* [ ] Rewrite the class, dropping or reducing the sub classes and have a single class handling it

aidandavey:
> * [ ] Add a sort id independent of texture_id/mesh_id 

This may be achieved as a side effect of the search box PR, #792

The entry.id  (position in the entrylist) will  no longer used to set the brush data. Instead resource.id is passed.

PR #739, opened 2025-06-25
Add multiselect to asset dock

aidandavey:
Allows selection of multiple mesh assets in the dock for instancing or removal.

Always keeps at least one asset selected.

Click tile to select/deselect

image


Xtarsia:
Theres a few issues testing this:

Having more than 1 mesh asset, and then trying to remove the last asset when not currently selected, forces it to be selected. Which then causes errors when painting assets.

Having CrystalC as asset ID 0, Rock A as ID 1, have both selected, then add a 3rd asset, Rock B, causes ID 0 to render as Rock B, along with the thumbnail to swap to Rock B, even though CrystalC is still the chosen Scene file in the inspector.
Saving and reloading didnt restore the rendered meshes to the correct state, even though the Scene file for ID0 was still showing as CrystalC.
Re-loading CrystalC for ID0 results in ID2 becoming CrystalC.
This issue seemed to only get worse as the number of meshes inceased.

When it did work, it was nice to use, I can see it being very useful!

aidandavey:
> Theres a few issues testing this:
> 
> Having more than 1 mesh asset, and then trying to remove the last asset when not currently selected, forces it to be selected. Which then causes errors when painting assets.
> 
> Having CrystalC as asset ID 0, Rock A as ID 1, have both selected, then add a 3rd asset, Rock B, causes ID 0 to render as Rock B, along with the thumbnail to swap to Rock B, even though CrystalC is still the chosen Scene file in the inspector. Saving and reloading didnt restore the rendered meshes to the correct state, even though the Scene file for ID0 was still showing as CrystalC. Re-loading CrystalC for ID0 results in ID2 becoming CrystalC. This issue seemed to only get worse as the number of meshes inceased.
> 
> When it did work, it was nice to use, I can see it being very useful!

Both of these issues have now been addressed, thanks!

aidandavey:
This had gone through a few changes since the last tests to bring it up to date and to add support for 4.6

************

This adds the ability to select multiple Mesh Assets and paint them in a single operation. I changed the selected_id int to an Array

Left clicking a tile in the asset dock will add it to the selection, or deselect it if it is already selected.

Right clicking a tile will add it to the selection, but will not remove from the selection.

Density is independent for each MA, so you can paint grass at high density with lower density trees.

Instance removal is similarly multi-select, it will remove instances of the MAs that are currently selected.

Search - because we are selectively showing tiles here, we get a mismatch between the tile id and the asset id. To fix it, I have changed to a system whereby the selected_ids array contains the asset_ids, rather than tile ids. This means we can make a selection, perform a search, add a new MA to the selection, clear the search without losing our original selection.

src/terrain_3d_instancer.h
```diff
@@ -50,9 +50,9 @@ class Terrain3DInstancer : public Object {
 	V2IIntPair _queued_updates;
 
 	InstancerMode _mode = NORMAL;
-	uint32_t _density_counter = 0;
+	PackedInt64Array _density_counters;
```
TokisanGames: PackedInt32Array is sufficient, or even `std::vector<uint32_t>`, which has a `resize(n, val)`. We don't need Godot's cowdata overhead so for this we should either use Godot's LocalVector (which we haven't used anywhere yet) or std::vector (which we have, and has the resize+val function). We can look at all usage of Packed*Arrays and ensure we only use them when we are passing arrays back to Godot.

src/terrain_3d_instancer.h
```diff
@@ -50,9 +50,9 @@ class Terrain3DInstancer : public Object {
...
 
-	uint32_t _get_density_count(const real_t p_density);
+	uint32_t _get_density_count(const real_t p_density, const uint32_t p_id);
```
TokisanGames: Generally the id/type/region_loc would come first, value affecting that id second. 

Also this function should be called _update_density_count() or _calc, since get implies const.

src/terrain_3d_instancer.h
```diff
@@ -107,13 +107,22 @@ VARIANT_ENUM_CAST(Terrain3DInstancer::InstancerMode);
 
 // Allows us to instance every X function calls for sparse placement
 // Modifies _density_counter, not const!
-inline uint32_t Terrain3DInstancer::_get_density_count(const real_t p_density) {
+inline uint32_t Terrain3DInstancer::_get_density_count(const real_t p_density, const uint32_t p_idx) {
```
TokisanGames: idx should come first, and name it the same as the declaration.

src/terrain_3d_editor.cpp
```diff
@@ -872,7 +880,6 @@ void Terrain3DEditor::set_brush_data(const Dictionary &p_data) {
 
 	_brush_data["enable_texture"] = p_data.get("enable_texture", true);
 	_brush_data["texture_filter"] = p_data.get("texture_filter", false);
-	_brush_data["asset_id"] = CLAMP(int(p_data.get("asset_id", 0)), 0, ((_tool == INSTANCER) ? Terrain3DAssets::MAX_MESHES : Terrain3DAssets::MAX_TEXTURES) - 1);
```
TokisanGames, line 875: This defaulted to id 0. Now texture painting is ignored if there is no asset_ids. Perhaps we should insert [0] if it doesn't exist.

Also it doesn't sanitize the brush data. Sanitization was limited only to 0-32/256 before instead of the actual count. We could check _terrain->get_assets()->get_texture_count() and _terrain->get_assets()->get_mesh_count().

src/terrain_3d_instancer.h
```diff
@@ -107,13 +107,22 @@ VARIANT_ENUM_CAST(Terrain3DInstancer::InstancerMode);
...
 	uint32_t count = 0;
-	if (p_density < 1.f && _density_counter++ % uint32_t(1.f / p_density) == 0) {
-		count = 1;
-	} else if (p_density >= 1.f) {
+	if (p_idx >= _density_counters.size()) {
+		_density_counters.push_back(0);
```
TokisanGames: This is fine if idx is 1 greater than the count. If it's 2 or more L115 will crash. It should resize(idx+1).

src/terrain_3d_instancer.cpp
```diff
@@ -627,130 +627,168 @@ void Terrain3DInstancer::set_mode(const InstancerMode p_mode) {
...
 	}
 }
 
 void Terrain3DInstancer::remove_instances(const Vector3 &p_global_position, const Dictionary &p_params) {
 	IS_DATA_INIT_MESG("Instancer isn't initialized.", VOID);
 
-	int mesh_id = p_params.get("asset_id", 0);
+	if (p_params.has("asset_id")) {
+		LOG(ERROR, "We are no longer using asset_id, we need an array of asset_ids");
+	}
```
TokisanGames: This can be removed and instead it can be documented [here](https://github.com/TokisanGames/Terrain3D/discussions/756) and [here](https://terrain3d.readthedocs.io/en/latest/api/class_terrain3dinstancer.html#class-terrain3dinstancer-method-add-instances) and [here](https://terrain3d.readthedocs.io/en/latest/api/class_terrain3dinstancer.html#class-terrain3dinstancer-method-remove-instances).

src/terrain_3d_instancer.cpp
```diff
@@ -627,130 +627,168 @@ void Terrain3DInstancer::set_mode(const InstancerMode p_mode) {
...
 	}
-	LOG(EXTREME, "Adding ", count, " instances at ", p_global_position);
 
-	real_t fixed_spin = CLAMP(real_t(p_params.get("fixed_spin", 0.f)), .0f, 360.f); // degrees
-	real_t random_spin = CLAMP(real_t(p_params.get("random_spin", 360.f)), 0.f, 360.f); // degrees
-	real_t fixed_tilt = CLAMP(real_t(p_params.get("fixed_tilt", 0.f)), -180.f, 180.f); // degrees
-	real_t random_tilt = CLAMP(real_t(p_params.get("random_tilt", 10.f)), 0.f, 180.f); // degrees
-	bool align_to_normal = bool(p_params.get("align_to_normal", false));
+	for (int i = 0; i < mesh_ids.size(); i++) {
+		int mesh_id = mesh_ids[i];
```
TokisanGames, line 646: Can be `for (const int mesh_id : mesh_ids)`

src/terrain_3d_instancer.cpp
```diff
@@ -627,130 +627,168 @@ void Terrain3DInstancer::set_mode(const InstancerMode p_mode) {
...
-	// Density based on strength, mesh AABB and input scale determines how many to place, even fractional
-	uint32_t count = _get_density_count(density);
-	if (count <= 0) {
-		return;
+
+	Array mesh_ids = p_params.get("asset_ids", Array());
+
+	if (_density_counters.size() != mesh_ids.size()) {
+		_density_counters.resize(mesh_ids.size());
+		_density_counters.fill(0);
```
TokisanGames: Since add_instances() is repeatedly called during operate(), this clears all the density counters if somehow the selected mesh ids change mid-stroke, which could be done via the API or future gdscript changes.
So you could set from old_size to new size. Or look at std::vector::resize(n, val) which will initialize on resize.

src/terrain_3d_instancer.h
```diff
@@ -107,13 +107,22 @@ VARIANT_ENUM_CAST(Terrain3DInstancer::InstancerMode);
...
+	}
+	uint32_t density_counter = static_cast<uint32_t>(_density_counters[p_idx]);
+	density_counter++; // Increment first
```
TokisanGames: The original only incremented if p_density < 1.f. Idk if it changes behavior.

src/terrain_3d_instancer.h
```diff
@@ -107,13 +107,22 @@ VARIANT_ENUM_CAST(Terrain3DInstancer::InstancerMode);
...
+		uint32_t interval = uint32_t(1.f / p_density);
+		if (density_counter % interval == 0) {
+			count = 1;
+		}
+	} else {
 		count = uint32_t(p_density);
 	}
+	_density_counters[p_idx] = static_cast<int64_t>(density_counter);
 	return count;
 }
```
TokisanGames, line 120: Look at this version:
```c++
std::vector<uint32_t> _density_counters;

inline uint32_t Terrain3DInstancer::_update_density_count(const uint32_t p_id, const real_t p_density) {
	if (p_id >= _density_counters.size()) {
		_density_counters.resize(p_id + 1, 0);
	}
	uint32_t &counter = _density_counters[p_id];
	if (p_density < 1.f) {
		++counter;
		const uint32_t interval = uint32_t(1.f / p_density);
		return uint32_t(counter % interval == 0);
	}
	return uint32_t(p_density);
}
```

TokisanGames (changes requested):
Here's a first pass reviewing only the C++. So far so good, only one crash and some minor things.

src/terrain_3d_instancer.h
aidandavey: I ended up here while trying to put an unsigned int in the signed packed int array, but as you say it's not necessary at all. Agreed that std::vector is the better way here.

src/terrain_3d_instancer.cpp
aidandavey: We need the current index from the selected_index array when we call _update_density_count(index, p_density)

src/terrain_3d_editor.cpp
aidandavey: I have implemented sanitation and need to test it

src/terrain_3d_instancer.cpp
aidandavey: Removed the code and updated the docs (edited the xml files) 
When it's merged could you update the tracker issue please? Or I can add it as a new comment.

src/terrain_3d_editor.cpp
```diff
@@ -888,6 +895,32 @@ void Terrain3DEditor::set_brush_data(const Dictionary &p_data) {
...
+			}
+		}
+		asset_ids = valid_ids;
+	}
+	// Default to first asset id if none provided
+	if (asset_ids.is_empty()) {
+		asset_ids.push_back(0);
+	}
+	_brush_data["asset_ids"] = asset_ids;
+
```
TokisanGames:
```suggestion
	{ // Get and sanitize asset ids. If invalid, remove from list and log error
		TypedArray<int> asset_ids = p_data.get("asset_ids", TypedArray<int>());
		if (_tool == Tool::TEXTURE) {
			TypedArray<int> valid_ids;
			int max_asset_id = -1;
			if (_tool == Tool::TEXTURE) {
				max_asset_id = _terrain->get_assets()->get_texture_count();
			} else if (_tool == Tool::INSTANCER) {
				max_asset_id = _terrain->get_assets()->get_mesh_count();
			}
			for (const int id : asset_ids) {
				if (id < max_asset_id) {
					valid_ids.push_back(id);
				} else {
					LOG(ERROR, "Brush data contains invalid asset id: ", id, ". Max asset id is: ", max_asset_id - 1);
				}
			}
			asset_ids = valid_ids;
		}
		// Default to first asset id if none provided
		if (asset_ids.is_empty()) {
			asset_ids.push_back(0);
		}
		_brush_data["asset_ids"] = asset_ids;
	}
```

TokisanGames: Hard to see, but I just enclosed the section in a {} block.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -451,49 +453,44 @@ class ListContainer extends Container:
...
-			plugin.ui.toolbar.change_tool("PaintTexture")
-		elif type == Terrain3DAssets.TYPE_MESH and plugin.editor.get_tool() != Terrain3DEditor.INSTANCER:
-			plugin.ui.toolbar.change_tool("InstanceMeshes")
-		set_selected_id(p_id)
-
+		if type == Terrain3DAssets.TYPE_TEXTURE:
+			if not plugin.editor.get_tool() in [ Terrain3DEditor.TEXTURE, Terrain3DEditor.COLOR, Terrain3DEditor.ROUGHNESS ]:
+				plugin.ui.toolbar.change_tool("PaintTexture")
+			var res: Terrain3DTextureAsset = entries[p_id].resource
+			selected_asset_ids = [ res.get_id() ]
```
TokisanGames: This will break if res is null. We do have null resources at times, such as after clicking `add new`. Applies to the other branches here and in the other asset dock.

It will also break if p_id <0 or >= entries.size(). Maybe collect the ids at the beginning of the function and audit them first. Use a local variable called something like resource_id so it's clear from the code that entries[p_id] and entry.resource.get_id() are not the same ID type at all.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -451,49 +453,44 @@ class ListContainer extends Container:
...
+		elif type == Terrain3DAssets.TYPE_MESH:
+			if plugin.editor.get_tool() != Terrain3DEditor.INSTANCER:
+				plugin.ui.toolbar.change_tool("InstanceMeshes")
+			var res: Terrain3DMeshAsset = entries[p_id].resource
+
+			if selected_asset_ids.has(res.get_id()):
+				if not p_mouse_button == MOUSE_BUTTON_RIGHT:
```
TokisanGames: how about `!= right`, or `== left` ?

project/addons/terrain_3d/src/asset_dock_45.gd
```diff
@@ -565,7 +567,7 @@ class ListContainer extends Container:
 				add_item(mesh)
```
TokisanGames: The 4.6+ version checks `if mesh:`

project/addons/terrain_3d/src/asset_dock_45.gd
```diff
@@ -565,7 +567,7 @@ class ListContainer extends Container:
...
 			if mesh_count < Terrain3DAssets.MAX_MESHES:
 				add_item()
-		set_selected_id(selected_id)
+			update_selection()
```
TokisanGames: The 4.6+ version does not indent this.

project/addons/terrain_3d/src/asset_dock.gd
TokisanGames: Both of these can be updated to the new style, as well as any others in the this or the other asset_dock.

TokisanGames (changes requested):
Now I've actually used it in 4.7. It's great in that it brings in everything at the existing densities.

I'd suggest refining the design and functionality of the 4.6+ version first, then when that's set, we can modify 4.5 to match, rather than keeping the two in sync under development. Unless you find it easier to do both.

* Clicking around the asset dock, adding assets, not yet painting gave this error twice:
```
ERROR: Error calling from signal 'clicked' to callable: 'Container::clicked_id': Method expected 2 argument(s), but called with 1.
   at: emit_signalp (core/object/object.cpp:1311)
   GDScript backtrace (most recent call first):
       [0] _drop_data (res://addons/terrain_3d/src/asset_dock.gd:880)
```

* I don't care for the always on multiselect. I'd like to see:
  - click 1, 2, 3, 4, 5 -> changes individual selection from 1, 2, 3, 4, 5
  - click 5, 5, 5, 5, 5 -> selects 5 only (currently reverts to 1 on second)
  - click 1, shift click 5 -> selects 1 through 5
  - click 1, ctrl click 5 -> selects 1 and 5
  - If searching, they can do ctrl+click to add to the existing selection, shift-click to deselect and grab only these filtered ones (eg. "grass" ids 3, 10, 12, 78), or click to deselect all and grab only that one. (ctrl+shift click could do both, add this selection to the existing, but that's not required, it could be treated as shift).
  - Multi-asset action button clicks, say 5 are selected: 
    - Click action button of one that is selected: if they click visible or highlight, all selected should toggle. If they click delete _it should_ delete those but that requires more redesign in a later PR. It could loop a delete call starting with the largest number, which will delete or clear. If they click the pencil, it should edit only the one they clicked. These actions shouldn't change the selection.
    - Click action button one that isn't selected: I think it makes sense to deselect all, select the one, and do that action.
    - Right click should do the same as above, deselect all, select the one and edit


* I found a bug. I had 6 assets (card, lod4, card, card, crystal, rock) and had selected 1, 3, 4, 5. I painted the terrain and saw the expected assets appear, the crystal at a lower density, ok. I selected only the crystal. Then ctrl-click painted to remove only the crystal, ok. I ctrl+shift+clicked to remove all, ok. Then while the crystal (5) alone was selected, I repainted the terrain and assets 1-4 painted on the terrain, not 5.

  * I then clicked the crystal and all other assets except the crystal selected. This demonstrates some activities are leaving a dirty state.

* We've been inching our way towards #956, where we have a separate dock index and a Terrain3DAsset ID. This PR does some ID stuff, but the usage is confusing as I noted in clicked_id, and it's not clear to me where we stand on it. It would be good to more clearly define which kind of an ID we're talking about at all times. I think that starts with what we name them.
  * We use slot in a few places, but that should be reserved for the Godot editor dock slots.
  * Generic ID, asset_id, texture_id, mesh_id are used all over the C++, and resource_id in gdscript to refer to the index into the assets array.
  * We are using ListContainer and ListEntry subclasses for these dock tiles / icons. Perhaps always using _idx for the dock and _id for the C++ object. entry_idx, list_idx, dock_idx, tile_idx, or _tile as a number like _slot might be if we could use it. Entry matches our existing code, and _idx would be short and distinct from _id. We could use names like selected_entry, selected_entries, selected_idx, selected_indices, clicked_idx, clicked_entry, clicked_entry_idx. What do you think?

aidandavey:
Yes let's get the 4 6+ version nailed down first and I'm on board with the renaming suggestion.

Hopefully the bug you found will be fixed while making the selection handling changes. 

Shift select means we need to track the last entry_id that was clicked (not the asset_id). That will get messy in search mode as it is now. Currently search only adds entries that match. Perhaps the answer is to add all entries when searching and just hide the ones that don't match the search. This should leave the entry_ids intact and we can then revert to selecting by entry_id.

I'll work through it and may have some questions as I go.

TokisanGames:
> Shift select means we need to track the last entry_id that was clicked (not the asset_id). 

I think it's important for usability to have shift click and ctrl click as I described. I'm ok with tracking the most recently selected index in addition to the selected indices.

> That will get messy in search mode as it is now. 

What if we had multiple lists and indices? By default we show the unfiltered list of all indices, and have selected_indices.
In search mode, it has it's own indices, say filtered_indices. Then clicking, shift clicking, ctrl clicking is relative to what is shown. If search is active, clicks are relative to filtered_indices, which get translated into selected_indices upon selection. The mapping of filtered_idx to selected_idx already exists as resource_id.

> Currently search only adds entries that match. Perhaps the answer is to add all entries when searching and just hide the ones that don't match the search. This should leave the entry_ids intact and we can then revert to selecting by entry_id.

I don't follow the suggestion.

aidandavey:
> 
> > Currently search only adds entries that match. Perhaps the answer is to add all entries when searching and just hide the ones that don't match the search. This should leave the entry_ids intact and we can then revert to selecting by entry_id.
> 
> I don't follow the suggestion.

I'm finding this hard to explain fully, it may be garbage... but here goes. 

When we display the assets, we add the ListEntry to an array named entries - and it is assigned an id based on entries.size()

They are added to the array in add_item() - which also creates the control node and adds it to the container control. 

Search returns early inside the add_item function, so those items are skipped and not added to the array or the panel. 

entries[0] could be a different asset when searching vs not searching.

That's a problem when you want to make a selection, than add to the selection from a filtered list. This is why I took the select by asset_id approach instead. 

However if we want to select UI entries 2-5, it make far more sense to say 2-5 than "from the tile that matches id of the last asset I selected, and the tile that matches the id of the tile I just selected."

My proposal is to add all of the assets to the entries array and simply hide the controls for the entries which don't match the search, which should maintain the UI/entries id relationship - meaning we can rely on the previous method of selecting by ui_id. It should look the same to the user and be cleaner to work and reason with for us.

TokisanGames:
Let's talk about it on video.

TokisanGames:
Here are the notes from our meeting. I added some general / future notes to #501 .

* Extract subclasses from both 4.5 and 4.6 (hopefully they're identical) and add Terrain3D Prefix.
ListContainer -> Terrain3DListContainer
ListEntry -> Terrain3DListEntry

* New Workflow
mesh_list = Terrain3DAssets.get_mesh_assets()
tex_list = Terrain3DAssets.get_tex_assets()
current_list = instancer_tool ? mesh_list : tex_list
display_list = current_list.sort() # no-op
display_list = display_list.search("grass")
display_list.highlight_selected()

* Asset dock always shows display_list (instead of current list, right now)
* `selected_list` stores entry/asset/asset_id and populates based on input
* When >= 2 are selected, modify the meshes button label `Meshes (5)`

* Input:
click - clear selected_list, select this one (regardless of search filter)
ctrl+click - add this one to selected_list
shift+click - if last_selected is in display_list, add all assets from display_list to the clicked asset (could go forward or backwards). If last selected is not in the display list, this could act the same as ctrl+click.

* To select all rock and grass:
search "grass"
click grass1
shift+click grass5
search "rock"
ctrl+click (or shift+click) rock1
shift+click rock5

* Actions:
  * Any highlight button - highlight all selected
  * Any visible button - toggle visible all selected
  * Any X button - reverse loop over asset id and delete for all selected
  * Any edit button - edit that one
  * These don't change the selection

* Remember when referring to an asset dock entry or list index in variable and function names, use _idx or entry. When referring to a Terrain3D Asset ID, use _id. e.g. `clicked_id(tile_id)` -> `clicked_idx(entry_idx)` or `clicked_entry(entry)`

aidandavey:
I'm pretty happy with this now and think it's in a good place for re-testing and review. It got a little bigger than planned but I think it's heading in a good direction.

src/terrain_3d_texture_asset.cpp
```diff
@@ -79,7 +79,7 @@ void Terrain3DTextureAsset::set_id(const int p_new_id) {
 
 void Terrain3DTextureAsset::set_highlighted(const bool p_highlighted) {
 	SET_IF_DIFF(_highlighted, p_highlighted);
-	LOG(INFO, "Set mesh ID ", _id, " highlight: ", p_highlighted);
+	LOG(INFO, "Set texture ID ", _id, " highlight: ", p_highlighted);
```
TokisanGames: This should print _highlighted. It doesn't matter here, but it does matter in other places where there's a clamp, and this will be consistent.

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,344 @@
...
+## Filtered list of asset resources
+var display_list: Array
+## Array of ASSET_IDS
+var selected_list : Array[int] = [0]
+
+var height: float = 0.
+var width: float = 90.
+var focus_style: StyleBox
+# TODO: See if this is needed following the changes
+var _clearing_resource: bool = false
```
TokisanGames, line 22: Is this needed following the changes?

src/terrain_3d_assets.cpp
```diff
@@ -722,6 +724,8 @@ void Terrain3DAssets::_bind_methods() {
 	BIND_CONSTANT(MAX_TEXTURES);
 	BIND_CONSTANT(MAX_MESHES);
 
+	ADD_SIGNAL(MethodInfo("ids_swapped"));
```
TokisanGames: We haven't done this elsewhere yet, but I wonder if defining the types we're going to send is a good idea for error checking at the least.
```
ADD_SIGNAL(MethodInfo("ids_swapped",
    PropertyInfo(Variant::INT, "type"),
    PropertyInfo(Variant::INT, "old_id"),
    PropertyInfo(Variant::INT, "new_id")));
```

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,344 @@
...
+
+func _on_resource_hovered(p_id: int):
+	if type == Terrain3DAssets.TYPE_MESH:
+		if plugin.terrain:
+			plugin.terrain.assets.create_mesh_thumbnails(p_id, Vector2i(512, 512), true)
+
+
+func set_selected_after_swap(p_type: Terrain3DAssets.AssetType, p_old_id: int, p_new_id: int) -> void:
+	EditorInterface.mark_scene_as_unsaved()
+	selected_list = [ p_new_id ]
```
TokisanGames, line 151: This will deselect all and only select the new id, that's intended?

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -260,52 +255,53 @@ func _on_slider_changed(value: float) -> void:
...
 func _on_meshes_pressed() -> void:
 	if plugin.debug:
 		print("Terrain3DAssetDock: _on_meshes_pressed")
-	if _updating_list or current_list == mesh_list:
+	plugin.ui.toolbar.change_tool("InstanceMeshes")
+
+	
+func set_dock_mode(tool: Terrain3DEditor.Tool) -> void:
+	var mesh_mode: bool = tool == Terrain3DEditor.Tool.INSTANCER
+	if not mesh_mode and not tool == Terrain3DEditor.Tool.TEXTURE:
```
TokisanGames: `and tool !=`

project/addons/terrain_3d/src/asset_dock_45.gd
```diff
@@ -311,33 +306,29 @@ func _on_slider_changed(value: float) -> void:
...
 func _on_meshes_pressed() -> void:
 	if plugin.debug:
 		print("Terrain3DAssetDock: _on_meshes_pressed")
-	if _updating_list or current_list == mesh_list:
+	plugin.ui.toolbar.change_tool("InstanceMeshes")
+
+	
+func set_dock_mode(tool: Terrain3DEditor.Tool) -> void:
+	var mesh_mode: bool = tool == Terrain3DEditor.Tool.INSTANCER
+	if not mesh_mode and not tool == Terrain3DEditor.Tool.TEXTURE:
```
TokisanGames: `and tool !=`

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -221,7 +216,7 @@ func _on_search_text_changed() -> void:
 	mesh_list.search_text = search_box.text
 	texture_list.search_text = search_box.text
 	current_list.update_asset_list()
-	current_list.set_selected_id(0)
+	current_list.update_selection()
```
TokisanGames: The 4.5 version calls `current_list.update_list_entries()` instead of
```
	current_list.update_asset_list()
	current_list.update_selection()
```

Perhaps the 4.5 version is better for each search character?

TokisanGames (commented):
Ok so far so good, but I can't test it due to a crash. I loaded into gd4.7.2.

I used to be able to drag a mesh scene onto the add new button, and now that doesn't work, with these errors:
```
ERROR: Terrain3DEditor#7570:set_brush_data:911: Brush data contains invalid asset id: 2. Max asset id is: 1
   at: push_error (core/variant/variant_utility.cpp:1023)
   GDScript backtrace (most recent call first):
       [0] _on_setting_changed (res://addons/terrain_3d/src/ui.gd:299)
       [1] update_selection (res://addons/terrain_3d/src/asset_dock_list_container.gd:211)
       [2] update_list_entries (res://addons/terrain_3d/src/asset_dock_list_container.gd:117)
       [3] add_entry_to_selection (res://addons/terrain_3d/src/asset_dock_list_container.gd:201)
       [4] _on_entry_clicked (res://addons/terrain_3d/src/asset_dock_list_container.gd:185)
       [5] _drop_data (res://addons/terrain_3d/src/asset_dock_list_entry.gd:314)
SCRIPT ERROR: Invalid access to property or key 'resource' on a base object of type 'previously freed'.
          at: Terrain3DListContainer._on_edit_clicked (res://addons/terrain_3d/src/asset_dock_list_container.gd:216)
          GDScript backtrace (most recent call first):
              [0] _on_edit_clicked (res://addons/terrain_3d/src/asset_dock_list_container.gd:216)
```

If I then click the add new mesh asset button I get a crash:
```
void Terrain3DAssets::set_mesh_asset():557
	_set_asset(TYPE_MESH, p_id, p_mesh_asset);
p_id=2
p_mesh_asset is null

Terrain3DAssets::_set_asset():167
	res->clear(); // crash
res is null
```

Here we could obviously gate the call to clear only if res is valid. But the bigger question is why is res null now when it hasn't been for years?

src/terrain_3d_assets.cpp
aidandavey: Yes I think it's a good idea - I'll tackle the signals throughout the rest of the project in a separate PR

aidandavey:
> But the bigger question is why is res null now when it hasn't been for years?

It turns out I wasn't setting the type of the list entry - which prevented drop_data() from working for mesh assets or packed scenes. The resource was never set.

project/addons/terrain_3d/src/asset_dock_list_container.gd
aidandavey: I couldn't make it work properly while maintaining the selection and figured swaps are rare enough that I'd accept losing the selection.

Watch the card in list entry idx 4 as I move the crystal entry

https://github.com/user-attachments/assets/5ea2820e-f39c-4449-979d-8ed7dfdccb21

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: Yes both should have been as per the 4.5 version, there is no need to update_asset_list() as we just update the display list.

project/addons/terrain_3d/src/asset_dock_list_container.gd
aidandavey: I think it won't be needed, when I make the changes I was referring to! But those changes are out of scope for this PR - for now I removed the comment (and a few others which were notes for future me).

src/terrain_3d_assets.cpp
aidandavey: I had a quick look and there were not too many that needed updating, so I included as a separate commit in this PR

src/terrain_3d_editor.cpp
```diff
@@ -888,6 +895,30 @@ void Terrain3DEditor::set_brush_data(const Dictionary &p_data) {
 	_brush_data["brush_spin_speed"] = CLAMP(real_t(p_data.get("brush_spin_speed", 0.f)), 0.f, 1.f);
 	_brush_data["gradient_points"] = p_data.get("gradient_points", PackedVector3Array());
 
+	if (_terrain && (_tool == Tool::TEXTURE || _tool == Tool::INSTANCER)) { // Get and sanitize asset ids. If invalid, remove from list and log error
```
TokisanGames, line 898: We should audit this for all tools every time this function is called.
set_brush_data() isn't called as part of operate() so who knows when it will be called. Our UI calls it only on settings change. I'm not sure it's called on tool changes, and I don't think it's called when we start operating. And users might work with the API in a different order than our UI is using.

project/addons/terrain_3d/src/asset_dock_list_container.gd
TokisanGames: Yeah in the video changing selection to another asset isn't good. Erasing the selection is the better path.

src/terrain_3d_instancer.cpp
```diff
@@ -627,130 +627,162 @@ void Terrain3DInstancer::set_mode(const InstancerMode p_mode) {
...
-	real_t height_offset = CLAMP(real_t(p_params.get("height_offset", 0.f)), -100.0f, 100.f); // meters
-	real_t random_height = CLAMP(real_t(p_params.get("random_height", 0.f)), 0.f, 100.f); // meters
+		real_t fixed_scale = CLAMP(real_t(p_params.get("fixed_scale", 100.f)) * .01f, .01f, 100.f); // 1-10k%
+		real_t random_scale = CLAMP(real_t(p_params.get("random_scale", 0.f)) * .01f, 0.f, 10.f); // +/- 1000%
+		Ref<Terrain3DMeshAsset> mesh_asset = _terrain->get_assets()->get_mesh_asset(mesh_id);
+		real_t density = CLAMP(.1f * brush_size * strength * mesh_asset->get_density() /
+						MAX(0.01f, fixed_scale + .5f * random_scale),
+				.001f, 1000.f);
+		// Density based on strength, mesh AABB and input scale determines how many to place, even fractional
+		uint32_t count = _update_density_count(i, density);
```
TokisanGames, line 659: This should use mesh_id not i. Possibly the cause for the 1% density issue.

src/terrain_3d_editor.cpp
```diff
@@ -888,6 +895,30 @@ void Terrain3DEditor::set_brush_data(const Dictionary &p_data) {
...
+		TypedArray<int> asset_ids = p_data.get("asset_ids", TypedArray<int>()[0]);
```
TokisanGames, line 899: I think this default value will cause a crash. You're accessing the first item of an empty array. It should just be TypedArray<int>(). Then resize or push into it if a default is needed.

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,331 @@
+# Copyright © 2025 Cory Petkovsek, Roope Palmroos, and Contributors.
+class_name Terrain3DListContainer
```
TokisanGames, line 2: I think the file names should be `terrain_3d_list_entry.gd`, and `terrain_3d_list_container.gd`

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,331 @@
...
+	if p_clear:
+		selected_list.clear()
+	for entry_idx: int in p_entry_idx_array:
+		if not _entry_idx_to_asset_id_map.has(entry_idx):
+			push_error("%s tried to select entry_idx %s, was not present in the map")
+			continue
+		var asset_id: int = _entry_idx_to_asset_id_map[entry_idx]
+		if not selected_list.has(asset_id):
+			selected_list.push_back(asset_id)
+			_last_entry_clicked = entry_idx
```
TokisanGames, line 189: CTRL deselect: There's no deselection path. How about something like this?
```
		if selected_list.has(asset_id):
			if not p_clear:
				selected_list.erase(asset_id)
		else:
			selected_list.push_back(asset_id)
		_last_entry_clicked = entry_idx
```

TokisanGames: SHIFT-select: `_last_entry_clicked = entry_idx` resets the anchor item.

What if we only set _last_entry_clicked when p_clear is true?

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,331 @@
...
+
+	# Both valid: alphabetical compare
+	return a.name < b.name	
+	
+
+## This updates the visuals only for sorting and/or filtering - see also update_asset_list() 
+func update_list_entries() -> void:
+	clear_list_entries()
+	display_list = asset_list.duplicate()
+	display_list.filter(matches_search)
```
TokisanGames, line 107: I don't think this changes the array. I think it returns a changed array.

`display_list = display_list.filter(matches_search)`

https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-filter

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,331 @@
...
+
+
+func _on_highlight_clicked(p_entry: Terrain3DListEntry) -> void:
+	if not p_entry.resource:
+		return
+	var highlight_state: bool = p_entry.highlighted
+	if p_entry.resource.id in selected_list:
+		# Highlight all selected assets
+		for i: int in selected_list:
+			var entry: Terrain3DListEntry = _entries[i]
```
TokisanGames, line 216: Double check that i is the right index. 
Also `_on_clear_clicked()` and `_on_enable_clicked()`.
Do selected_list[] and entries[] always have the same number of elements and same indices?
selected_list stores asset_ids.
_entries stores Terrain3DListEntries.
Are the asset_ids aligned with the entries index ?

Do we need the reverse map of asset_id to entry index? Or to iterate over _entries and test entry.resource.id?

project/addons/terrain_3d/src/asset_dock_list_container.gd
```diff
@@ -0,0 +1,331 @@
...
+		if not _entry_idx_to_asset_id_map.has(entry_idx):
+			push_error("%s tried to select entry_idx %s, was not present in the map")
+			continue
+		var asset_id: int = _entry_idx_to_asset_id_map[entry_idx]
+		if not selected_list.has(asset_id):
+			selected_list.push_back(asset_id)
+			_last_entry_clicked = entry_idx
+			if plugin.debug:
+				print("Added %s to the selection, asset_id = %s" % [entry_idx, asset_id])
+	update_list_entries()
```
TokisanGames, line 192: This rebuilds the whole list. Would calling only update_selection() work?

TokisanGames (changes requested):
Ok testing usage:
 
* I think right click, and the edit button, should also select the asset. It feels a bit clunky now, as when I look back at the asset dock, I'm not sure which one I'm actually editing.

* It's not likely caused by this PR, but I noticed the strength input field size shrinks a little bit and cuts off the unit. So strength shows `30` instead of `30%`. At 100% it looks fine. This is in 4.7.2. We likely just need to expand the minimum size of the field.

* I selected 5 assets with CTRL, but I cannot deselect any of them with CTRL.

* I'm a little confused by the start/end of shift select. Say I have 5 assets. 
  * I click the middle. 3 selected.
  * Then I shift-click the left. 1-3 selected.
  * I shift-click the right: 3-5 selected.
  * I shift-click the left: 1-5 selected.

Windows retains the first selection and extends from there, regardless of how many times you click before or after the first selected. It seems like you're doing that with the first three steps. Then the last step changes the behavior by selecting all instead of 1-3. 

I could work with either model of selecting from the first selected. Or from the last selected. But right now, the behavior is inconsistent.

* At a strength of 30%, using all texture cards, they are added at a uniform rate. But
  * Removal is a little slower, and gets even slower the closer the count gets to 0 (probably out of scope)
  * At a strength of 1%, the selection order dramatically affects the rate. After a few seconds of painting I have counts of 304, 76, 19, 4. Whereas at 30% they would all be 304.

* The selection order adjusts instancing. Probably the same as above and maybe caused by using `mesh_id` vs `i`. If you select a texture card (high density) then a crystal (low density), then paint you get a different flow than if you select the crystal first then the texture card.

* I wonder if density or strength should be weighted by the number of items selected. So if you have 1 item at a density of 10 painted with a strength of 30%, compared to 3 items at a density of 10, painted at a strength of 30%. Should the output have the same amount of items on the ground or 3x the items on the ground? I'm leaning towards the first and maybe we can discuss it at today's meeting.

aidandavey:
> * I think right click, and the edit button, should also select the asset. It feels a bit clunky now, as when I look back at the asset dock, I'm not sure which one I'm actually editing.

If not already selected, should this add the edited asset to the selection, or select only the edited asset?

Or should this dependent on CTRL? i.e. add to selection and edit - or select and edit?

src/terrain_3d_editor.cpp
aidandavey: I'd intended this to default to an array containing [ 0 ] but I see that's not what I coded and agree it would cause a crash.

We already check if the array is empty later, so will stick to TypedArray() as suggested

src/terrain_3d_editor.cpp
aidandavey: The reason for this check was that I can only get a valid max_asset_id if either the texture tool or instancer tool is selected. If neither, we don't know what type of asset is selected, only what ids are selected. Unless we set the tool type in the brush data as well?

src/terrain_3d_instancer.cpp
aidandavey: I think 'i' is correct - _update_density_count() uses it as the index into an array of density_counters

src/terrain_3d_editor.cpp
TokisanGames: Terrain3DAssets should always know how many assets it has loaded, so we can always get the max count.
The question you bring up is are we checking asset_ids against meshes or textures, which is valid.

What we did before is validate:
* if instancer tool, clamp 0, MAX_MESHES-1
* else, clamp 0, MAX_TEXTURES-1
* if no value, = 0

We should maintain this.

It looks like you have no case for no value. Your `if asset_ids.is_empty` check requires terrain and the texture/instancer tool.

So what I'd say the new logic should be is:
* if instancer tool, clamp 0, _terrain?mesh_count():MAX_MESHES-1
* if texture tool, clamp 0, _terrain?texture_count():MAX_TEXTURES-1
* if no value, push 0

src/terrain_3d_instancer.cpp
TokisanGames: hmm, ok

TokisanGames:
> > I think right click, and the edit button, should also select the asset. 

> If not already selected, should this add the edited asset to the selection, or select only the edited asset?

I was referring only to single select. Just a single right-click, or a click on edit opens it in the inspector. If I look back at the asset dock and I see another tile selected it's confusing, especially when we have several very similar assets.

In regards to multi-select, 
Multi-select, visible/highlight -> should affect all and keep selection
Multi-select, delete ->reverse order delete, clear selection. I didn't test this.
Multi-select, edit/right-click -> we only edit one, I think a deselect all first is fine


>Or should this dependent on CTRL? i.e. add to selection and edit - or select and edit?

I don't think right-click needs to respond to shift or ctrl. It can just act as deselect, single select and edit.

project/addons/terrain_3d/src/asset_dock_list_container.gd
aidandavey: Yes, you are right! I had left the original filtering code so didn't notice this wasn't working at all.

project/addons/terrain_3d/src/asset_dock_list_container.gd
aidandavey: This only worked when the list is unfiltered. I think the reverse map would work however we wouldn't have any other use for it and iterating over _entries works properly - so I've gone with this.

project/addons/terrain_3d/src/asset_dock_list_container.gd
aidandavey: Yes, should have been using update_selection()

TokisanGames:
I ran through testing again.
We're getting close enough that updates should be in both 4.6 and 4.5 versions so its ready to merge when finished.

Improved:
* Click one, then shift click left/right now works like windows and is consistently selecting around its first anchor.
* Painting many meshes at 1% is now uniform
* Density no longer depends on select order
* Edit/right-click resets the selction to only the one

Issues:

* Multi select highlighting / visibility changes them all sequentially over several frames rather than all on one frame. Minor / won't block merging.

* Ctrl+shift+click does weird things. You probably need a clean if, else instead of if, if. In windows, shift+ctrl is additive. If you click middle, then shift+click above and below the highlight moves up/down around the anchor. But if you shift+ctrl+click, both up and down are added to the selection. We could copy that, or you could have shift override ctrl. Currently it's not good.


* I added several texture cards, shift-selected 2-6, painted and asset id 0 was instanced on the ground. If I ctrl click several IDs, it will paint the correct ones. Even when the list is filtered and ID 0 isn't in the list, and I select some of those available, it still paints ID 0.

* Deleting an embedded asset should just clear it without errors, but it produces all of these errors:
```
ERROR: Terrain3DAssets#3732:update_mesh_list:673: Terrain3DMeshAsset has null mesh at index 0
   at: push_error (core/variant/variant_utility.cpp:1023)
   GDScript backtrace (most recent call first):
       [0] update_resource (res://addons/terrain_3d/src/asset_dock_list_container.gd:293)
       [1] clear_resources (res://addons/terrain_3d/src/asset_dock_list_container.gd:270)
       [2] <anonymous lambda> (res://addons/terrain_3d/src/asset_dock.gd:114)
ERROR: Terrain3DAssets#3732:create_mesh_thumbnails:608: 0: Mesh is null, skipping
   at: push_error (core/variant/variant_utility.cpp:1023)
   GDScript backtrace (most recent call first):
       [0] update_asset_list (res://addons/terrain_3d/src/asset_dock_list_container.gd:60)
       [1] update_resource (res://addons/terrain_3d/src/asset_dock_list_container.gd:293)
       [2] clear_resources (res://addons/terrain_3d/src/asset_dock_list_container.gd:270)
       [3] <anonymous lambda> (res://addons/terrain_3d/src/asset_dock.gd:114)
ERROR: Terrain3DAssets#3732:create_mesh_thumbnails:608: 0: Mesh is null, skipping
   at: push_error (core/variant/variant_utility.cpp:1023)
   GDScript backtrace (most recent call first):
       [0] update_asset_list (res://addons/terrain_3d/src/asset_dock_list_container.gd:60)
       [1] _on_resource_changed (res://addons/terrain_3d/src/asset_dock_list_container.gd:299)
       [2] set_edited_resource (res://addons/terrain_3d/src/asset_dock_list_entry.gd:340)
       [3] clear_resources (res://addons/terrain_3d/src/asset_dock_list_container.gd:279)
       [4] <anonymous lambda> (res://addons/terrain_3d/src/asset_dock.gd:114)
```

* After painting, clearing middle asset texture cards, then going through and changing colors of the materials, I hit upon one texture card where after I changed the color, it wouldn't change the thumbnail. Then I selected and started painting anyway and it updated the thumbnail and painted fine. I haven't isolated the circumstance yet, but it might have been related to or caused by above.

PR #743, opened 2025-06-26
Duplicate meshes on create Terrain3DMeshAssets from scene file

aidandavey:
This solves a bug where MeshAssets could swap materials with other MeshAssets which share the same mesh resource. 

image

These were my steps:

1. Add RockA.tscn in slot 3
2. Add RockB.tscn in slot 4
3. Add RockC.tscn in slot 5
4. Add CrystalC.tscn in slot 6
5. Hover mouse over slot 5


TokisanGames:
Related to #617. In that fix, I initially duplicated the mesh, then reverted that change and used geometry overrides. I probably did so to avoid the minor duplication. I can't think of any other reason why, or any other issues with duplication.

TokisanGames:
Great, nice find. Thank you!

Issue #740, opened 2025-06-26
Make the debug heightmap range configurable


bkmeneguello:
### Description

When sculpting maps on small height scales (mine goes from 0 up to 10m) it's hard to use the heightmap debug because it doesn't have enough contrast. It would be useful to be able to adjust both the minimum value and the maximum in [this setting](https://github.com/TokisanGames/Terrain3D/blob/df8a175bb95ac05990c4ef319c5ef16dd9e165bd/src/shaders/debug_views.glsl#L35).
Another more advanced feature would be to add a custom color ramp, maybe a 1D texture, to avoid the pure gray shades on the height map.
A third suggestion on this, could be implemented a contour lines shader, with thin lines each 1un and thicker each 10un (configurable).

Thanks in advance.

TokisanGames:
Enable the override shader, insert the code and you can adjust it as you like with variable range or ramp. Perhaps we'll expand the built in one later, but it will be low priority. Or you could provide a PR.

Contour lines were already implemented in #680.

bkmeneguello:
Oh, glad to know the contour lines are already implemented. But I think it's not in the v1.0.0-stable release, probably on the next release or building from the repo.
I probably will work on a PR for the configurable heightmap. Do you have any constraints/approach for this implementation or can I just expose it to a material property?

TokisanGames:
Either a material property (appearing below the heightmap debug view), or a uniform (appearing in the shader variables) is fine. Either a range or a color ramp is fine. I don't think we need a texture.

bkmeneguello:
I added a partial implementation, with only the black to white ramp, I'll work on the color ramp in the next days.

TokisanGames:
Partial implementation in #744. I'd have closed this, but since you're working on the ramp, I'll leave it open.

bkmeneguello:
@TokisanGames after using the black/white heightmap with the new contour lines for a few days, I decided  not to pursue the color ramp because it would probably clutter unnecessarily the shader. Feel free to close this issue as completed.

PR #744, opened 2025-06-26
Add debug heightmap min and max values

bkmeneguello:
Adds Heightmap Min and Max to Terrain3DMaterial, the default values are from 0 - 300 as the original.
terrain_3d_heightmap
Related to #740


TokisanGames:
Thank you.
Did you ask copilot to review it or have your settings to do so, or are did Github automatically enable it?

bkmeneguello:
I think I enabled it recently, but I didn't like this intrusion, I'll disable it.

PR #747, opened 2025-07-01
texture height vertex displacement and mesh tessellation

Xtarsia:
This went through many itterations, but is finally getting close.

Implements:

- Clipmap Mesh subdivision, upto 5 levels, for 32 vertices per meter.
- A Displacement Buffer via Veiwport, that renders 1 pixel per vertex, for each subdivision, upto the full 1m vertices.
- A debug view that renders the buffer in fragment.
- Per asset height offset and scale settings.
- Material option to disable / enable displacement (seperate from the tessellation level mesh setting)
- use propper TNB martix for world space texture normals. Previous version was wrong (needed transforming from view space)

Todo:

- Fix cleanup / crash on quit
- Probably need to check vertex spacing
- fully disable more code when tessellation level is 0
- test mobile
- test web


TokisanGames:
You said you want to get #717 in next, before this is finalized?

Xtarsia:
> You said you want to get #717 in next, before this is finalized?

It can probably go after at this point (if for some reason this got merged first - tho given the scope of this PR, i think thats not likley!), since I managed to get the buffer syncing perfectly with the camera position. Originally I had a seperate buffer snap target as well, but thats no longer required.

Saul2022:
Testing on mobile  with the artifact the  terrain system seems to not work at all throwing these errors 

```

 Godot Engine v4.4.stable.official (c) 2007-present Juan Linietsky, Ariel Manzur & Godot Contributors.
--- Debug adapter server started on port 6006 ---
--- GDScript language server started on port 6005 ---
  ERROR: godot-cpp/src/godot.cpp:333 - Cannot load a GDExtension built for Godot 4.4.1 using an older version of Godot (4.4.0).
  ERROR: core/extension/gdextension_library_loader.cpp:233 - GDExtension initialization function 'terrain_3d_init' returned an error.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:20 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:21 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:24 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:399 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:44 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:82 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:109 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:185 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:223 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:229 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:233 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:234 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:373 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:374 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:400 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:414 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: modules/gdscript/gdscript.cpp:3022 - Failed to load script "res://addons/terrain_3d/src/editor_plugin.gd" with error "Parse error".
  ERROR: godot-cpp/src/godot.cpp:333 - Cannot load a GDExtension built for Godot 4.4.1 using an older version of Godot (4.4.0).
  ERROR: core/extension/gdextension_library_loader.cpp:233 - GDExtension initialization function 'terrain_3d_init' returned an error.
  ERROR: res://addons/terrain_3d/src/region_gizmo.gd:37 - Parse Error: Identifier "Terrain3DData" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/region_gizmo.gd:37 - Parse Error: Identifier "Terrain3DData" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/region_gizmo.gd:49 - Parse Error: Identifier "Terrain3DData" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:48 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:52 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:53 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:54 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:55 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:151 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:151 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:530 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:83 - Parse Error: Could not resolve external class member "tool_changed".
  ERROR: res://addons/terrain_3d/src/ui.gd:87 - Parse Error: Could not resolve external class member "picking".
  ERROR: res://addons/terrain_3d/src/ui.gd:99 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:99 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:142 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:143 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:165 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:170 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:174 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:174 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:176 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:180 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:188 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:192 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:204 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:215 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:226 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:226 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:226 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:231 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:267 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:301 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:302 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:305 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:307 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:308 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:310 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:343 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:358 - Parse Error: Identifier "Terrain3DMaterial" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:359 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:365 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:372 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:381 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:385 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:387 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:389 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:402 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:404 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:411 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:418 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:421 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:424 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:427 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:429 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:432 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:435 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:438 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:441 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:444 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:447 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:450 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:453 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:461 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:537 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:541 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:551 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:554 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:554 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:556 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:557 - Parse Error: Identifier "Terrain3DRegion" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:558 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:560 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:562 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:570 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/ui.gd:573 - Parse Error: Could not resolve external class member "pick".
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:20 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:21 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:24 - Parse Error: Could not find type "Terrain3DEditor" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:399 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:44 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:82 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:109 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:185 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:223 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:229 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:233 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:234 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:373 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:374 - Parse Error: Identifier "Terrain3DEditor" not declared in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:400 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/src/editor_plugin.gd:414 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://demo/src/DemoScene.gd:4 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: modules/gdscript/gdscript.cpp:3022 - Failed to load script "res://demo/src/DemoScene.gd" with error "Parse error".
  ERROR: res://addons/terrain_3d/extras/particle_example/terrain_3D_particles.gd:13 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: res://addons/terrain_3d/extras/particle_example/terrain_3D_particles.gd:13 - Parse Error: Cannot use simple "@export" annotation because the type of the initialized value can't be inferred.
  ERROR: res://addons/terrain_3d/extras/particle_example/terrain_3D_particles.gd:125 - Parse Error: Could not find type "Terrain3D" in the current scope.
  ERROR: modules/gdscript/gdscript.cpp:3022 - Failed to load script "res://addons/terrain_3d/extras/particle_example/terrain_3D_particles.gd" with error "Parse error".
  ERROR: Cannot get class 'Terrain3DMeshAsset'.
  ERROR: Cannot get class 'Terrain3DMeshAsset'.
  ERROR: Cannot get class 'Terrain3DTextureAsset'.
  ERROR: Cannot get class 'Terrain3DTextureAsset'.
  ERROR: Cannot get class 'Terrain3DAssets'.
  ERROR: Cannot get class 'Terrain3DMaterial'.
  ERROR: Cannot get class 'Terrain3D'.

```

TokisanGames:
Saul, that's your issue. Read the first error and use the correct Godot version, 4.4.1.

Saul2022:
> Saul, that's your issue. Read the first error and use the correct Godot version, 4.4.1.

Yea you right sorry  also the performance here is so funny lol ( this is with tessellation level 5 and its faster than dual scaling 🤣 great job, this could make it to the lightmapper  shader considering how well it performs.

Screenshot_20250705_115126
Screenshot_20250705_115119

src/terrain_3d.cpp
```diff
@@ -260,6 +267,51 @@ void Terrain3D::_destroy_mouse_picking() {
...
+		return;
+	}
+	_destroy_displacement_buffer();
+	LOG(INFO, "Setting up displacement buffer");
+	_d_buffer_vp = memnew(SubViewport);
+	_d_buffer_vp->set_name("DBufferViewport");
+	add_child(_d_buffer_vp, true);
+	_d_buffer_vp->set_size(Vector2i(2, 2));
+	_d_buffer_vp->set_disable_3d(true);
+	_d_buffer_vp->set_update_mode(SubViewport::UPDATE_ONCE);
```
TokisanGames, line 278: Some subviewport default settings have caused us troubles. Are you sure the defaults are OK? The mouse_vp sets a lot of defaults. Scaling bilinear, anisotropic filtering 4x (maybe should be disabled for the mouse), Shadow atlas? 

How about gui_disable_input (should be enabled for the mouse) `set_disable_input(bool p_disable);`. I set this on both VPs.

src/shaders/displacement_buffer.glsl
```diff
@@ -0,0 +1,375 @@
...
+
+// Public uniforms
+//INSERT: AUTO_SHADER_UNIFORMS
+uniform float displacement_sharpness : hint_range(0.0, 1.0, 0.01) = 0.5;
+uniform bool vertical_projection = true;
+uniform float projection_threshold : hint_range(0.0, 0.99, 0.01) = 0.8;
+
+// Varyings & Types
+
+// We only care about 1 value here.
```
TokisanGames: There are 5. Which one?

src/shaders/displacement_buffer.glsl
```diff
@@ -0,0 +1,375 @@
+R"(shader_type canvas_item;
+
+// Displacement buffer shader, mimics the main shader in 2d and out puts
```
TokisanGames: 'outputs'
and below, 'copies'

doc/docs/texture_displacement.md
```diff
@@ -0,0 +1,24 @@
...
+:target: ../_images/displacement_example.png
+```
+
+It should be noted that `mesh_size` & `vertex_spacing` will have a noticable effect on the subdivision distance, and vertex density. Generally it is recommended to keep `mesh_size` as high as performance allows, and `vertex_spacing` at `1.0` for the best results.
+
+```{image} images/displacement_wireframe.png
+:target: ../_images/displacement_wireframe.png
+```
+
+Setting up textures with height data as described in [Texture Prep](texture_prep.md) is a requirement for tessellated texture displacement.
```
TokisanGames: Should go first or second and bold, or prefaced with a bold `Note: `

doc/docs/texture_displacement.md
```diff
@@ -0,0 +1,24 @@
...
+
+Included in the `Terrain3DTextureAsset` settings, are `displacement_offset` & `displacement_scale`. These should be adjusted so that the surface is realisticly displaced, and centered around zero to minimize collision missmatch between displaced verticies and the terrain collision.
```
TokisanGames: mismatch, vertices.

doc/docs/texture_displacement.md
```diff
@@ -0,0 +1,24 @@
...
+
+Enabling the `displacement_buffer`debug view, will render the ammout a vertex has been displaced. Red for depressions, and Green for extrusions. This view can be handy when setting the offset and scale values, to ensure the texture is centered around 0.
```
TokisanGames: amount that.

The red/green info should go in the debug view XML so it shows in the tool tip. Actually, all of our debug views should have a better description explaining the visual.

godot-cpp
TokisanGames, line 1: This shouldn't be in the PR.

src/shaders/dual_scaling.glsl
```diff
@@ -13,7 +13,7 @@ group_uniforms;
 
 //INSERT: DUAL_SCALING
 	// dual scaling
-	float far_factor = clamp(smoothstep(dual_scale_near, dual_scale_far, length(v_vertex - _camera_pos)), 0.0, 1.0);
+	float far_factor = clamp(smoothstep(dual_scale_near, dual_scale_far, length(v_vertex - _target_pos)), 0.0, 1.0);
```
TokisanGames: Should dual scaling be relative to the target point or the camera?

src/shaders/editor_functions.glsl
```diff
@@ -21,7 +21,7 @@ uniform bool _editor_decal_visible[3]; // show decal: brush, slope point1, point
 uniform bool _editor_decal_part[2]; // show decal[0] component: texture, reticle
 
 float get_reticle(vec2 uv, int index) {
-	float cam_dist = length(_camera_pos - v_vertex);
+	float cam_dist = length(_target_pos - v_vertex);
```
TokisanGames: Shouldn't the decal be relative to the camera?

src/shaders/main.glsl
```diff
@@ -150,24 +151,45 @@ vec3 get_index_uv(const vec2 uv2) {
...
+	v_vertex_xz_dist = length(v_vertex.xz - _target_pos.xz);
 
 	// Geomorph vertex, set end and start for linear height interpolate
 	float scale = MODEL_MATRIX[0][0];
 	float vertex_lerp = smoothstep(0.55, 0.95, (v_vertex_xz_dist / scale - _mesh_size - 4.0) / (_mesh_size - 2.0));
 	vec2 v_fract = fract(VERTEX.xz * 0.5) * 2.0;
 	// For LOD0 morph from a regular grid to an alternating grid to align with LOD1+
-	vec2 shift = (scale < _vertex_spacing + 1e-6) ? // LOD0 or not
+	vec2 shift = (scale < _vertex_spacing / _subdiv + 1e-6) ? // LOD0 or not
 		// Shift from regular to symetric
```
TokisanGames: symmetric

src/shaders/main.glsl
```diff
@@ -150,24 +151,45 @@ vec3 get_index_uv(const vec2 uv2) {
...
 		mix(v_fract, vec2(v_fract.x, -v_fract.y),
-			round(fract(round(mod(v_vertex.z * _vertex_density, 4.0)) *
-			round(mod(v_vertex.x * _vertex_density, 4.0)) * 0.25))
+			round(fract(round(mod(v_vertex.z * _vertex_density * _subdiv, 4.0)) *
+			round(mod(v_vertex.x * _vertex_density * _subdiv, 4.0)) * 0.25))
 			) :
 		// Symetric shift
```
TokisanGames: symmetric

src/terrain_3d_mesher.cpp
```diff
@@ -315,15 +316,18 @@ void Terrain3DMesher::destroy() {
...
+	real_t tessellation_density = 1.f / pow(2.f, _terrain->get_tessellation_level());
+	real_t vertex_spacing = _terrain->get_vertex_spacing() * tessellation_density;
+	if (MAX(ABS(_last_target_position.x - target_pos_2d.x), abs(_last_target_position.y - target_pos_2d.y)) < vertex_spacing) {
 		return;
 	}
 	_last_target_position = target_pos_2d;
-
-	real_t vertex_spacing = _terrain->get_vertex_spacing();
 	Vector3 snapped_pos = (target_pos / vertex_spacing).floor() * vertex_spacing;
 	RS->material_set_param(_terrain->get_material()->get_material_rid(), "_camera_pos", snapped_pos);
```
TokisanGames: _camera_pos is now _target_pos? Thanks to grok.

src/terrain_3d.cpp
Xtarsia: the buffer viewport is 2D only, with 3D set to disabled. So most of the other settings for the mouse picking have no effect, and are skipped entirley.

I have added the nearest filtering.

src/terrain_3d_mesher.cpp
Xtarsia: This just needed removing.

godot-cpp
Xtarsia: oops. this slipped by me when I rebased.

Xtarsia:
> Can the displacement buffer uniform group be merged in with the displacement group?

I tried this, and when a uniform is present in the buffer shader, but not the terrain shader, it stops being part of any group.

> If displacement is disabled in the material..

I just removed the toggle from the shader, it was really only for AB testing.

> The minimum shader is broken.

Iv'e fixed the minimum shader, tho part of me is thinking that maybe we should be injecting the required vertex() snippets and functions as well, similar to auto-shader / dual scaling? That would be for later consideration and not for this PR however.

> Light weight shader is broken.

Fixed. Rather than add any support for displacement, I added a comment that it is not supported.

> If tessellation is set to 0, the displacement buffer uniform group is still visible.

Fixed.

As for grok:
> Mesher LOD Calculation for Tessellation..

It is correct, the method does adds extra rings, however the entire clipmap is downscaled by tesselation level, beyond vertex spacing values.

> Buffer Sharpness Uses Wrong Uniform..

Intentional, I added a small explanation to the docs: `This difference allows flatter materials to be more influenced by bumpier materials when blending.`

> Debug View Branching..

The #ifdef is the condition that prevents any of the issues occuring.

> Interpolated Weights in Buffer Only for Forward+

This is also evaluated correctly in both the buffer, and main shader.

> "explodes verts"

Not really, each increase in Tesselation Level, is the same ammount of addional vertices, as an increase of LOD by 1. Which Is linear. At Maximum settings, the buffer size is 1536 x 256.

> Optimize Buffer Updates

I see what its saying, but innitial attempts at this caused some weird stuttering. Also when tesselation is enabled, any movement at all is likley to cause an update, and the margin can potentially be a very small 0.015625m.

TokisanGames:
Please let me know whenever you're ready for another look. You can also rebase as I see there are some conflicts, as well as squashing, including fixup my commits.

src/terrain_3d_material.cpp
```diff
@@ -412,6 +444,23 @@ void Terrain3DMaterial::_update_shader() {
...
+		if (_buffer_shader_override->get_code().is_empty()) {
+			_buffer_shader_override->set_code(_generate_buffer_shader_code());
+		}
+		code = _buffer_shader_override->get_code();
+		if (!_buffer_shader_override->is_connected("changed", callable_mp(this, &Terrain3DMaterial::_update_shader))) {
+			LOG(DEBUG, "Connecting changed signal to _update_shader()");
+			_buffer_shader_override->connect("changed", callable_mp(this, &Terrain3DMaterial::_update_shader));
+		}
+	} else {
+		code = _generate_buffer_shader_code();
```
TokisanGames: We don't need to generate this if tessellation == 0. We should still set the buffer material with an empty so we can toggle it on and off and have it clear. Check my proposed commit.

TokisanGames:
Getting closer. I reviewed the usage and uploaded some commits that you can review. I will look closer at the code next.

Here are some further issues I've found:

* Like the autoshader and dual scale, I think we should disable tessellation by default, but enable it in the demo. I've done so.

* The lightweight shader is now broken when tessellation > 0, like the minimum shader was.

* Displacement doesn't work reliably outside of regions. Tessellation and the displacement buffer do. Go to a mountain top in the world noise and play with displacement offset/scale and it won't do anything. Go to the ground shortly outside a region and swap textures and displacement does work. Go to a mountain top within a region, even a far one, and it works fine.

## Texture Asset

* If texture asset displacement scale is 1 and offset of the cliff texture is changed 0 - 0.85 it moves evenly, but .85 to 1.0 the peaks squish like its clamped. Do we really need that clamp? What if the texture offset parameter allowed great than 1 if they input it?

* If offset is 0, and I slide displacement scale of cliff down to 0.2 it moves evenly, but from .2 to 0 it increases in size. Is there something that should be clamped there because it's going negative? If offset is 1, this doesn't occur.

## Material

* auto_slope, auto height reduction, base/overlay textures now appear twice in the auto shader group. They have the same values, so you need to de-duplicate what you add to the material parameters.
* Vertical projection/projection threshold appeared in the general and auto shader group. Now it appears in general/world noise.
* Displacement sharpness was in a displacement buffer group far down the list. Displacement scale appeared in the autoshader group. I moved both to a displacement group and sorted it.
* With the minimum shader, if tessellation > 0, it shows the vertical projection, auto shader and displacement blocks.

src/terrain_3d.cpp
```diff
@@ -634,9 +701,10 @@ void Terrain3D::set_vertex_spacing(const real_t p_spacing) {
 			update_region_labels();
 			_instancer->_update_vertex_spacing(_vertex_spacing);
 			_mesher->reset_target_position();
-			_material->_update_maps();
```
TokisanGames: This function is and was private. How did we call it? Is there a left over `friend` or `public:`? I don't see the removal of that facility.

TokisanGames: Indeed, Terrain3DMaterial has friend Terrain3D. However we're still using it to access `_update_texture_arrays()`.

src/terrain_3d.cpp
```diff
@@ -71,9 +72,9 @@ void Terrain3D::_initialize() {
 		_data->connect("region_map_changed", callable_mp(_collision, &Terrain3DCollision::build));
 	}
 	// Any map was regenerated or regions changed, update material
-	if (!_data->is_connected("maps_changed", callable_mp(_material.ptr(), &Terrain3DMaterial::_update_maps))) {
+	if (!_data->is_connected("maps_changed", callable_mp(_material.ptr(), &Terrain3DMaterial::update).bind(false))) {
 		LOG(DEBUG, "Connecting _data::maps_changed signal to _material->_update_maps()");
```
TokisanGames: The comment can be updated.

src/terrain_3d_material.cpp
```diff
@@ -479,7 +536,7 @@ void Terrain3DMaterial::_update_shader() {
 	notify_property_list_changed();
 }
 
-void Terrain3DMaterial::_update_maps() {
+void Terrain3DMaterial::_update_maps(const RID &p_material) {
```
TokisanGames: Terrain3DData::update_maps() actually regenerates the TextureArray maps.
This function could be renamed to _update_uniforms or _update_material perhaps. Though it does all uniforms except the texture arrays. 🤔

src/terrain_3d_material.cpp
```diff
@@ -596,18 +670,26 @@ void Terrain3DMaterial::destroy() {
...
+	if (_buffer_material.is_valid()) {
+		RS->free_rid(_buffer_material);
+		_buffer_material = RID();
+	}
 }
 
-void Terrain3DMaterial::update() {
-	_update_shader();
-	_update_maps();
+void Terrain3DMaterial::update(bool p_full) {
```
TokisanGames, line 694: Interestingly, this exposed, public function update() even with full, doesn't pass the texture arrays. It regenerates the shader, then passes all of the non-texture array uniforms. For completeness I think it should call _update_texture_arrays() in case the user does something and needs a full update, though I don't know what.

src/terrain_3d_material.cpp
```diff
@@ -582,9 +651,14 @@ void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
...
 	}
+	if (!_buffer_material.is_valid()) {
+		_buffer_material = RS->material_create();
+	}
 	_shader.instantiate();
+	_buffer_shader.instantiate();
 	_update_shader();
-	_update_maps();
+	_update_maps(_material);
+	_update_maps(_buffer_material);
```
TokisanGames: This should just call `update(true)`

src/terrain_3d.cpp
```diff
@@ -1,5 +1,6 @@
 // Copyright © 2025 Cory Petkovsek, Roope Palmroos, and Contributors.
 
+#include <godot_cpp/classes/color_rect.hpp>
```
TokisanGames: You added this in terrain_3d.h so isn't needed here.

src/shaders/main.glsl
```diff
@@ -186,21 +213,31 @@ void vertex() {
...
+		// Set final vertex height.
+		float h;
+		// This branch is static for each of the clipmap segments
+		// Interpolated reads only occur where sub-texel values are required.
+		if (scale < _vertex_spacing) {
+			h = mix(interpolated_height(start_pos), interpolated_height(end_pos), vertex_lerp);
+		} else {
+			ivec3 coord_a = get_index_coord(start_pos, VERTEX_PASS);
+			ivec3 coord_b = get_index_coord(end_pos, VERTEX_PASS);
+			h = mix(texelFetch(_height_maps, coord_a, 0).r,texelFetch(_height_maps, coord_b, 0).r,vertex_lerp);
```
TokisanGames: Please add spaces after the commas.

src/shaders/displacement_buffer.glsl
```diff
@@ -0,0 +1,376 @@
...
+uniform vec2 _texture_detile_array[32];
+uniform vec2 _texture_displacement_array[32];
+uniform highp sampler2DArray _height_maps : repeat_disable;
+uniform highp sampler2DArray _control_maps : repeat_disable;
+//INSERT: TEXTURE_SAMPLERS_NEAREST
+//INSERT: TEXTURE_SAMPLERS_LINEAR
+
+// Public uniforms
+uniform bool vertical_projection = true;
+uniform float projection_threshold : hint_range(0.0, 0.99, 0.01) = 0.8;
```
TokisanGames, line 62: You could wrap these in `group_uniforms general;` `group_uniforms;` to at least put them in the right group.

However, that still leaves them displaying twice in the main shader, and they appear on minimum and lightweight.

src/shaders/displacement_buffer.glsl
```diff
@@ -0,0 +1,376 @@
...
+
+// Inline Functions
+#define DECODE_BLEND(control) float(control >> 14u & 0xFFu) * DIV_255
+#define DECODE_AUTO(control) bool(control & 0x1u)
+#define DECODE_BASE(control) int(control >> 27u & 0x1Fu)
+#define DECODE_OVER(control) int(control >> 22u & 0x1Fu)
+#define DECODE_ANGLE(control) float(control >>10u & 0xFu) * TAU_16TH
+// This math recreates the scale value directly rather than using an 8 float const array.
+#define DECODE_SCALE(control) (0.9 - float(((control >>7u & 0x7u) + 3u) % 8u + 1u) * 0.1)
+#define DECODE_HOLE(control) bool(control >>2u & 0x1u)
```
TokisanGames, line 27: Spaces after >> on 3 lines above.

TokisanGames:
Ok I'm finished. Back to you to review my changes and fix up the remaining things.

src/terrain_3d_material.cpp
Xtarsia: ive dropped _update_texture_arrays() and folded it into ~~_update_maps()~~ _update_uniforms(), along with pointing the signal connection to update() as well.

src/terrain_3d.cpp
Xtarsia: with some minor restructuring, the class is no longer friends.

Xtarsia:
>     * Like the autoshader and dual scale, I think we should disable tessellation by default, but enable it in the demo. I've done so.

Agreed.

>     * The lightweight shader is now broken when tessellation > 0, like the minimum shader was.

Fixed.

>     * Displacement doesn't work reliably outside of regions. Tessellation and the displacement buffer do. Go to a mountain top in the world noise and play with displacement offset/scale and it won't do anything. Go to the ground shortly outside a region and swap textures and displacement does work. Go to a mountain top within a region, even a far one, and it works fine.

This is because dualscaling & world noise are not present in the buffer. I dont see any value in adding all the extra work to the background mode here.

> ## Texture Asset
>     * If texture asset displacement scale is 1 and offset of the cliff texture is changed 0 - 0.85 it moves evenly, but .85 to 1.0 the peaks squish like its clamped. Do we really need that clamp? What if the texture offset parameter allowed great than 1 if they input it?
> 
>     * If offset is 0, and I slide displacement scale of cliff down to 0.2 it moves evenly, but from .2 to 0 it increases in size. Is there something that should be clamped there because it's going negative? If offset is 1, this doesn't occur.
> 

There is an inherent "clamp" from the 8bit values of each channel in the buffer, which is essentially limited to +/- 0.5m (default).

as for the offset/scale, they are the maximum ranges, but it really can vary depening on the texture heights, so its upto the artist to set them to reasonable values. The offset is meant to work in tandem with scale, where a small scale would generally have an offset of at most 0.5, but for something like snow, it could be desirable to have it sit higher.

> ## Material
> 
>     * auto_slope, auto height reduction, base/overlay textures now appear twice in the auto shader group. They have the same values, so you need to de-duplicate what you add to the material parameters.
> 
>     * Vertical projection/projection threshold appeared in the general and auto shader group. Now it appears in general/world noise.
>
>     * Displacement sharpness was in a displacement buffer group far down the list. Displacement scale appeared in the autoshader group. I moved both to a displacement group and sorted it.
>
>     * With the minimum shader, if tessellation > 0, it shows the vertical projection, auto shader and displacement blocks.

Still fiddling with this.

src/terrain_3d_material.cpp
```diff
@@ -849,13 +941,23 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
...
+
 	_active_params.clear();
 	for (int i = 0; i < param_list.size(); i++) {
 		Dictionary dict = param_list[i];
 		StringName name = dict["name"];
 
-		// Filter out private uniforms that start with _
-		if (!name.begins_with("_") && !name.is_empty()) {
+		// Filter out private uniforms that start with _ and prevent duplicate entries
+		if (!name.begins_with("_") && !name.is_empty() && !_active_params.has(name)) {
```
TokisanGames: This filters out the duplicate vertical project. But also filters out `group_uniforms displacement`, which means displacement_sharpness, in the second shader, goes into world noise. This is what dict looks like for a group vs a uniform:

```
Terrain3DUtil:print_dict:43: Dictionary: General/VP:
Terrain3DUtil:print_dict:63: "name": Value: general
Terrain3DUtil:print_dict:63: "class_name": Value:
Terrain3DUtil:print_dict:63: "type": Value: 0
Terrain3DUtil:print_dict:63: "hint": Value: 0
Terrain3DUtil:print_dict:63: "hint_string": Value:
Terrain3DUtil:print_dict:63: "usage": Value: 64

Terrain3DMaterial#3693:_get_property_list:960: Param name: vertical_projection has: true
Terrain3DUtil:print_dict:43: Dictionary: General/VP:
Terrain3DUtil:print_dict:63: "name": Value: vertical_projection
Terrain3DUtil:print_dict:63: "class_name": Value:
Terrain3DUtil:print_dict:63: "type": Value: 1
Terrain3DUtil:print_dict:63: "hint": Value: 0
Terrain3DUtil:print_dict:63: "hint_string": Value:
Terrain3DUtil:print_dict:63: "usage": Value: 6
```

Type is Variant::Nil vs Bool
Usage is PROPERTY_USAGE_GROUP vs PROPERTY_USAGE_DEFAULT

So I'm keeping groups but de-duplicating entries.

TokisanGames:
Remaining:
* Establishing sensible defaults
* Addressing a crash in OOTA changing scenes in the editor between Wildeburn and a new one. Crashes after emitting "meshes_changed". Probably not related to displacement. Addressed in #825

TokisanGames:
Itemizing some new found issues discussed on discord and more.

* [ ] Material should have a global offset
* [x] Adjust offset ranges to -1 to 1
* [x] If texture and material offsets are 0, and scale is 0, the mesh should be pixel perfect with collision. I think this is the case currently.
* [x] Material scale is currently negative, and maybe off by one. This is very confusing. 
  * When Currently, w/ assets are offset 0.0, scale 0, as Material Scale increases, the mesh goes _below_ collision. 
  * What should happen is for both textures and material: as offsets increase, the mesh rises above collision. As offsets decrease, the mesh drops below collision. As scale increases above 0, the texture peaks rise higher. As scale goes towards 0 the displacement flattens. Setting the base line of the texture to 0.5 height is likely fine, which means scale would both increase and decrease the amplitude of the texture peak and troughs. So, what does a scale of 1 mean? I'm not sure, I think it's just whatever you've arbitrarily coded into the shader. Whereas an offset of 0 has a specific meaning: scale 0, and/or texture height value 0.5 matches collision.
* [x] The displacement tutorial documentation needs more guidance on workflow:
  * Disabling show_instances, particle shader, and enabling collision editor collision and gizmos so you can see where collision is.
  * The current comment about the displacement view is a good start, but could use elaboration on how to use the colors.
  * Maybe the debug view should also include a white contour line for where 0 / collision is. 
  * You mentioned this: _Another thing to mind when setting up textures with displacement, is the nornal map depth should be set so that the 'sides' in the normal map actually match the geometry_. This should be elaborated on, likely with pictures.
* [x] Changing tessellation_level in game causes the textures disappear
* [x] Changing the displacement offset of rock affects areas that have been painted (manual or auto) with grass and vica versa. I would expect them to be independent. 
* [x] Painting an auto grass area with grass changes the displacement. I would expect them to be the same.


Other noted issues present on `main`:
* [x] tri_scale_reduction modifies textures other than the dual_scale_texture
* [ ] Changing the asset settings in the editor doesn't update in game. Material settings do. I don't recall if it ever did update. Testing 1.0.1 it doesn't occur there. Perhaps we can look at adding this.

TokisanGames:

src/terrain_3d_texture_asset.cpp
```diff
@@ -140,16 +143,30 @@ void Terrain3DTextureAsset::set_normal_depth(const real_t p_normal_depth) {
...
 
-void Terrain3DTextureAsset::set_roughness(const real_t p_roughness) {
-	_roughness = CLAMP(p_roughness, -1.0f, 1.0f);
-	LOG(INFO, "Setting roughness modifier: ", _roughness);
+void Terrain3DTextureAsset::set_displacement_offset(const real_t p_displacement_offset) {
+	_displacement_offset = CLAMP(p_displacement_offset, -1.0f, 1.0f);
+	LOG(INFO, "Setting displacement_offset: ", _displacement_offset);
+	LOG(DEBUG, "Emitting setting_changed");
+	emit_signal("setting_changed");
+}
```
TokisanGames, line 187: After rebase, new form:
```
void Terrain3DTextureAsset::set_displacement_offset(const real_t p_displacement_offset) {
	SET_IF_DIFF(_displacement_offset, CLAMP(p_displacement_offset, -1.0f, 1.0f));
	LOG(INFO, "Setting displacement_offset: ", _displacement_offset);
	LOG(DEBUG, "Emitting setting_changed");
	emit_signal("setting_changed");
}
```

TokisanGames:
It works a lot better, and I intuitively get it now. 

* Since MA/Displacement Offset does nothing unless scale is >0, scale should be listed first.
* The red/green view is usable, but I prefer to see the collision.
  * I considered if we should put it on a hotkey like `5`, or if that hotkey could simultaneously change to displacement debug view, change to editor collision, and enable view gizmos. However, maybe just enabling editor collision is enough and better even than the debug view. 🤔 Either way, we've got to get rid of that region gizmo and start drawing the region boundaries in the shader only. #100 

{A8F5ED4A-B68A-4883-8181-CC64E8F614CE}

{0147537D-EBB2-4AFF-9895-B94AEE5F8287}


* We can and should update the displacement buffer when it's changed in the material. It's hard to see the effect in non-R/G view because moving the camera is jarring. This works to get the signal:
```
bool Terrain3DMaterial::_set(const StringName &p_name, const Variant &p_property) {
...
	if (p_name == StringName("displacement_sharpness")) {
		LOG(MESG, "displacement_sharpness: ", p_property);
	}
```

* Once I change the tessellation setting, or enable the displacement debug view, I get this error message once, 60-80% of the time when I run the game. It seems to be more often after I change tessellation or debug view:
```
ERROR: Method/function failed.
   at: get_dependencies (scene/resources/resource_format_text.cpp:1535)
```

* Can be rebased and conflicts addressed. `Re-expose generated_type` can be dropped.

---
That's my usage review. I'll look at the code and docs again a bit later. 

Some notes for me:
* Recommended view: gizmos + editor collision, or displacement debug view, change to editor collision, and enable view gizmos
* Displacement Sharpness changes the buffer, and is only updated on camera movement - (but could be fixed)

TokisanGames:
Why is the displacement buffer shader in the material? Maybe it makes more sense to put it in a subgroup in Mesh, enabled by tessellation? I anticipate the ocean mesh and ocean shader to be a separate group as well.

Xtarsia:
> Why is the displacement buffer shader in the material? Maybe it makes more sense to put it in a subgroup in Mesh, enabled by tessellation? I anticipate the ocean mesh and ocean shader to be a separate group as well.

Yeah, that does make sense actually.

TokisanGames:
Displacement scale could remain in the material as is. Or it could be a private uniform in the shader, set by a setting combined with the other displacement settings. Then everything is in one place.

Xtarsia:
> Displacement scale could remain in the material as is. Or it could be a private uniform in the shader, set by a setting combined with the other displacement settings. Then everything is in one place.

Yeah the terrain shader displacement scale is a "global" scale that determines the maximum displacement distance.

Eg if set to 2 (meters) then the maximum extrusion, and depresion is 1m, for a total of 2m.

Making it private and moving to the mesh section as a part  of a subgroup would make sense as well. (as well as adding some blurb about what I just said to the docs)

Xtarsia:
Moved the override, and displacement scale into a subgroup under mesh. (not sure how I make it hide if tessellation == 0)

Added the sharpness property update. (it really is much better seeing it in real time)

Dropped the generated type commit.

Added the extra notes to the docs.

TokisanGames:
> not sure how I make it hide if tessellation == 0

_get_property_list()

See OOTA/Lod.gd for an example. Add a subgroup there.

Xtarsia:
found out about _validate_property() after a bit of searching. Seems to do the trick nicley for this use case. (hiding the displacement group from the inspector)

updated docs a bit more to include your notes about displaying the collision mesh.

re-ordered displacement scale and offset.

moved the displacement scale inspector property from material to terrain3d as well.

Hopefully its ready now.

TokisanGames (approved):
Ok, looks good. I've adjusted the docs and parameter ordering.

The last remaining issue for this PR is Flat mode texturing is broken with the lightweight shader.
image

Please fix that, then you can merge in this PR.

PR #748, opened 2025-07-02
Instancer LOD0 AABB copied to all LODs

Xtarsia:
Further adresses issues with billboard LODs when using the instancer.

improves #691
#728 could be reverted if this holds, the tiny extra margin can probably stay imo.

### DIFF
--- a/src/terrain_3d_instancer.cpp
+++ b/src/terrain_3d_instancer.cpp
@@ -174,6 +174,25 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
 					// Clear the cell modified state
 					triple[2] = false;
 				}
+
+				// Set all LOD mmi AABB to match LOD0 to ensure no gaps between transitions.
+				AABB mmi_custom_aabb = AABB();
+				for (int lod = 0; lod <= ma->get_last_lod(); lod++) {
+					Vector2i mesh_key(mesh_id, lod);
+					CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
+					MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
+					if (lod == 0) {
+						mmi_custom_aabb = mmi->get_aabb();
+					} else {
+						mmi->set_custom_aabb(mmi_custom_aabb);
+					}
+				}
+				if (ma->get_shadow_impostor() >= 0) {
+					Vector2i mesh_key(mesh_id, Terrain3DMeshAsset::SHADOW_LOD_ID);
+					CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
+					MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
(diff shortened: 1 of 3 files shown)


TokisanGames:
Thanks for the fix. I think all of this could be integrated above and trimmed down to about 4 lines.

At the top of the `for m` loop, we know ma is valid and lods are >0, so you could just retrieve the lod0 aabb before the `for c` loop. Then later when creating all of the MMIs you could set that custom aabb.

Xtarsia:
> Thanks for the fix. I think all of this could be integrated above and trimmed down to about 4 lines.
> 
> At the top of the `for m` loop, we know ma is valid and lods are >0, so you could just retrieve the lod0 aabb before the `for c` loop. Then later when creating all of the MMIs you could set that custom aabb.

Thinking about this more, i dont think its reasonable to get the AABB for the multimesh any earlier - especially given the loop runs backwards from last to the potential shadow imposter. The AABB will be unique for each cell, depending on where the contained transforms are.

Its not the AABB of the mesh for LOD0 that is being copied, but rather the AABB of the actual mmi for LOD0.

src/terrain_3d_instancer.cpp
```diff
@@ -174,6 +174,25 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
+					Vector2i mesh_key(mesh_id, lod);
+					CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
+					MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
+					if (lod == 0) {
+						mmi_custom_aabb = mmi->get_aabb();
+					} else {
+						mmi->set_custom_aabb(mmi_custom_aabb);
+					}
+				}
+				if (ma->get_shadow_impostor() >= 0) {
```
TokisanGames: Should be > 0; _shadow_impostor defaults to 0, which means no impostor and use lod0 as the shadow caster.

src/terrain_3d_instancer.cpp
```diff
@@ -174,6 +174,25 @@ void Terrain3DInstancer::_update_mmis(const Vector2i &p_region_loc, const int p_
...
+					Vector2i mesh_key(mesh_id, Terrain3DMeshAsset::SHADOW_LOD_ID);
+					CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
+					MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
+					mmi->set_custom_aabb(mmi_custom_aabb);
```
TokisanGames: This crashed in the demo on the generated cards w/o a shadow mesh. I now check if mmi is valid.

TokisanGames:
I see, you're copying the MMI AABB not the mesh AABB. Ok.

My commit reverts #728 and #639, removing the overlap entirely for testing. I don't see any issues in the demo.

TokisanGames:
Thank you!

PR #752, opened 2025-07-04
Update control blend debug view

### LINKED ISSUE #745: Control blend view and Albedo out of sync
### Description

The control blend debug view doesn't match up with the look in albedo. Maybe it doesn't need to and should reflect the real values rather than apparent values. However in the demo when you turn it on it's all pink because the autoshader is a low value. So it looks broken. It needs a review at the least.

Xtarsia:
closes #745 

image

black <-> white, blend value.
red, auto shader enabled.


TokisanGames:
Looks good, thanks!

PR #759, opened 2025-07-08
Limit clipmap/collision targets to game mode; set for demo

TokisanGames:
* Changes Demo to set player as clipmap/collision targets
* Fixes vertex grid sticking to the clipmap target in editor - now editor only follows camera
* Fixes Dynamic editor collision sticking to collision target in editor - now editor only follows camera


TokisanGames:
> The collision target however, perhaps should stick to the chosen target, as I can see there being reasons supporting that behaviour instead.

What is an instance where we would want collision in the editor (which is usually off except during debugging/editing) to stick to the target?

I'm thinking we turn collision on in the editor for debugging, or for **level design**. And that's actually an important point that makes me think the current implementation is right. If we're using AssetPlacer or another tool that snaps stuff to collision in the editor, and collision sticks to the target, then I need either of the editor modes enabled. If I have a large landscape I have to generate collision for the whole world just to edit the level, or I need to clear the collision target, then remember to set it again before I run the scene.

Instead, I'd rather have the mesh/collision stuck to the targets in game mode, and stick to me in editor mode so it looks correct, and I can do level design. But I'll listen to a counter argument.

Xtarsia:
> > The collision target however, perhaps should stick to the chosen target, as I can see there being reasons supporting that behaviour instead.
> 
> What is an instance where we would want collision in the editor (which is usually off except during debugging/editing) to stick to the target?
> 
> I'm thinking we turn collision on in the editor for debugging, or for **level design**. And that's actually an important point that makes me think the current implementation is right. If we're using AssetPlacer or another tool that snaps stuff to collision in the editor, and collision sticks to the target, then I need either of the editor modes enabled. If I have a large landscape I have to generate collision for the whole world just to edit the level, or I need to clear the collision target, then remember to set it again before I run the scene.
> 
> Instead, I'd rather have the mesh/collision stuck to the targets in game mode, and stick to me in editor mode so it looks correct, and I can do level design. But I'll listen to a counter argument.

I just had visions of users with tools scripts expecting collision around the target in editor, and it not being there, causing them issues.

However I think your argument is FAR more likely to occur, so im in agreement with you now :)

PR #760, opened 2025-07-08
Experiments

TokisanGames:
* Add toggleable blend mode to Spray
* Fix Smooth not working with gradient - point select interfered with smoothing
* Fix Select Gradient tool -> hold Shift -> brush -> let go of Shift -> see hole created from divide by zero


Xtarsia:
Took it for a spin!

Some notes:

I think that Nav and Autoshader shouldnt do anything when shift is pressed.

When Averaging roughness/wetness, the brush decal still shows, it would be better if it mimiced the texture tool behaviour as well.

I think the alt mode should be the default, and perhaps limited to brushalpha of >0.5?

Average Color tends to over-darken for me. I commented on discord about swapping to an overlay style blend, but maybe out of scope here.

new crosshair is nice, the cross seems a little large at first, but its much more easy to use!

TokisanGames:
Smooth operations moved to #773.

> I think the alt mode should be the default, and perhaps limited to brushalpha of >0.5?

You mean the `alt blend mode` for texturing? That's the test I'm having Marcel perform. He agrees but I asked him to spend a few days confirming since it's a default change. I don't plan on keeping the option; just one or the other.

PR #761, opened 2025-07-09
implement group uniforms

### LINKED ISSUE #650: Support Uniform Groups
### Description

Rip any updated code in the godot material, especially for supporting uniform groups. Then add groups to the shader.

Uniform groups PR
https://github.com/godotengine/godot/pull/62972/files#diff-28012627e073e805396f46f62ddfbe09668a539433e113b48445a369eb1615c3

Latest
https://github.com/GodotEngine/godot/blob/master/scene/resources/material.cpp

Xtarsia:
Closes #650 

Groups uniforms together
image
When options are disabled, and thus not inserted:
image

Works with ShaderOverride, subgroups also work fine:
image

Caveat, Overlays and Debug views are on top..


TokisanGames:
We've lived all this time without any groups at all, we don't really need subgroups. We could make all groups/subgroups be subgroups under one shader uniforms group:

{E31B1DD9-40E1-4CFB-A77F-66F69C091148}

Or we could hide the debug views and overlays in the material since they're all aliased in the main node. I don't really like using them in the material because they make the inspector bounce up and down, whereas the main node ones do not.

{71263A2C-A7E8-4B6F-B474-FD6CCA8FB885}

Which is better?

Xtarsia:
I like moving the debug views and overlays more.

My other argument for subgroups is it keeps compatability with the override shader, that has groups & subgroups as a default feature.

TokisanGames:
Ok, I agree. Organized some things. Excellent work, thanks for taking care of that one.

PR #764, opened 2025-07-13
Fix import and region bugs

TokisanGames:
* Terrain3DData: Allows maps to be imported individually, without overwriting what is there
* Terrain3DRegion: Fixes a bug that would allow invalid region sizes and failed sanitization.
  * set_height_map(null)
  * _region_size == 0
  * set_region_size(0)
  * _region_size = CLAMP(p_region_size, 64, 2048); turns into 64
  * set_control/color_map(valid 1024 sized map)
  * region size does not get set properly, santize fails because of mismatch. Should have ignored heightmap, used control/color size, and sanitize generates a new heightmap

Pending issue:
* After importing, or changing assets, the terrain is invisible. Once the camera moves the mesher redraws. We're missing a distance reset to force snapping reset somehwere, maybe in update_maps.


src/terrain_3d.cpp
```diff
@@ -509,15 +521,13 @@ Vector3 Terrain3D::get_collision_target_position() const {
...
-	if (p_camera && p_camera->is_queued_for_deletion()) {
-		LOG(ERROR, "Attempted to set a node queued for deletion");
-		_camera.clear();
-		return;
+void Terrain3D::snap() {
+	if (_mesher) {
+		_mesher->reset_target_position();
+	}
+	if (_collision) {
+		_collision->reset_target_position();
```
Xtarsia, line 529: This might as well be called by __physics_process()

src/terrain_3d_region.cpp
```diff
@@ -13,13 +13,24 @@
...
 		LOG(WARN, "Region ", get_path(), " version ", vformat("%.3f", _version),
 				" will be updated to ", vformat("%.3f", Terrain3DData::CURRENT_VERSION), " upon save");
 	}
 }
 
+void Terrain3DRegion::set_region_size(const int p_region_size) {
+	if (_region_size > 0 && _region_size != p_region_size) {
+		_modified = true;
+	}
+	_region_size = CLAMP(p_region_size, 64, 2048);
```
Xtarsia: Should this also ensure that the region size saved is a power of 2 as well?

Xtarsia (commented):
I tested importing some hieghtmaps into the demo data, and the color & control maps were correctly left in place.

Everything saved correctly too.

src/terrain_3d.cpp
TokisanGames: That will recalculate collision every frame. Currently it only recalculates the collision or mesh once the target has moved enough distance.

src/terrain_3d_region.cpp
TokisanGames: Might as well cast it to the region_size enum so it's a valid size.

src/terrain_3d.cpp
Xtarsia: Of course. Never mind!

src/terrain_3d_region.cpp
TokisanGames: Actually I learned in C++ `43` is a valid value for `enum RegionSize { 64, 256, 512, 1024 }` because it's an int. No way other than conditionals and assignments to clamp it to the enum values. So I added a check in utils.

PR #767, opened 2025-07-18
Add fade to slope range tool settings when out of range

### LINKED ISSUE #725: Texture Blend/Smooth Brush
### Description

In some situations the blend borders of the overlay texture end up very harsh and jagged.
This is especially prominent when using the slope limiter (see picture of "Control Blend" Debug view after using a 40° slope limit).
It would be very useful to have a texture blend brush to blur the blend mask in those areas.
This would also help with complex intersections of more than 2 textures.

Image

Xtarsia:
Closes #725

example with the color map, painting red with a high slope limit, gives soft edges outside the range when painting.

image

any time the original was "true" will still be 1.0 here, only the false conditions get faded out here.

### DIFF
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -692,18 +692,18 @@ Vector3 Terrain3DData::get_normal(const Vector3 &p_global_position) const {
 	return normal;
 }
 
-bool Terrain3DData::is_in_slope(const Vector3 &p_global_position, const Vector2 &p_slope_range, const bool p_invert) const {
+real_t Terrain3DData::is_in_slope(const Vector3 &p_global_position, const Vector2 &p_slope_range, const bool p_invert) const {
 	// If slope is full range, it's disabled
 	const Vector2 slope_range = CLAMP(p_slope_range, V2_ZERO, Vector2(90.f, 90.f));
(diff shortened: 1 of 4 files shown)


src/terrain_3d_data.cpp
```diff
@@ -692,18 +692,18 @@ Vector3 Terrain3DData::get_normal(const Vector3 &p_global_position) const {
 	return normal;
 }
 
-bool Terrain3DData::is_in_slope(const Vector3 &p_global_position, const Vector2 &p_slope_range, const bool p_invert) const {
+real_t Terrain3DData::is_in_slope(const Vector3 &p_global_position, const Vector2 &p_slope_range, const bool p_invert) const {
 	// If slope is full range, it's disabled
 	const Vector2 slope_range = CLAMP(p_slope_range, V2_ZERO, Vector2(90.f, 90.f));
 	if (slope_range.y - slope_range.x > 89.99f) {
-		return true;
+		return real_t(1.f);
```
TokisanGames, line 699: Casting float literals isn't necessary. Nor for the float math in the returns below. They'll be converted automatically for free. It's only necessary for some templates like Lerp or MIN/MAX because the linter is particular about the exact type.

PR #773, opened 2025-07-22
Expand smoothing

TokisanGames:
* Fixes Smooth not working with gradient - point select interfered with smoothing
* Fixes (after the above fix) Select Gradient tool -> hold Shift -> brush -> let go of Shift -> see hole created from divide by zero
* Shift smooth key now works in multiple modes:
  * Sculpt, Holes, Instancer tools -> Average heights
  * Texture tools -> Average blend (includes Alt mode, difference visible in debug_views_control_blend)
  * Color -> Average color
  * Roughness -> Average wetness
* Adds missing decal changes on various smooth and invert modes. Every tool mode should change now. Also reviewed and updated some colors.
* Updated docs to list all keys per tool

Partial implementation #725.


src/terrain_3d_editor.cpp
```diff
@@ -688,6 +707,86 @@ void Terrain3DEditor::_apply_undo(const Dictionary &p_data) {
...
+	Color down = data->get_pixel(TYPE_COLOR, down_position);
+	if (std::isnan(down.r)) {
+		down = COLOR_WHITE;
+	}
+	return Color(
+			(p_base.r + left.r + right.r + up.r + down.r) * 0.2f,
+			(p_base.g + left.g + right.g + up.g + down.g) * 0.2f,
+			(p_base.b + left.b + right.b + up.b + down.b) * 0.2f,
+			1.f);
+}
```
Xtarsia, line 802: This didnt quite sit right with me. So I did a little testing, and converting from srgb to linear, averaging, and then back to srgb stopped things from devolving to black.

Current | Color space shifted
image

Godot Color has this built in, so trivial to implement:

```c++
	}
	Color down = data->get_pixel(TYPE_COLOR, down_position).srgb_to_linear();
	if (std::isnan(down.r)) {
		down = COLOR_WHITE;
	}
	Color base = p_base.srgb_to_linear();
	return Color(
			(base.r + left.r + right.r + up.r + down.r) * 0.2f,
			(base.g + left.g + right.g + up.g + down.g) * 0.2f,
			(base.b + left.b + right.b + up.b + down.b) * 0.2f,
			1.f).linear_to_srgb();
}
```

src/terrain_3d_editor.cpp
```diff
@@ -688,6 +707,86 @@ void Terrain3DEditor::_apply_undo(const Dictionary &p_data) {
...
+	left = std::isnan(pixel.r) ? p_nan_val : pixel[index];
+	pixel = data->get_pixel(map_type, right_position);
+	right = std::isnan(pixel.r) ? p_nan_val : pixel[index];
+	pixel = data->get_pixel(map_type, up_position);
+	up = std::isnan(pixel.r) ? p_nan_val : pixel[index];
+	pixel = data->get_pixel(map_type, down_position);
+	down = std::isnan(pixel.r) ? p_nan_val : pixel[index];
+
+	if (p_mode == AVG_BLEND) {
+		return real_t(get_blend(p_base) + get_blend(left) + get_blend(right) + get_blend(up) + get_blend(down)) * 0.2f;
```
Xtarsia: In practice this can often be a bit strange, it works well when feathering a spray texture over a base, but if there are many textures in an area it started to feel like the result didnt make sense.

I think it may be better served as a lerp to 0.5 without any need to do the adjacent lookups - or even call this function when "averaging blend".

Xtarsia (commented):
Works well, I actually found myself wanting to use this feature the other day, and was disapointed when i couldnt.

Everything looks good, pending my two suggestions.

src/terrain_3d_editor.cpp
TokisanGames: Wow, much better!

TokisanGames: I'm not able to reproduce this. Whether blending in linear or srgb, the middle always blends towards black. I added it in a separate commit. Did I do it right?
{77D8ADC9-09D8-4CD9-B5E5-D45444B07ABD}

Xtarsia: I just built and tested, and yeah it is correct. Smoothing without the conversion degenerates to black very quickly.

The image above was with the overlay blend method included.

PR #778, opened 2025-08-01
fix - "Channel Packer" menu is hard to see in light UI theme

### LINKED ISSUE #777: Channel Packer menu's text is hard to read when using light theme
### Terrain3D version

v1.0.1

### System information

Godot v4.5.beta3 - macOS Sequoia (15.6.0) - Multi-window, 2 monitors - Metal (Forward+) - integrated Apple M4 Pro (Apple9) - Apple M4 Pro (14 threads) - 48.00 GiB memory

Not applicable

#### Description

The text such as "Albedo texture" is hard to see if using light theme.

Image

#### Possible Reason

The reason could be the fixed background colour.

https://github.com/TokisanGames/Terrain3D/blob/8307e97e957e6adb103a846b105651978968d268/project/addons/terrain_3d/menu/channel_packer.tscn#L3-L4

#### Possible Solution

Use the accent colour (got from current editor theme by code) could be a solution.

If this solution looks good to you, I am happy to fix this issue by myself.

### Logs

Not Applicable.

OzelotVanilla:
Changed fixed background colour to current editor theme colour, controlled by script.

The `tscn` file is changed because Godot reassign the value if there is a script setting that. Currently, the value is assigned in dark mode to keep consistency.

Fixes #777.

Effect now:

image

image

### DIFF
--- a/project/addons/terrain_3d/menu/channel_packer.gd
+++ b/project/addons/terrain_3d/menu/channel_packer.gd
@@ -115,6 +115,9 @@ func pack_textures_popup() -> void:
 
 	(window.find_child("PackButton") as Button).pressed.connect(_on_pack_button_pressed)
 
+	# This method will also be called if the `window` is inited and the editor's theme had changed.
+	prepare_theme()
+
 
 func _on_close_requested() -> void:
 	last_file_selected_fn = no_op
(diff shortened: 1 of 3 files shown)


project/addons/terrain_3d/menu/channel_packer.gd
```diff
@@ -463,3 +466,26 @@ func _pack_textures(p_rgb_image: Image, p_a_image: Image, p_dst_path: String, p_
...
+	var alb_nrm_stylebox: StyleBoxFlat = window.find_child("AlbedoHeightPanel").get_theme_stylebox("panel")
+	alb_nrm_stylebox.bg_color = editor_theme.get_color("base_color", "Editor")
+	alb_nrm_stylebox.border_color = editor_theme.get_color("contrast_color_1", "Editor")
+
+	window.print_tree_pretty()
+	var img_drop_box_stylebox: StyleBoxFlat = window.get_node(
+		"PanelContainer/MarginContainer/VBoxContainer/AlbedoHeightPanel/MarginContainer/HBoxContainer/AlbedoVBox/MarginContainer/Panel"
+	).get_theme_stylebox("panel")
+	# The following line seems not setting the colour.
+	img_drop_box_stylebox.bg_color = editor_theme.get_color("dark_color_1", "Editor")
```
OzelotVanilla: This line seems not being able to set the background colour of the image drop box.

project/addons/terrain_3d/menu/channel_packer.gd
```diff
@@ -463,3 +466,26 @@ func _pack_textures(p_rgb_image: Image, p_a_image: Image, p_dst_path: String, p_
...
+	var editor_theme := EditorInterface.get_editor_theme()
+
+	var main_panel_stylebox: StyleBoxFlat = window.find_child("PanelContainer").get_theme_stylebox("panel")
+	main_panel_stylebox.bg_color = editor_theme.get_color("background", "Editor")
+
+	var alb_nrm_stylebox: StyleBoxFlat = window.find_child("AlbedoHeightPanel").get_theme_stylebox("panel")
+	alb_nrm_stylebox.bg_color = editor_theme.get_color("base_color", "Editor")
+	alb_nrm_stylebox.border_color = editor_theme.get_color("contrast_color_1", "Editor")
+
+	window.print_tree_pretty()
```
TokisanGames: We don't want to leave debugging code in PRs.

TokisanGames (commented):
Thanks for taking a crack at this.

AcceptDialog introduces its own theming elements to Window that the latter doesn't have. ChannelPacker uses Window so doesn't get those changes. However the previous solution and this solution aren't ideal. The scene should not have colored styleboxes in it, because it will often be wrong. I found the best solution, which seems to be what AcceptDialog does is to just take the base_color from the theme. 

I have made changes. Please test them on your system.

OzelotVanilla:
I have it tested and found your solution works well, thank you !

OzelotVanilla:
But still the window seems to be too small for showing the text and checkbox (maybe it is a good idea that I check it later with different resolution, and if it is a bug, then make it fixed in another PR) ?

TokisanGames:
The weird layout is from macos. It doesn't respect the sizing and margins we put in to the window. Other systems don't look like that. Maybe it's hdpi. Separate PR, but I don't have a Mac to develop on.

PR #780, opened 2025-08-13
Allow greater UV scale

Joy-less:
See https://github.com/TokisanGames/Terrain3D/discussions/628

Not sure if I've missed a change to another place as well, but this should allow you to increase a texture's UV scale above 2.0 without changing the slider.

### DIFF
--- a/src/terrain_3d_texture_asset.cpp
+++ b/src/terrain_3d_texture_asset.cpp
@@ -124,13 +124,13 @@ void Terrain3DTextureAsset::set_ao_strength(const real_t p_ao_strength) {
 }
 
 void Terrain3DTextureAsset::set_roughness(const real_t p_roughness) {
-	_roughness = CLAMP(p_roughness, -1.f, 1.0f);
+	_roughness = CLAMP(p_roughness, -1.0f, 1.0f);
 	LOG(INFO, "Setting roughness modifier: ", _roughness);
 	emit_signal("setting_changed");
 }
 
 void Terrain3DTextureAsset::set_uv_scale(const real_t p_scale) {
-	_uv_scale = CLAMP(p_scale, .001f, 2.f);
+	_uv_scale = CLAMP(p_scale, 0.001f, 100.0f);
 	LOG(INFO, "Setting uv_scale: ", _uv_scale);
 	emit_signal("setting_changed");
 }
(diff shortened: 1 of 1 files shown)


src/terrain_3d_texture_asset.cpp
```diff
@@ -124,13 +124,13 @@ void Terrain3DTextureAsset::set_ao_strength(const real_t p_ao_strength) {
...
 void Terrain3DTextureAsset::set_roughness(const real_t p_roughness) {
-	_roughness = CLAMP(p_roughness, -1.f, 1.0f);
+	_roughness = CLAMP(p_roughness, -1.0f, 1.0f);
 	LOG(INFO, "Setting roughness modifier: ", _roughness);
 	emit_signal("setting_changed");
 }
 
 void Terrain3DTextureAsset::set_uv_scale(const real_t p_scale) {
-	_uv_scale = CLAMP(p_scale, .001f, 2.f);
+	_uv_scale = MIN(p_scale, 0.001f);
```
TokisanGames: I want a reasonable upper bound that gives results at every value. It could be 10 or 100, but we don't need 1 quadrillion.

Joy-less: OK. From what I can tell, it always makes the texture the same size regardless of the resolution. In that case, about 50 should probably work for almost every use case (although it would skew the slider).
I think if you want to preserve the slider's integrity then setting the max to 10 should also be fine.

TokisanGames: The range in bind_methods changes the slider. Allow_greater allows them to type in anything larger. Clamp here will filter that typed in value. This will allow the slider to be reasonable for most(0-2), yet special cases can go larger.

Joy-less: I chose 100.0f because it seems safe and simple. Let me know if you would like to change it.

TokisanGames:
This is fine, but please squash your commits. https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase

src/terrain_3d_texture_asset.cpp
TokisanGames: The value shouldn't be chosen just because it's a round number. It should be chosen based on need. You wanted a number greater than 2. Why? What size of texture are you using and why do you need the details scaled down to such a miniscule size, 1000x smaller than the default? Did you test to ensure 100 still renders the texture properly, and gives usable results that are meaningfully different from 50 or 10?

TokisanGames:
I'm willing to merge this in, but I don't see the point of it. You didn't answer my questions. I built it and at 20+ on the demo textures, the grass turns solid green. Yet if the camera is very close to the ground you can see a repeating pattern of the grass, which looks like an artifact. Why not just use a 1px green texture? or a 1px white texture with a green albedo tint. You'd get a better result with less memory consumption. What size of textures are you using, and what does your terrain look like that requires a >2 uv scale?

Joy-less:
image

@TokisanGames This is the snow texture in my game at a UV scale of 2.0 (the maximum). The snow texture is 32x32. Sorry about your questions.

TokisanGames:
Alright, thank you.

PR #784, opened 2025-08-15
Replace checkboxes with buttons

### LINKED ISSUE #783: Use @export_tool_button in importer.gd
### Description

In Godot 4.4 there is a [new exported variable annotation](https://docs.godotengine.org/en/4.4/tutorials/scripting/gdscript/gdscript_exports.html#export-tool-button) @export_tool_button which allows to create a button to perform actions. Is a bit more expressive than a checkbox.

Image

The only drawback is to make the gdscript incompatible with Godot < 4.4

bkmeneguello:
fixes #783


TokisanGames:
Thanks for the effort. I don't particularly like the look of those buttons in the inspector so haven't done this. ie, At the top where they're all different sizes, and they have a bullhorn icon, are tacky looking. I'll merge it in for now, but we'll be replacing these scripts eventually.

PR #788, opened 2025-08-25
Redesign update_mmis()

TokisanGames:
Now update_mmis() is safe to use multiple times per frame.
* Queues all options to be processed once during `RS.frame_pre_draw` (post _process, before the frame is rendered).
* Update_mmis() has more options that allow queuing:
  * A full destroy and rebuild all
  * All meshes, all regions
  * All meshes in one region
  * All regions for one mesh
  * Specific region and mesh
* Queue de-duplicates, so if a user flags mesh 52 for all regions, then later requests mesh 52, region 3,4, this is ignored since it's already included.


src/terrain_3d_instancer.cpp
```diff
@@ -8,207 +8,266 @@
...
+	if (p_mesh_id < 0 || p_mesh_id > Terrain3DAssets::MAX_MESHES) {
+		LOG(ERROR, "p_mesh_id is out of bounds");
+		return;
+	}
+	const Terrain3DRegion *&region = p_region;
+	const int &mesh_id = p_mesh_id;
+	const Vector2i &region_loc = region->get_location();
+	const Dictionary &mesh_inst_dict = region->get_instances();
+
+	// Verify mesh id is valid and has some meshes
```
TokisanGames, line 108: No changes from here.....

src/terrain_3d_instancer.cpp
```diff
@@ -8,207 +8,266 @@
...
+		}
+		if (ma->get_shadow_impostor() > 0) {
+			Vector2i mesh_key(mesh_id, Terrain3DMeshAsset::SHADOW_LOD_ID);
+			CellMMIDict &cell_mmi_dict = mesh_mmi_dict[mesh_key];
+			MultiMeshInstance3D *mmi = cell_mmi_dict[cell];
+			if (mmi) {
+				mmi->set_custom_aabb(mmi_custom_aabb);
+			}
 		}
 	}
```
TokisanGames, line 272: ... until here.

PR #790, opened 2025-09-03
Fix-crash-on-showing-long-names-in-asset-dock

aidandavey:
As mentioned on Discord, showing names in name_label longer than 16 characters will trigger a TextServer error. This PR fixes by simply truncating the string and displaying it in full on a tooltip instead.


TokisanGames:
Nice find. `main` has a problem now with OOTA, so it's fine to just develop and test against our version in OOTA. Do you think it's a problem with the length? Or a problem trying to display it in the small box?

This has fixed the problem, but let's look at the cosmetics a bit now. Please consider and test with larger and smaller thumbnails. If larger it would be best to display the full length we allow. If smaller, there's a minimum size it uses to hide the label altogether. But it would be good if the tool tip worked even when small. We should probably also put a max limit on the MeshAsset name length in the C++ of like 40 or 32 characters.

At some of the thumbnail sizes, long names cause the thumbnail to get wider. Maybe an alternate would be to change the label  overflow to clip or replace with elipses. Maybe that will fix both costmetic and crashing issues?

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -854,16 +869,14 @@ class ListEntry extends MarginContainer:
...
+					name_label.visible = false
 				else:
-					name_label.text = (resource as Terrain3DMeshAsset).get_name()
+					name_label.visible = true
 				is_hovered = true
 				emit_signal("hovered")
 				queue_redraw()
 			NOTIFICATION_MOUSE_EXIT:
-				name_label.text = ""
+				name_label.visible = false
```
TokisanGames, line 886: I believe I used text = "" instead of visible because visible removes it from the UI, causing the lines in the vbox to adjust. text = "" keeps it there and lets the count label to stay on the bottom of the thumbnail.

aidandavey:
> Do you think it's a problem with the length? Or a problem trying to display it in the small box?
> 

On further testing, it's actually the font_size change on draw notification that is causing the crash.

Once you get over a certain nb characters, the Label size changes, it's parent Rect is larger, which means the font gets larger to suit, then the Label gets larger, Rect gets larger, then....

I'll move this to on_slider_changed(), and clamp the max font size for good measure. 

I've also attended to the cosmetic changes you mentioned.

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: This is handled in the PR by removing the vbox, the count label no longer moves when name label is hidden

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -854,16 +873,14 @@ class ListEntry extends MarginContainer:
 					draw_style_box(focus_style, rect)
 			NOTIFICATION_MOUSE_ENTER:
 				if size.x < 86 or not resource:
```
TokisanGames: Now we no longer go lower than 90, so this *shouldn't* be needed.

But somehow it happened here on the right. Even with the scene slider min set to 90, on all docks, except for popped out window, the smallest sizes (90-91?) cause the last thumbnail to shrink and the icons to wrap. Make the size larger then smaller, and it should scrunch up like this. The label then disappears because of this code. 

It would be good to know why this is happening, but also see the other comment about separation.

{059A95F4-A419-4BF4-BC09-A8435CFB8187}


Sometimes I got it to occur on _not_ the last column!? (left and right)

{87F2079A-6913-4DA1-87B3-72839781CE77}

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -768,19 +781,31 @@ class ListEntry extends MarginContainer:
...
-		name_label.add_theme_constant_override("shadow_offset_y", 1.)
-		name_label.add_theme_font_size_override("font_size", 15)
-		label_rows.add_child(name_label, true)
+		name_label.add_theme_font_size_override("font_size", 14)
+		name_label.visible = false
+		name_label.autowrap_mode = TextServer.AUTOWRAP_OFF
+		name_label.text_overrun_behavior = TextServer.OVERRUN_TRIM_ELLIPSIS
+	
+		name_label.text = get_resource_name()
+		#name_label.mouse_filter = Control.MOUSE_FILTER_PASS
```
TokisanGames: If this is unnecessary, it can be removed.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -634,14 +641,20 @@ class ListContainer extends Container:
 		var separation: float = 4
```
TokisanGames: The asset dock _meshes_ gives the impression that it is a little slow and sluggish in OOTA. One reason is when clicking meshes it seems to be counting and generating the thumbnails. Now that slowness has been established in my mind, when moving the mouse over each, it "feels" slow. I think this is because a) it's possibly regenerating a thumbnail (though I think it is already skipping it once made), and b) I think it's purely my imagination due to the fact that my mouse is moving 4 (or 8?) pixels across the gap between thumbnails, doing nothing, which registers in my brain as lag. 

I found that reducing this value makes hovering seem a little snappier. Also going to 2 seems to fix the icon wrapping issue. Perhaps this is a bandaid and there's another issue there, but we don't need to spend too much time on it. I think 1 or 2 are good. 0 is best for meshes, but makes the textures too close. Add a decimal to the float literal.

4.0

{ACDD6F00-570F-4D69-B419-2E64CCE3B6E2}

2.0

{66DE64A3-5955-46C5-8500-513C5305911F}

1.0

{7060DC50-CBE8-44D5-992E-CCF6B29E43FB}

TokisanGames (changes requested):
It works well and fixes the crashes. 🥳 
Now just some cosmetic issues.

* Increasing the minimum size is fine. The slider in asset_dock.tscn also needs it's minimum set to 90 or the slider will be able to access a dead zone between 66 and 90. I think you could also increase the max from 230 to 256 in the tscn, and in the code.
* The only thing I don't like is the counter, and the overlay text are grey and washed out looking like they're behind another layer, shown below. I'd like them a crisp white with black drop shadow as before.
* Though it handles long names well, let's put in a max cap on the name in C++ in both the texture and mesh assets. Anywhere under 100 is fine, just so we don't trigger other weird bugs in the godot text server in the future.

{C9425D2D-DBAE-48D3-93E6-B473AF37E5DE}

* Very minor detail, but when you click the X, the thumbnail goes away, but the label remains until mouse out. Then it no longer appears.

{A74E8D58-4910-42B2-B23C-9B8A7E2314EA}

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: The size of the tile is adjusted by the separation, for a slider setting of 90 the size is actually 86*86. Sometimes a rounding error is introduced, which puts the size at 85.99999 making size.x < 86 == true

I agree this can go though now, it's not needed.

project/addons/terrain_3d/src/asset_dock.gd
aidandavey: Here we could get rid of the separation entirely and put the controls in a secondary MarginContainer. There will still be a visual gap, but the mouse will go straight from one tile to another.

aidandavey: For now I have adjusted the separation to 2.0 and will build a new tile scene in a future PR if needs be.

TokisanGames:
UIs are super challenging to work with and get right. 

* I didn't like the buttons are now 'centered' in a narrow column, instead of being 'justified', extending to the edges. I added the button spacer back in.

* Custom_minimum_size 90 caused the smallest size thumbnail to lose the gap. For a separation of 4, min size should be 86 to avoid the compressed gap and wrapped buttons. Also works with separation 2.

* Found and fixed a bug related to naming New Texture to "assets" when they are saved in the scene.

Great work, thank you.

PR #792, opened 2025-09-08
Asset Dock - Add search box

aidandavey:
Adds a search box to the asset dock panel to filter the assets as mentioned in #501

Additionally thumbnails will only be generated once when first loading a mesh asset in the dock - this can be separated out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Terrain3D
 project/addons/terrain_3d/bin/
+project/addons/explore-editor-theme/
 project/_dev/
 project/_export/
 project/_tests/
--- a/src/terrain_3d_assets.cpp
+++ b/src/terrain_3d_assets.cpp
@@ -15,7 +15,7 @@
 ///////////////////////////
 
 void Terrain3DAssets::_swap_ids(const AssetType p_type, const int p_src_id, const int p_dst_id) {
-	LOG(INFO, "Swapping asset id: ", p_src_id, " and id: ", p_dst_id);
+	LOG(INFO, "Swapping asset ID: ", p_src_id, " and ID: ", p_dst_id);
 	Array list;
 	switch (p_type) {
 		case TYPE_TEXTURE:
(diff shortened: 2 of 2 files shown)


TokisanGames:
I used it on the other PR, and had these notes. I like it, and it works well.
* Capitalize Search
* Clicking anywhere outside of the input box should cause it to lose focus. Right now only clicking the slider works, so it's very sticky and won't let go of focus.
* When in sidebar or window mode, I was thinking the search box should go to a new line just like the slider does. In window mode, the pin should be in the top right position.

TokisanGames (commented):
Really great addition. I rebased and made the following changes:
* Added a bit more margin on the right of the search button.
* Search button changes to an X if there's something in it and is always visible.
* X clears the box, applies it, and _does not grab focus_, that latter is easily changed, but I suspect most clicking the button will want to clear and move on, not clear and retype, which could be done by double clicking the text instead of the button.
* Moved the search box to the top as it will be used more than anything else on the panel
* Removed unused `var search_icon`.
* Search result is trimmed of escape characters like newline.
* Tweaked the TextEdit settings:
  * caret blink
  * no multiple carets
  * no emojis
  * fixed size so a lot of text won't expand the UI
* Consolidated button layouts from 3 to wide and vertical.

I only have one remaining issue I haven't found the cause for yet. The search box loses focus when:
* The engine is just started
* Terrain3D is selected
* Typing in the second character into the search

After that first time, or if Terrain3D isn't selected, or any other case, it types in the search box just fine.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -250,6 +266,29 @@ func update_thumbnails() -> void:
...
+	if len > 0:
+		search_box.set_caret_column(len)
+		search_button.icon = get_theme_icon("Close", "EditorIcons")
+	else:
+		search_button.icon = get_theme_icon("Search", "EditorIcons")
+		
+	mesh_list.search_text = search_box.text
+	texture_list.search_text = search_box.text
+	_current_list.update_asset_list()
+	_current_list.set_selected_id(0)
```
TokisanGames: > set_selected_id(0)

This causes the first load, select terrain3d, can only enter one character. Commenting this out fixes that issue, but the selection is hidden/not right. Calling it deferred doesn't help.

Also _on_search_text_changed() is called twice, only from the first character on the first load.

Looks like I fixed it by telling set_selected_id() to skip if selected == p_id

aidandavey: I used set_selected_id(0) to handle the case that the previous selection was not in the search results.

If that's still handled then great, otherwise perhaps we can call search_box.grab_focus() after set_selected_id(0)

aidandavey: Just tested and it looks good

TokisanGames:
Found some more issues:

1. Clicking Textures / Meshes no longer switches the tools. Allowing the tool switching revealed a larger issue. Changing the tools also calls _on_meshes_pressed() or _on_textures_pressed(), and causes the search box handler to get two signal calls strangely. Preventing these two functions from running if already on that list resolves the first click fail, double signal calls, and switches the tools.

Godot has definite bugs if some UI events occur twice in the same frame. That's why _on_meshes_pressed has an _updating_list gate. Longer term, the asset dock needs to go the way of update_mmis() is headed, where all updates are queued and occur only once per frame.

2. Type in 'g', it shows grass is selected, but it paints Rock. That seems to have been fixed by the work done above.

I added some UI debugging enabled in editor_plugin.gd:debug = 1 or 2.

So far it looks ok, but needs further testing of edge cases: Starting from other tools selected, before searching, or selecting, filtering while unselecting terrain3d and reselecting, changing scenes and ensuring what is selected is still valid, etc.

TokisanGames:
Here's another subtle bug:

`entries.size()`: "Add new" is the final entry only when search box is blank, and not when it has data.

This affects selection, `get_selected_resource_id()` needs to account for the difference, and deleting an asset from the filtered list deletes the wrong one. This actually gets the IDs out of order, non-contiguous, and can trigger a crash in the C++ _swap_ids().

I'm working on this one.

project/addons/terrain_3d/src/asset_dock.gd
```diff
@@ -522,6 +582,10 @@ class ListContainer extends Container:
 		var id: int = entries.size()
```
TokisanGames: This is the cause of the latest bug that led to corruption and crashing. Below _on_resource_changed.bind(id) takes the filtered size as the ID instead of the resource ID. When deleting an item from the filtered list, it deleted the wrong ID. That led the asset list getting out order or adding nulls in it and the C++ wasn't prepared for that. I've now fixed those issues as well.

TokisanGames:
Ok, after another day of rest and testing, and testing in OOTA, I think this is good to merge finally.
Nice feature. Thanks for the addition.

PR #805, opened 2025-09-23
Brush updates

### LINKED ISSUE #797: Brush values from exr file should not be clamped between 0 and 1
### Terrain3D version

v.1.0.1

### System information

Godot v4.4.1.stable - Windows 10 (build 19045) - Multi-window, 2 monitors - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4080 (NVIDIA; 32.0.15.6094) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 threads)

Yes

### Issue description

I recently started creating my own exr brushes and discovered that values below 0 and higher than 1 get clamped.
When sculpting with these inside blender everything works as expected, here the results as reference.

Image

In godot it looks like this
Image

It feels like a bug, as with a format that supports values outside 0 and 1 I expected it to work, especially when I saw it in the preview, that it shows negative values. Otherwise it could be seen as a feature request too. 

Here are the attached exr files, tested in the demo project.
[clamping_issue_brushes.zip](https://github.com/user-attachments/files/22318990/clamping_issue_brushes.zip)

### Logs

_No response_

### LINKED ISSUE #798: Bad handling of terrain brushes with low value range, showing no preview
### Terrain3D version

v1.0.1

### System information

Godot v4.4.1.stable - Windows 10 (build 19045) - Multi-window, 2 monitors - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4080 (NVIDIA; 32.0.15.6094) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 threads)

Yes

### Issue description

This might be related to https://github.com/TokisanGames/Terrain3D/issues/797

When I created the following brush (the low range is the original), I wasn't able to see a preview.
After some experiments I found out, that the issue came from the unusual low values.
I then increased the range and now I can see the preview.

While this feels more like a minor issue with an easy workaround, I am thinking of maybe something like detecting images with low range data, printing a warning for the user / creator, or for example auto remapping the range in the preview to show it's contents.

Image

https://github.com/user-attachments/assets/3df45c61-562a-49f2-8949-4c70795fabc2

[missing_preview_brush.zip](https://github.com/user-attachments/files/22319303/missing_preview_brush.zip)

### Logs

_No response_

### LINKED ISSUE #799: Broken / Corrupted exr files mess up the brush UI
### Terrain3D version

v.1.0.1

### System information

Godot v4.4.1.stable - Windows 10 (build 19045) - Multi-window, 2 monitors - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4080 (NVIDIA; 32.0.15.6094) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 threads)

Yes

### Issue description

I somehow managed to create a seemingly broken / corrupted exr file, which is messing up the brush UI.
First, the exr file isn't completely broken, opening the file in gimp throws an error too, but I can at least open it in krita.

I attached some screenshots, what this broken brush does with the UI, which looks different depending on the name, where it lands, sorting by the alphabet.
Image
Image

I attached a log string and also a working file, which I fixed by opening / re-exporting in krita.

[broken_brush.zip](https://github.com/user-attachments/files/22319357/broken_brush.zip)

### Logs

ERROR: modules/tinyexr/image_loader_tinyexr.cpp:136 - Condition "idxR == -1" is true. Returning: ERR_FILE_CORRUPT
ERROR: core/io/image_loader.cpp:100 - Error loading image: 'res://addons/terrain_3d/brushes/brush_broken.exr'.
ERROR: Failed to load image. Error 16
ERROR: res://addons/terrain_3d/src/tool_settings.gd:278 - Cannot call method 'duplicate' on a null value.

### LINKED ISSUE #800: Painting Colors with strength values higher than 100 produce strange results
### Terrain3D version

v1.0.1

### System information

Godot v4.4.1.stable - Windows 10 (build 19045) - Multi-window, 2 monitors - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4080 (NVIDIA; 32.0.15.6094) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 threads)

Yes

### Issue description

If I paint colors with strength values higher than 100 % it gives strange results, it starts to remove colors in certain areas if painting over.
This is really weird and should not happen, one quick solution would be to clamp the max strength too 100.
https://github.com/user-attachments/assets/4de230b3-3e28-486d-b813-02d7b27ebfd1

Image

### Logs

_No response_

aidandavey:
Some changes to brushes as suggested, commits reference the issues

------------
Fixes #800
Fixes #799
Fixes #797
Fixes #798

### DIFF
--- a/project/addons/terrain_3d/src/tool_settings.gd
+++ b/project/addons/terrain_3d/src/tool_settings.gd
@@ -280,6 +280,9 @@ func add_brushes(p_parent: Control) -> void:
 			if !dir.current_is_dir() and file_name.ends_with(".exr"):
 				var img: Image = Image.load_from_file(BRUSH_PATH + "/" + file_name)
 				if img:
+					var value_range: Vector2 = Terrain3DUtil.get_min_max(img)
+					if value_range.y - value_range.x < 0.333:
+						push_warning("'%s' has a low value range and may not be visible in the brush gallery or cursor" % file_name)
 					var thumbimg: Image = img.duplicate()
 					img.convert(Image.FORMAT_RF)
 


src/terrain_3d_editor.cpp
```diff
@@ -223,7 +223,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
 			// Start brushing on the map
 			real_t brush_alpha = brush_image->get_pixelv(brush_pixel_position).r;
 			brush_alpha = real_t(Math::pow(double(brush_alpha), double(gamma)));
-			brush_alpha = std::isnan(brush_alpha) || std::isnan(brush_alpha) ? 0.f : CLAMP(brush_alpha, 0.f, 1.f);
+			brush_alpha = std::isnan(brush_alpha) || std::isnan(brush_alpha) ? 0.f : CLAMP(brush_alpha, -1.f, 1.f);
```
TokisanGames: This does not fully address #797 which requested no clamp. I have confirmed blender sculpting and UE have no clamp when using exr brushes, so we should remove the clamp on brush_alpha.

src/terrain_3d_editor.cpp
```diff
@@ -495,6 +495,7 @@ void Terrain3DEditor::_operate_map(const Vector3 &p_global_position, const real_
 				}
 				switch (_tool) {
 					case COLOR:
+						strength = CLAMP(strength, 0.01f, 1.0f);
```
TokisanGames: While this addressed color, wetness was still affected. I reveiewed all operations and saw every single one has it's own individual `clamp(brush_alpha * strength)`, often with different values, except raise/lower. So I'm moving to that, and unrestricting brush_alpha.

src/terrain_3d_editor.cpp
TokisanGames: `std::isnan(brush_alpha) || std::isnan(brush_alpha) ` is redundant. Weird that it slipped by at least 3 of us devs.

PR #811, opened 2025-10-02
Scroll to selected asset

aidandavey:
As discussed on Discord, scrolls the asset dock on selection to show the selected asset


TokisanGames:
Thank you.

PR #813, opened 2025-10-08
Improve Create thumbnails, update_mmis()

TokisanGames:
* Generate thumbnails when settings changed
  * Still can't do it on material settings changed, since materials don't emit Resource::changed.
* Fixes generated mesh settings not updating the display
* Removes multiple signal calls to update_mmis
* Moves scroll asset dock to frame_pre_draw instead of physics frame, which fixes not being in the tree (test OOTA)
* Fixes edge case bugs: 
  * min thumbnail size 1x1 - min viewport is 2x2. 
  * Clamps must be MAX_TEXTURE/MESH_SIZE -1.


aidandavey:
Tested this and it works great, feels much faster and reliable.

PR #815, opened 2025-10-12
Docs: Updated dynamic collision reference in Troubleshooting

jamonholmgren:
Since PR #278 has already landed, this section should point to the documentation for dynamic collision generation rather than saying it is coming later.


TokisanGames:
Thank you.

PR #831, opened 2025-10-23
Limit godot namespace to GDExtension

Jackerty:
This is first part of a series of commits to make GDNative build easier (but not fully supported).

GDExtension has everything under `godot` namespace. In engine, no such namespace exists. Since it is an error in C++ to use namespace which is not defined preprocessing is used to detect should godot namespace be used.

To avoid having to write this check to every header "using.inc" is added. This way using.inc only needs to be added every header without worrying about details. (I felt that .inc was better extension but I don't particularly care either way)

register_types is specially handled as gdextension_interface.h gives information in the GDExtension which is given modules/register_module_types.h in engine.

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -3,7 +3,10 @@
 #ifndef CONSTANTS_CLASS_H
 #define CONSTANTS_CLASS_H
 
+// GDExtension uses the godot namespace, custom modules do not.
+#if defined(GDEXTENSION) && !defined(GODOT_MODULE)
 using namespace godot;
+#endif
 
 // Engine Shortcuts
 #define RS RenderingServer::get_singleton()
--- a/src/generated_texture.h
+++ b/src/generated_texture.h
@@ -7,8 +7,6 @@
 
 #include "constants.h"
 
-using namespace godot;
-
 class GeneratedTexture {
 	CLASS_NAME_STATIC("Terrain3DGenTex");
 
--- a/src/logger.h
+++ b/src/logger.h
@@ -5,9 +5,8 @@
 
 #include <godot_cpp/variant/utility_functions.hpp>
 
+#include "constants.h"
 #include "terrain_3d.h"
-
-using namespace godot;
 
 /**
  * Prints warnings, errors, and messages to the console.
--- a/src/target_node_3d.h
+++ b/src/target_node_3d.h
@@ -5,7 +5,7 @@
 
 #include <godot_cpp/classes/node3d.hpp>
 
-using namespace godot;
+#include "constants.h"
 
 class TargetNode3D {
 	CLASS_NAME_STATIC("Terrain3DTargetNode3D");
--- a/src/terrain_3d_asset_resource.h
+++ b/src/terrain_3d_asset_resource.h
@@ -7,7 +7,6 @@
 
 #include "constants.h"
 
-using namespace godot;
 class Terrain3DAssets;
 
 // Parent class of Terrain3DMeshAsset and Terrain3DTextureAsset
--- a/src/terrain_3d_collision.h
+++ b/src/terrain_3d_collision.h
@@ -11,8 +11,6 @@
 #include "constants.h"
 #include "terrain_3d_util.h"
 
-using namespace godot;
-
 class Terrain3D;
 
 class Terrain3DCollision : public Object {
--- a/src/terrain_3d_instancer.h
+++ b/src/terrain_3d_instancer.h
@@ -11,8 +11,6 @@
 #include "constants.h"
 #include "terrain_3d_region.h"
 
-using namespace godot;
-
 class Terrain3D;
 class Terrain3DAssets;
 
--- a/src/terrain_3d_material.h
+++ b/src/terrain_3d_material.h
@@ -10,8 +10,6 @@
 
 class Terrain3D;
 
-using namespace godot;
-
 class Terrain3DMaterial : public Resource {
 	GDCLASS(Terrain3DMaterial, Resource);
 	CLASS_NAME();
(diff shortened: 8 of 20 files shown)


src/using.inc
```diff
@@ -0,0 +1,6 @@
+// Since GDExtension uses godot namespace but GDNative does not
+// we use this header to make GDExtension to force godot namespace
+// as global.
+#if defined(GDEXTENSION) && !defined(GD_NATIVE)
+using namespace godot;
+#endif
```
TokisanGames: * `using namespace` is already in constants.h, so is unnecessary in every other file and can be removed except for constants.h.
* using.inc is unnecessary, as this code here should be added to constants.h.
* GD_NATIVE is not a Godot 4 concept and should not be added in this PR in the code or comment.

Jackerty: Shit, I meant GODOT_MODULE, not GD_NATIVE. How did I fuck that one? I even [checked](https://github.com/godotengine/godot/blob/084d5d407e62efcd5be9de44148c5dedce3b9386/modules/SCsub#L13) it.

But, okay, I'm fine with constants.h

Jackerty: Changes made.

Added constants.h to the target_node_3d.h

src/target_node_3d.h
```diff
@@ -4,8 +4,7 @@
 #define TARGET_NODE3D_CLASS_H
 
 #include <godot_cpp/classes/node3d.hpp>
-
-using namespace godot;
+#include <constants.h>
```
TokisanGames: This should be using "", which denotes local files. <> is for libraries.

src/generated_texture.h
```diff
@@ -27,4 +25,4 @@ class GeneratedTexture {
 	RID get_rid() const { return _rid; }
 };
 
-#endif // GENERATEDTEXTURE_CLASS_H
\ No newline at end of file
+#endif // GENERATEDTEXTURE_CLASS_H
```
TokisanGames: Please don't strip my new lines at the end of files. There are a few files you modified.

src/logger.h
```diff
@@ -36,4 +35,4 @@ using namespace godot;
 #define LOG(...)
 #endif
 
-#endif // LOGGER_CLASS_H
\ No newline at end of file
+#endif // LOGGER_CLASS_H
```
TokisanGames: here

src/terrain_3d_asset_resource.h
```diff
@@ -34,4 +33,4 @@ class Terrain3DAssetResource : public Resource {
 	bool _highlighted = false;
 };
 
-#endif // TERRAIN3D_ASSET_RESOURCE_CLASS_H
\ No newline at end of file
+#endif // TERRAIN3D_ASSET_RESOURCE_CLASS_H
```
TokisanGames: here

src/terrain_3d_collision.h
```diff
@@ -107,4 +105,4 @@ inline Vector2i Terrain3DCollision::_snap_to_grid(const Vector3 &p_pos) const {
 			_shape_size;
 }
 
-#endif // TERRAIN3D_COLLISION_CLASS_H
\ No newline at end of file
+#endif // TERRAIN3D_COLLISION_CLASS_H
```
TokisanGames: here

src/terrain_3d_instancer.h
```diff
@@ -108,4 +106,4 @@ inline uint32_t Terrain3DInstancer::_get_density_count(const real_t p_density) {
 	return count;
 }
 
-#endif // TERRAIN3D_INSTANCER_CLASS_H
\ No newline at end of file
+#endif // TERRAIN3D_INSTANCER_CLASS_H
```
TokisanGames: here

src/terrain_3d_material.h
```diff
@@ -170,4 +168,4 @@ class Terrain3DMaterial : public Resource {
 VARIANT_ENUM_CAST(Terrain3DMaterial::WorldBackground);
 VARIANT_ENUM_CAST(Terrain3DMaterial::TextureFiltering);
 
-#endif // TERRAIN3D_MATERIAL_CLASS_H
\ No newline at end of file
+#endif // TERRAIN3D_MATERIAL_CLASS_H
```
TokisanGames: here

src/constants.h
```diff
@@ -3,7 +3,12 @@
 #ifndef CONSTANTS_CLASS_H
 #define CONSTANTS_CLASS_H
 
+// Since GDExtension uses godot namespace but in-engine there is no such
+// namespace. Since it is and error in C++ use are namespace which does not exist
+// preprocessing guard is used to make custom module builders happier.
```
TokisanGames: Let's simplify this to:
// GDExtension uses the godot namespace, modules do not.

src/generated_texture.h
Jackerty: wtf...Does getid or code do it....

src/constants.h
Jackerty: Hmmm.

`// GDExtension uses the godot namespace, custom modules do not.`

Hmmm.

I can agree that my current proposal has too long a sentence in it.  I would still do it, but I'm targeting too much of a noob C++ coder with that? Maybe...

I don't like just modules because that can refer to a lot of other things. (One .cpp .hpp combo for example). I would at least say "custom module". Bigger change to get someone who does not know to google "godot custom module".

TokisanGames: That's fine.

src/generated_texture.h
Jackerty: Looks denoised now. I think, gedit was doing something I didn't expect.

TokisanGames:
Thank you.

Jackerty:
You're welcome :)

Issue #101, opened 2023-07-03
Implement terrain generation


TokisanGames:
Look at procedural terrain generation w/ erosion like in Zylann's HTerrain.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/1ead3c18-7d82-4a9c-93fc-9da195c4f57e)

Probably wait until compute shaders are running. Then look at:

https://github.com/SebLague/Hydraulic-Erosion 

https://github.com/dandrino/terrain-erosion-3-ways

https://discourse.panda3d.org/t/dynamic-terrain-erosion-on-the-gpu/14935

https://search.brave.com/search?q=terrain+erosion+glsl&source=android

https://github.com/jice-nospam/wgen

Auburn offers his whole library in GLSL which could be included as a compute shader based terrain generator, which would depend on https://github.com/TokisanGames/Terrain3D/issues/174.
https://github.com/Auburn/FastNoiseLite/tree/master/GLSL

Also the current worldnoise algorithm can be included and it should be stampable - add a region with the current worldnoise settings.

Tracked in #81

Swivelgames:
Any interest in implementing different types of Noise algorithms? For instance, those that don't require baking in order to simulate erosion?

Inigo Quilez has an [exceptional article](https://iquilezles.org/articles/morenoise/) that implements some clever techniques. I ended up finding a [related video](https://youtu.be/gsJHzBTPG0Y) that summarizes it nicely for those interested.

While it might be out of the scope of this tool, the prospect of this is actually pretty intriguing, especially when you take into account the possibility for procedurally generated LODs at a theoretically "infinite" resolution. But I digress 😅

TokisanGames:
IQ's value noise is already implemented as infinite world noise, and mentioned in the first post. This ticket is for adding terrain generation to the UI.

otto-link:
Hi, I'm developing a C++ library for terrain generation https://github.com/otto-link/HighMap
HighMap is the core solver of the [Hesiod](https://github.com/otto-link/Hesiod) GUI (a desktop app for terrain generation I'm also developing).
If you're interested, I'll be happy to help interface it with Terrain3D.

TokisanGames:
[...]

PR #833, opened 2025-10-24
Add terrain generation tool

aidandavey:
Adding a tool to the Terrain3D menu which generates Terrain3D regions from noise.

I have submitted the PR as I'd like some feedback on the general direction and usability of the tool before committing more time to it. Bugs aside, is this roughly how we want it to work?

A test scene can be found at \project\examples\generation_test.tscn

To use the tool:

1. Select a Terrain3D node
2. Accept that all of your data for all regions will be overwritten (this is a destructive process)
3. From the Terrain3D menu select Generate Terrain
4. Play with the sliders and values
5. Middle mouse to orbit in the visualizer
6. Press Generate

Things to do: 

- [ ] Allow creation of multiple regions, 2x2, 3x3 etc...
- [ ] Add support for every Terrain3D region size
- [ ] Fix visualizer height scale may not match T3D exactly
- [ ] Set sensible defaults
- [ ] Auto-Texture visualizer?
- [ ] Fix slider ranges, show values etc... 
- [ ] Anything else following feedback

### DIFF
--- a/project/addons/terrain_3d/extras/TerrainGeneration/sub_viewport_container.gd
+++ b/project/addons/terrain_3d/extras/TerrainGeneration/sub_viewport_container.gd
@@ -1,0 +1,28 @@
+@tool
+extends SubViewportContainer
+
+var orbit_capture: bool = false
+
+@onready var camera_pivot_yaw: Node3D = $SubViewport/CameraPivotYaw
+@onready var camera_pivot_pitch: Node3D = $SubViewport/CameraPivotYaw/CameraPivotPitch
+@onready var preview_camera_3d: Camera3D = %PreviewCamera3D
+
+
+func _gui_input(event: InputEvent) -> void:
+	if event is InputEventMouseButton:
+		if event.button_index == MOUSE_BUTTON_MIDDLE:
+			orbit_capture = !orbit_capture
+			return
+		elif event.button_index == MOUSE_BUTTON_WHEEL_UP:
+			preview_camera_3d.fov -= 10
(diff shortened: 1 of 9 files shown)


TokisanGames:
- You could link up #101, and it's related to #81. 
- This is a feature for 1.2, so I won't spend much time on it until 1.1 is out.
- It should create a new world to do its preview generation in, and not apply it to the current scene until applied.
- The preview should be able to adopt the assets of the current scene, and specify two textures for the autoshader.
- It needs full access to the noise library. It should basically be a noise library 3D previewer.
- Long term we'll have a GPU workflow and will be using Auburn's GLSL noise library. We could entirely skip the CPU library and do that now with a 16-bit viewport. But we must make sure it works in compatibility mode. If not we'll have to do the CPU version for compatibillity and GPU version for Vulkan. 

It will take the GLSL library, IQ's noise, and erosion in shaders to close #101.

MeteorShower2004:
I think you could get inspired in [HTerrain's in-editor terrain-generation](https://github.com/Zylann/godot_heightmap_plugin/tree/master/addons%2Fzylann.hterrain%2Ftools%2Fgenerator). (this had get mentioned in #101)
Their image format is mostly same to ours, except that it might leave seams in the upper and left edge of region. (edited)
Note that they use multi-pass to generate, and they said that each pass costs a render frame though. 

I also hope that this feature could work during runtime. It would help a lot.
(For custom generation, users can write their part of shader and before generation we "include" that )

PR #834, opened 2025-10-24
Remove dependency on Godot "Editor..." classes.

IndigoBeetle:
* Abstract the UndoRedo functionality out to methods on the plugin class.
 * Change the plugin interface to expect an Object rather than an EditorPlugin derived class to allow alternative implementations outside of the Godot editor.

### DIFF
--- a/project/addons/terrain_3d/src/editor_plugin.gd
+++ b/project/addons/terrain_3d/src/editor_plugin.gd
@@ -477,3 +477,26 @@ func has_setting(p_str: String) -> bool:
 
 func erase_setting(p_str: String) -> void:
 	editor_settings.erase(p_str)
+
+
+## Undo / Redo Functions
+
+
+func create_undo_action(p_action_name: String) -> void:
+	get_undo_redo().create_action(p_action_name, UndoRedo.MERGE_DISABLE, terrain)
+
+
+func add_undo_method(p_method: Callable) -> void:
+	var args := [ p_method.get_object(), p_method.get_method() ]
+	args.append_array(p_method.get_bound_arguments())
+	get_undo_redo().add_undo_method.callv(args)
+
+
+func add_do_method(p_method: Callable) -> void:
+	var args := [ p_method.get_object(), p_method.get_method() ]
+	args.append_array(p_method.get_bound_arguments())
+	get_undo_redo().add_do_method.callv(args)
+
+
+func commit_action(p_execute: bool) -> void:
+	get_undo_redo().commit_action(p_execute)
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -452,20 +452,20 @@ void Terrain3D::set_editor(Terrain3DEditor *p_editor) {
 		LOG(ERROR, "Attempted to set a node queued for deletion");
 		return;
 	}
-	LOG(INFO, "Set Terrain3DEditor: ", p_editor);
+	LOG(INFO, "Setting Terrain3DEditor: ", p_editor);
 	_editor = p_editor;
 	if (_material.is_valid()) {
 		_material->update();
 	}
 }
 
-void Terrain3D::set_plugin(EditorPlugin *p_plugin) {
+void Terrain3D::set_plugin(Object *p_plugin) {
 	if (p_plugin && p_plugin->is_queued_for_deletion()) {
 		LOG(ERROR, "Attempted to set a node queued for deletion");
 		return;
 	}
-	LOG(INFO, "Set EditorPlugin: ", p_plugin);
-	_plugin = p_plugin;
+	LOG(INFO, "Setting Editor Plugin: ", p_plugin);
+	_editor_plugin = p_plugin;
 }
 
 void Terrain3D::set_camera(Camera3D *p_camera) {
@@ -633,8 +633,8 @@ void Terrain3D::set_vertex_spacing(const real_t p_spacing) {
 			_collision->build();
 		}
 	}
-	if (IS_EDITOR && _plugin) {
-		_plugin->call("update_region_grid");
+	if (IS_EDITOR && _editor_plugin) {
+		_editor_plugin->call("update_region_grid");
 	}
 }
 
--- a/src/terrain_3d.h
+++ b/src/terrain_3d.h
@@ -4,10 +4,10 @@
 #define TERRAIN3D_CLASS_H
 
 #include <godot_cpp/classes/camera3d.hpp>
-#include <godot_cpp/classes/editor_plugin.hpp>
 #include <godot_cpp/classes/geometry_instance3d.hpp>
 #include <godot_cpp/classes/mesh.hpp>
 #include <godot_cpp/classes/mesh_instance3d.hpp>
+#include <godot_cpp/classes/object.hpp>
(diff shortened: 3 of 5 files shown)


src/terrain_3d.h
```diff
@@ -4,7 +4,7 @@
 #define TERRAIN3D_CLASS_H
 
 #include <godot_cpp/classes/camera3d.hpp>
-#include <godot_cpp/classes/editor_plugin.hpp>
+#include <godot_cpp/classes/object.hpp>
```
TokisanGames: We use clang-format which should be in your hooks. It will sort this list alphabetically. You can do it manually also.

src/terrain_3d_editor.cpp
```diff
@@ -1,8 +1,8 @@
 // Copyright © 2025 Cory Petkovsek, Roope Palmroos, and Contributors.
 
-#include <godot_cpp/classes/editor_undo_redo_manager.hpp>
 #include <godot_cpp/classes/engine.hpp>
 #include <godot_cpp/classes/time.hpp>
+#include <godot_cpp/variant/callable.hpp>
```
TokisanGames, line 5: Should be sorted.

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -477,3 +477,25 @@ func has_setting(p_str: String) -> bool:
...
 	editor_settings.erase(p_str)
+
+
+func create_undo_action(action_name: String) -> void:
+	get_undo_redo().create_action(action_name, UndoRedo.MERGE_DISABLE, terrain)
+
+
+func add_undo_method(method: Callable) -> void:
+	var args := [method.get_object(), method.get_method()]
+	for arg : Variant in method.get_bound_arguments():
```
TokisanGames: Rather than a for loop, how about
`args.append_array(method.get_bound_arguments())`

Same with add_do_method.

src/terrain_3d_editor.cpp
TokisanGames: update_grid was renamed update_region_grid but wasn't updated. This could be fixed here with has_method removed. If it was, we would have seen the error before. plugin is already validated so just do the LOG, then call update_region_grid(). No if block.

src/terrain_3d.cpp
```diff
@@ -459,12 +459,12 @@ void Terrain3D::set_editor(Terrain3DEditor *p_editor) {
...
 
-void Terrain3D::set_plugin(EditorPlugin *p_plugin) {
+void Terrain3D::set_plugin(Object *p_plugin) {
 	if (p_plugin && p_plugin->is_queued_for_deletion()) {
 		LOG(ERROR, "Attempted to set a node queued for deletion");
 		return;
 	}
-	LOG(INFO, "Set EditorPlugin: ", p_plugin);
+	LOG(INFO, "Set Plugin: ", p_plugin);
 	_plugin = p_plugin;
```
TokisanGames: This should now verify p_plugin is an object we can use for:
* calling update_region_grid() in Terrain3D/Terrain3DEditor (though this should change to a signal)
* calling ui.set_decal_rotation() in Terrain3DEditor (which should change to the UI polling the rotation value)
* handling undo/redo operations

It is sufficient to just move the `has_method("create_undo_action")` here to verify before setting. Then checking if the plugin exists in other functions is enough.

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -477,3 +477,25 @@ func has_setting(p_str: String) -> bool:
 
 func erase_setting(p_str: String) -> void:
 	editor_settings.erase(p_str)
+
```
TokisanGames, line 480: Add
```


## Undo / Redo Functions


```

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -477,3 +477,25 @@ func has_setting(p_str: String) -> bool:
...
+
+func create_undo_action(action_name: String) -> void:
```
TokisanGames: Match the other parameter style by adding `p_` to all of these.

src/terrain_3d.h
```diff
@@ -62,7 +62,7 @@ class Terrain3D : public Node3D {
 	Terrain3DCollision *_collision = nullptr;
 	Terrain3DMesher *_mesher = nullptr;
 	Terrain3DEditor *_editor = nullptr;
-	EditorPlugin *_plugin = nullptr;
+	Object *_plugin = nullptr;
```
TokisanGames: Let's also rename this _editor_plugin. The function names can stay the same.

src/terrain_3d_editor.cpp
```diff
@@ -629,23 +629,24 @@ void Terrain3DEditor::_store_undo() {
...
 	}
 
-	// Store data in Godot's Undo/Redo Manager
-	EditorUndoRedoManager *undo_redo = _terrain->get_plugin()->get_undo_redo();
-	LOG(INFO, "Storing undo snapshot");
-	String action_name = String("Terrain3D ") + OPNAME[_operation] + String(" ") + TOOLNAME[_tool];
-	LOG(DEBUG, "Creating undo action: '", action_name, "'");
-	undo_redo->create_action(action_name, UndoRedo::MERGE_DISABLE, _terrain);
+	// Request the plugin store the undo/redo data.
+	if (_terrain->get_plugin()->has_method("create_undo_action")) {
```
TokisanGames, line 636: If this is moved to set_plugin, this outer if block can be removed, and these changes will be minimal.

src/terrain_3d_editor.cpp
IndigoBeetle: Interesting, by some definition, it is sorted alphabetically, as "v"ariant comes after "c"lasses. I presume you mean sorted only by the final filename?

src/terrain_3d_editor.cpp
IndigoBeetle: But the plugin is now an Object, and might not implement the update_grid method.

src/terrain_3d.cpp
IndigoBeetle: What if a particular use case doesn't require some of these features? What if I wanted the grid and decal, but don't care about undo/redo?

src/terrain_3d.h
IndigoBeetle: Rename the variable to _editor_plugin, but leave the setter and getter as set/get_plugin?

src/terrain_3d_editor.cpp
IndigoBeetle: See above, about partial implementation.

src/terrain_3d_editor.cpp
TokisanGames: Oh, you're right, my mistake. I assumed both were unsorted in the same way.

src/terrain_3d.h
TokisanGames: Yes

src/terrain_3d_editor.cpp
TokisanGames: Alright, has_method before each call is fine throughout. Please rename this.

TokisanGames:
Thank you.

Issue #835, opened 2025-10-26
Dev mode


TokisanGames:
### Description

Some devs on our team are unable to play or work on our game because of not enough vram. They also don't need to work on the terrain. Often the rest of our developers never need to change the terrain. It would be good to have some options or modes that offer the following features to reduce VRAM, load time. Here are some ideas. Anything else you can think of?

* [x] Option to not create MMIs. Transforms still update w/ sculpting.
* [x] Disable color map #379 
* Locked / read only mode which disables the asset dock and tools.
* Replace all meshes with placeholder cubes (size matches the AABB, and a mesh-type-unique highlight color)
* Asset dock loads meshes to capture the thumbnail, then unloads
* Command line option to enable some of these
* Single option for multiple of these. Maybe 3-4 modes: Dev_mode = { Read Only, Editable, Placeholder, Saver }. The last being no color map, no instances.

aidandavey:
Maybe unload TextureAssets 2-31, fallback to full autoshader in saver/placeholder modes

Issue #379, opened 2024-05-21
add an option to disable the color map, to save ram/vram if desired


Xtarsia:
### Description

an option to disable the color map, to save ram/vram if desired.

TokisanGames:
Also an option to bake the color map to a compressed format. Can only be done in the editor and files must be saved to disk as the export engine doesn't have the compression libraries.

PR #836, opened 2025-10-26
Implement dev mode options

TokisanGames:
Begining of some options in #835 

* Added an option to disable the instancer
* Added rudimentary readonly mode. Currently it only sets a flag, which the editor_plugin.gd uses to disable cursor input
* Added an option to disable color_map generation in VRAM. Still writes and saves data. Partial #379 

Pending:
* If color_map is disabled, then at game run time free the maps from the regions to save RAM.
* In read_only mode:
  * have Terrain3DInstancer and Terrain3DEditor early exit on Operating. We'll leave the API, just no hand editing.
  * disable the tool bar and asset dock.
* Add placeholder mode that replaces all meshes with AABB sized cubes and enable highlight
* Add command line option for dev_mode
* Add dev_mode placeholder (no color map, placeholder instances), saver (no color map, no instances, autoshader?)

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -69,7 +69,7 @@ uniform highp sampler2DArray _height_maps : repeat_disable;
 uniform highp sampler2DArray _control_maps : repeat_disable;
 //INSERT: TEXTURE_SAMPLERS_NEAREST
 //INSERT: TEXTURE_SAMPLERS_LINEAR
-uniform bool _color_maps_enabled = true;
+uniform bool _color_map_enabled = true;
 // Public uniforms
 
 group_uniforms general;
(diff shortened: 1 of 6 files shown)


aidandavey:
Does hiding the MMI parent change the VRAM allocation, or only skip the draw calls? What if we instead destroy the MMIs and skip rebuild to 'hide' them? That should save RAM, VRAM, avoid processing the MMIs and avoid the draw calls.

In the Node to RS version there will be no MMI parent to hide, so would need to take the above approach anyway.

TokisanGames:
> Does hiding the MMI parent change the VRAM allocation, or only skip the draw calls? 

show_instances turns off visibility of the MMI nodes
instancer_mode = disabled destroys and prevents creation of the MMIs. 

The latter saves ram and vram. The former will go away with your RS version. Transforms still update in either case.

PR #840, opened 2025-11-03
Fix #830 - get_intersection() implementation does not align with the docs

### LINKED ISSUE #830: get_intersection() doesn't align with docs
### Discussed in https://github.com/TokisanGames/Terrain3D/discussions/829

<div type='discussions-op-text'>

<sup>Originally posted by **MeteorShower2004** October 23, 2025</sup>
Here is the link to official documents about [ `get_intersection` ](https://terrain3d.readthedocs.io/en/latest/api/class_terrain3d.html#class-terrain3d-method-get-intersection)

> If gpu_mode is enabled, it uses the GPU to detect the mouse. This works wherever the terrain is visible, even outside of regions, but may need to be called twice.

The document says that you can use `gpu_mode=true` to get the height of non-region terrain. (aka. the area outside the terrain)

However in the cpp implementation, `get_intersection`,  actually detect the direction first and when it finds out the `(direction-Vec3(1, -1, 0)).length_squared()<0.00001`(source code [here](https://github.com/TokisanGames/Terrain3D/blob/main/src%2Fterrain_3d.cpp#L712-L712) ), it **falls back to use `get_height`**, which **can't** get the height of non-region actually. 

(After experimenting, I figured out a vector that could bypass that check. It's `Vec3(0, -1, 0.005)`.)

This results in inconsistence between documents and the acutal behavior of the function.
So I suggest that the document should get fixed, or the implementation shoud match with the describtion in document.

Thanks.</div>

aidandavey:
(cherry picked from commit 8d17f6085df1db13a585eace85a3e99da6b5e93b)

Fixes #830 

If the direction is down or up in GPU mode, do not use look_at() just set the rotation on mouse_cam directly.

In non-GPU mode, determine whether we are below and looking up or above and looking down, then return get_height() or NAN accordingly. 

Grouped GPU and non-GPU logic

Handle impossible direction vectors

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -703,32 +703,33 @@ void Terrain3D::set_cull_margin(const real_t p_margin) {
  * Returns Vec3(NAN) on error or vec3(3.402823466e+38F) on no intersection. Test w/ if (var.x < 3.4e38)
  */
 Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_direction, const bool p_gpu_mode) {
-	if (p_direction == V3_ZERO || !p_direction.is_finite()) {
+	if (p_direction.is_zero_approx() || !p_direction.is_finite()) {
 		LOG(ERROR, "Invalid direction vector: ", p_direction);
+		return V3_NAN;
+	}
+	if (!p_src_pos.is_finite()) {
+		LOG(ERROR, "Invalid source vector: ", p_src_pos);
 		return V3_NAN;
 	}
 	Vector3 direction = p_direction.normalized();
 	Vector3 point;
 
+	// Raymarching mode
 	if (!p_gpu_mode) {
-		// Use raymarching mode
-
-		// If looking straight down and the source is above the terrain use get_height, otherwise return NaN
-		// If looking straight up and the source is below the terrain use get_height, otherwise return NaN
-		if (abs(direction.y) > 0.00001f) {
-			float height = _data->get_height(p_src_pos);
-			bool is_down = direction.y < 0.f;
-			if ((is_down && height < p_src_pos.y) || (!is_down && height > p_src_pos.y)) {
-				return Vector3(p_src_pos.x, height, p_src_pos.z);
-			}
-			return V3_NAN;
+		// Must start above terrain if in a region
+		real_t height = _data->get_height(p_src_pos);
+		if (height > p_src_pos.y) { // False if Nan
+			return V3_MAX;
 		}
-
+		// If looking straight down use get_height
+		if (direction.y < -.99999f) {
+			return Vector3(p_src_pos.x, height, p_src_pos.z);
+		}
 		// Raymarch down the ray in small increments until we find the terrain height
 		point = p_src_pos;
 		for (int i = 0; i < 4000; i++) {
-			real_t height = _data->get_height(point);
-			if (point.y - height <= 0) {
+			height = _data->get_height(point);
+			if (point.y - height <= 0.f) { // Nan comparison is false
 				return point;
 			}
 			point += direction;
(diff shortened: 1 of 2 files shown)


src/terrain_3d.cpp
```diff
@@ -698,27 +698,35 @@ void Terrain3D::set_cull_margin(const real_t p_margin) {
...
-		return V3_NAN;
-	}
 	Vector3 direction = p_direction.normalized();
 	Vector3 point;
 
-	// Position mouse cam one unit behind the requested position
-	_mouse_cam->set_global_position(p_src_pos - direction);
+	if (direction == Vector3(0.0f, 0.0f, 0.0f) || !direction.is_finite()) {
+		LOG(ERROR, "Invalid direction vector: ", direction);
+		return Vector3(NAN, NAN, NAN);
```
TokisanGames: This can be V3_NAN. It's not just syntactic sugar. It's a static constant that doesn't have to be recreated everytime this line runs, and uses the same memory as every other V3_NAN. Defined in constants.h

Also use V3_ZERO above.

This block can be moved to the top, before normalizing the direction

src/terrain_3d.cpp
```diff
@@ -732,7 +740,24 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
 	} else {
 		// Else use GPU mode, which requires multiple calls
 		// Get depth from perspective camera snapshot
-		_mouse_cam->look_at(_mouse_cam->get_global_position() + direction, V3_UP);
+
+		if (!_mouse_cam) {
+			LOG(ERROR, "Invalid mouse camera");
+			return Vector3(NAN, NAN, NAN);
```
TokisanGames: V3_NAN

src/terrain_3d.cpp
```diff
@@ -732,7 +740,24 @@ Vector3 Terrain3D::get_intersection(const Vector3 &p_src_pos, const Vector3 &p_d
...
+		_mouse_cam->set_global_position(p_src_pos - direction);
+
+		// If looking straight down or up, look_at() won't work, so set the rotation directly
+		if ((direction - Vector3(0.f, -1.f, 0.f)).length_squared() < 0.00001f) {
+			_mouse_cam->set_rotation_degrees(Vector3(-90.f, 0.f, 0.f));
+		} else if ((direction - Vector3(0.f, 1.f, 0.f)).length_squared() < 0.00001f) {
+			_mouse_cam->set_rotation_degrees(Vector3(90.f, 0.f, 0.f));
+		} else {
+			_mouse_cam->look_at(_mouse_cam->get_global_position() + direction, Vector3(0.f, 1.f, 0.f));
+		}
```
TokisanGames, line 754: Because direction is normalized, we only need to determine if abs(y) is very close to 1.0 So this block could be reduced down to:
```
if (abs(direction.y) > 1.f - 0.00001f) {
    _mouse_cam->set_rotation_degrees(Vector3(90.f * (direction.y > 0.f ? 1.f : -1.f), 0.f, 0.f));
} else {
    _mouse_cam->look_at(_mouse_cam->get_global_position() + direction, V3_UP);
}
```

src/terrain_3d.cpp
```diff
@@ -698,27 +698,35 @@ void Terrain3D::set_cull_margin(const real_t p_margin) {
...
+			}
+			return Vector3(NAN, NAN, NAN);
+		}
+		// If looking straight up and the source is below the terrain use get_height, otherwise return NaN
+		if ((direction - Vector3(0.f, 1.f, 0.f)).length_squared() < 0.00001f) {
+			float height = _data->get_height(p_src_pos);
+			if (height > p_src_pos.y) {
+				return Vector3(p_src_pos.x, height, p_src_pos.z);
+			}
+			return Vector3(NAN, NAN, NAN);
```
TokisanGames: Can be reduced to:
```
if (abs(direction.y) > 0.00001f) {
    float height = _data->get_height(p_src_pos);
    bool is_down = direction.y < 0.f;
    if ((is_down && height < p_src_pos.y) || (!is_down && height > p_src_pos.y)) {
        return Vector3(p_src_pos.x, height, p_src_pos.z);
    }
    return V3_NAN;
}
```

src/terrain_3d.cpp
aidandavey: Nice!

aidandavey:
> Ok, finished with my review. Thanks.

Thanks, updated following your review

TokisanGames:
Thanks for this. Looking at it again, I don't think we need to support looking up from below the terrain. The terrain is back face culled by default, so gpu mode won't even work. Now if below the terrain, it returns V3_MAX in both modes. I simplified some of those conditions.

TokisanGames:
This broke the editor ortho camera because it's >500k high. Fix in #848.

PR #841, opened 2025-11-04
Installing hook did not set execution bit on Linux.

Jackerty:
The precommit install script does not set execution bit. Linux user has to manually flip that. Commit adds this part to the install script.

Files ".git/hooks/pre-commit" and ".git/hooks/pre-commit-clang-format" are set to "-rwxr--r--" a.k.a owning user has read write execute rights; group with owning user and others have only read rights.

I put platform check because I'm not sure is this good in Apple products.

### DIFF
--- a/tools/install-hooks.py
+++ b/tools/install-hooks.py
@@ -1,4 +1,5 @@
 import os
+import stat
 import sys
 import shutil
 
@@ -11,5 +12,11 @@ if not os.path.exists(dist_dir):
 
 for file in os.listdir(src_dir):
     shutil.copy(os.path.join(src_dir, file), dist_dir)
+
+# Unix user needs the execute bit set.
+if os.name == "posix":
+    mode = stat.S_IRWXU | stat.S_IRGRP | stat.S_IROTH;
+    os.chmod(os.path.join(dist_dir, "pre-commit"), mode);
+    os.chmod(os.path.join(dist_dir, "pre-commit-clang-format"), mode);
 
 print("Copied tools/hooks/* to .git/hooks")


tools/install-hooks.py
```diff
@@ -12,4 +13,9 @@
 for file in os.listdir(src_dir):
     shutil.copy(os.path.join(src_dir, file), dist_dir)
 
+# Linux user needs the execute bit set.
+if sys.platform == "linux":
+    os.chmod(os.path.join(dist_dir, "pre-commit"), stat.S_IRWXU | stat.S_IRGRP | stat.S_IROTH);
+    os.chmod(os.path.join(dist_dir, "pre-commit-clang-format"), stat.S_IRWXU | stat.S_IRGRP | stat.S_IROTH);
```
TokisanGames: Maybe this should be something like this, so it only adds the executable and doesn't change the other bits.

```
# POSIX systems need the execute bit set.
if os.name == "posix":
    path = os.path.join(dist_dir, "pre-commit")
    os.chmod(path, os.stat(path).st_mode | stat.S_IXUSR)
    path = os.path.join(dist_dir, "pre-commit-clang-format")
    os.chmod(path, os.stat(path).st_mode | stat.S_IXUSR)
```

Jackerty: hmm... There is a security argument here that being precise is better:

- makes it faster to check what rights are given. This makes "the mandatory check" for no viruses in the scripts faster.
- Makes sure only what is actually needed is enabled.

TokisanGames: My argument would hold if we were changing their files. But then I realized, these are files we are placing, so setting the full permissions are fine.

But I still think you should change to posix so it works on macos.

Jackerty: I have never used Mac so I'm a bit uncomfortable with that. Kinda was hoping some mac user would confirm.

I guess if want it then why not remove whole check?

What I read from[ python manual](https://docs.python.org/3/library/os.html#os.chmod) only user read write bits are implement in windows and everything else is ignored.

TokisanGames: I did some reading on it. If you use chmod on windows, the execute bit is ignored, but it might set the read only flag on the files, which could cause other issues. I think changing the conditional to posix is best.

```
if os.name == "posix":
    mode = stat.S_IRWXU | stat.S_IRGRP | stat.S_IROTH
    os.chmod(os.path.join(dist_dir, "pre-commit"), mode)
    os.chmod(os.path.join(dist_dir, "pre-commit-clang-format"), mode)
```

Jackerty: done and rebased

TokisanGames:
Thank you.

PR #842, opened 2025-11-05
Option to use compressed color map

stan4dbunny:
Admin edit

Partially addresses https://github.com/TokisanGames/Terrain3D/issues/379

This PR compresses the color map by 75% (16mb to 4mb per 1k), and if you choose ETC even on desktop it goes to 87.5% (2mb).
Enable it in Terrain3D/Regions/Advanced/Color Compress Mode

How it works:
* The compressed map is freed in editor, and written if modified on save (or changing the compress mode).
* At runtime, the uncompressed map is freed. We cannot edit compressed maps at runtime, so color/wetness are readonly. There are no errors, it should silently fail.
* Both maps are stored in the region file - even on export. We haven't found a good way to strip it out yet. We can look at that in the future.

If you double-click a region file it will usually show you what's inside. However we free one of the color maps, depending on if we're in editor or in game, so viewing Godot's in-memory version of the file will show one map missing. You can view what is actually in the file by:
* Open Godot without opening a scene with a Terrain3D that loads those specific region files. Then you can open them in the inspector and see what they contain.
* You can run Terrain3DRegion.dump() to print out pointers and formats of the maps. A null map has a 0x0 pointer. Do this before the region goes through data.add_region() or it will get stripped.

------------

I have implemented an option to use compressed color maps when the game is running (i.e when not editing). When the compress option is used and the scene is saved, the regions will all get a compressed color map. When the game is run (either launched from editor or when exported), the compressed color map will be used and the uncompressed color map will be unloaded.

I have tested that the compressed texture is used instead of the normal one when the game is running with the option enabled by analyzing the frame in Renderdoc. The compressed color map is also updated after the terrain has been sculpted/painted in the editor.

With the compressed color map, the color map size is 4x smaller.

image

image

I have also added some documentation in the existing style.

One thing that I noticed is that when there is a large terrain with a lot of regions, the saving can take a long time (several minutes) when the option has been clicked, but this should only happen once. Maybe this time could be reduced if the compression happend on the entire color map at once instead of for the tile in each region.


src/terrain_3d_region.h
```diff
@@ -48,6 +50,11 @@ class Terrain3DRegion : public Resource {
 	Ref<Image> _height_map;
 	Ref<Image> _control_map;
 	Ref<Image> _color_map;
+
+	Ref<Image> _compressed_height_map;
+	Ref<Image> _compressed_control_map;
```
TokisanGames: The control map cannot be compressed. I don't know about the height, it's worth exploring, but since you didn't implement it, please remove these.

src/terrain_3d_region.cpp
```diff
@@ -60,6 +60,10 @@ Ref<Image> Terrain3DRegion::get_map(const MapType p_map_type) const {
 		case TYPE_CONTROL:
 			return get_control_map();
 		case TYPE_COLOR:
+			if (_compressed_color_map.is_valid() && !IS_EDITOR) {
+				const_cast<Ref<Image> &>(_color_map).unref();
```
TokisanGames: This breaks the promise of const on the function. Just return the compressed map and don't touch the original here.

Uncompressed maps could be freed immediately after being loaded in Terrain3DData:load_region and load_directory

src/terrain_3d.h
```diff
@@ -70,6 +72,7 @@ class Terrain3D : public Node3D {
 	// Regions
 	RegionSize _region_size = SIZE_256;
 	bool _save_16_bit = false;
+	bool _use_compressed_color_map = false;
```
TokisanGames: Let's reduce this to `_compress_color_map` everywhere.
`compressed_color_map` if referencing the Image resource.
`compress_color_map` if referencing a flag or function.

src/terrain_3d_data.cpp
```diff
@@ -48,7 +48,7 @@ void Terrain3DData::_copy_paste_dfr(const Terrain3DRegion *p_src_region, const R
 ///////////////////////////
 
 void Terrain3DData::initialize(Terrain3D *p_terrain) {
-	if (!p_terrain) {
+ 	if (!p_terrain) {
```
TokisanGames: We don't need to add a space here.

src/terrain_3d_region.cpp
```diff
@@ -281,9 +303,19 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
...
 		_height_map = original_map;
+	} else if (p_compressed_color_map) {
+		_compressed_color_map.unref();
+		_compressed_color_map = Image::create_empty(_color_map->get_width(), _color_map->get_height(), _color_map->has_mipmaps(), _color_map->get_format());
+		_compressed_color_map->copy_from(_color_map);
+		_compressed_color_map->compress(Image::COMPRESS_BPTC, Image::COMPRESS_SOURCE_SRGB);
+		err = ResourceSaver::get_singleton()->save(this, get_path(), ResourceSaver::FLAG_COMPRESS);
+	} else if (!p_compressed_color_map) {
+		_compressed_color_map.unref();	
+		err = ResourceSaver::get_singleton()->save(this, get_path(), ResourceSaver::FLAG_COMPRESS);
```
TokisanGames: This block is unnecessary. Move `_compressed_color_map.unref();` after set_version(), before the whole if statement, then drop this block and let else below take care of it.

src/terrain_3d.cpp
```diff
@@ -901,7 +920,7 @@ void Terrain3D::_notification(const int p_what) {
 			_setup_mouse_picking();
 			if (_free_editor_textures && !IS_EDITOR && _assets.is_valid() && !_assets->get_path().contains("Terrain3DAssets")) {
 				LOG(INFO, "free_editor_textures enabled, reloading Assets path: ", _assets->get_path());
-				_assets = ResourceLoader::get_singleton()->load(_assets->get_path(), "", ResourceLoader::CACHE_MODE_IGNORE);
+				_assets = ResourceLoader::get_singleton()->load(_assets->get_path(), "", ResourceLoader::CACHE_MODE_REPLACE);
```
TokisanGames: What does Assets have to do with the colormap? I'm pretty sure this will break free_editor_textures.

src/terrain_3d_region.h
```diff
@@ -48,6 +50,9 @@ class Terrain3DRegion : public Resource {
 	Ref<Image> _height_map;
 	Ref<Image> _control_map;
 	Ref<Image> _color_map;
+
```
TokisanGames: no blank line

src/terrain_3d_region.cpp
```diff
@@ -141,6 +145,24 @@ void Terrain3DRegion::set_color_map(const Ref<Image> &p_map) {
...
 
+void Terrain3DRegion::set_compressed_color_map(const Ref<Image> &p_map) {
+	LOG(INFO, "Setting compressed color map for region: ", (_location.x != INT32_MAX) ? String(_location) : "(new)");
+	if (_region_size == 0 && p_map.is_valid()) {
+		set_region_size(p_map->get_width());
+	}
+	Ref<Image> map = p_map;
+
+	if (map.is_valid() && !map->has_mipmaps()) {
+		map->generate_mipmaps();
```
TokisanGames: You have an issue here:
`This method returns an error if the image is compressed`
https://docs.godotengine.org/en/stable/classes/class_image.html#class-image-method-generate-mipmaps

src/terrain_3d_region.cpp
```diff
@@ -141,6 +145,24 @@ void Terrain3DRegion::set_color_map(const Ref<Image> &p_map) {
...
+	}
+
+	// If already initialized and receiving a new map, or the map was sanitized
+	if (_compressed_color_map.is_valid() && _compressed_color_map != p_map || map != p_map) {
```
TokisanGames: You're not sanitizing here. `map` may not be a different ref if mipmaps have been generated, ignoring the problem above. `map != map` can be removed, and instead mark _modified if generating mipmaps.

src/terrain_3d_region.cpp
```diff
@@ -274,6 +296,16 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
 	LOG(MESG, "Writing", (p_16_bit) ? " 16-bit" : "", " region ", _location, " to ", get_path());
 	set_version(Terrain3DData::CURRENT_VERSION);
 	Error err = OK;
+
+	if (p_compressed_color_map) {
+		_compressed_color_map.unref();
+		_compressed_color_map = Image::create_empty(_color_map->get_width(), _color_map->get_height(), _color_map->has_mipmaps(), _color_map->get_format());
+		_compressed_color_map->copy_from(_color_map);
+		_compressed_color_map->compress(Image::COMPRESS_BPTC, Image::COMPRESS_SOURCE_SRGB);
```
TokisanGames: `compress()` isn't great. It autoselects the format and may drop channels if there isn't enough data. We use `compress_from_channels()` in `Terrain3DUtil::get_filled_image()`

Are you sure our color map is in SRGB? I'm not sure. We should document it in Region.xml:color_map.

src/terrain_3d_region.cpp
```diff
@@ -444,7 +480,7 @@ void Terrain3DRegion::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_instances", "instances"), &Terrain3DRegion::set_instances);
 	ClassDB::bind_method(D_METHOD("get_instances"), &Terrain3DRegion::get_instances);
 
-	ClassDB::bind_method(D_METHOD("save", "path", "save_16_bit"), &Terrain3DRegion::save, DEFVAL(""), DEFVAL(false));
+	ClassDB::bind_method(D_METHOD("save", "path", "save_16_bit", "use_compressed_color_map"), &Terrain3DRegion::save, DEFVAL(""), DEFVAL(false));
```
TokisanGames: You need to add another DEFVAL

src/terrain_3d_data.cpp
```diff
@@ -1195,7 +1195,7 @@ void Terrain3DData::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("remove_region", "region", "update"), &Terrain3DData::remove_region, DEFVAL(true));
 
 	ClassDB::bind_method(D_METHOD("save_directory", "directory"), &Terrain3DData::save_directory);
-	ClassDB::bind_method(D_METHOD("save_region", "region_location", "directory", "save_16_bit"), &Terrain3DData::save_region, DEFVAL(false));
+	ClassDB::bind_method(D_METHOD("save_region", "region_location", "directory", "save_16_bit", "use_compressed_color_map"), &Terrain3DData::save_region, DEFVAL(false));
```
TokisanGames: You need to add a DEFVAL

src/terrain_3d_data.cpp
```diff
@@ -372,7 +372,7 @@ void Terrain3DData::load_directory(const String &p_dir) {
 			LOG(ERROR, "Cannot get region location from file name: ", fname);
 			continue;
 		}
-		Ref<Terrain3DRegion> region = ResourceLoader::get_singleton()->load(path, "Terrain3DRegion", ResourceLoader::CACHE_MODE_IGNORE);
+		Ref<Terrain3DRegion> region = ResourceLoader::get_singleton()->load(path, "Terrain3DRegion", ResourceLoader::CACHE_MODE_REPLACE);
```
TokisanGames, line 436: What is the intention behind changing this here, and the default to Util::load_image? 
This change requires careful testing with git, git-lfs, and other external activities. The previous mode was chosen for a reason.

https://github.com/TokisanGames/Terrain3D/commit/4c2cec78a9af78985b2fef2d81d8a82fe1162b4a

TokisanGames (changes requested):
Thanks for the PR. Here are some notes:

Your commits are coming on another account. You probably want to fix that before I merge. You can also squash your commits at the same time.

Image

src/terrain_3d_region.cpp
TokisanGames: Also, neither BPTC nor DXT are widely supported on mobile, which would prefer ETC2 or ASTC. We should have the option somewhere of specifying that. Probably changing p_compressed_color_map to p_compression_mode of type Image::CompressMode with the default being COMPRESS_MAX for none. And changing the checkbox in the Inspector to _color_map_compress_mode, exposing the enum.

src/terrain_3d_data.cpp
stan4dbunny: I had a problem with the thumbnail of the compressed color map not updating in the inspector when toggling compress_color_maps and saving. The thumbnail changed when the editor was restarted, but sometimes it wouldn't update for all regions, so sometimes one region had its compressed color map when it shouldn't have, while all the others didn't. Changing to CHANGE_MODE_REPLACE made the thumbnail update instantly and I didn't have the problem with not all regions changing correctly.

stan4dbunny: I don't know if changing both to CHANGE_MODE_REPLACE is needed though, so I can look into it.

TokisanGames: Viewing the region in the inspector by double-clicking on the file shows what is on disk, not what's in RAM. It's fine for casual inspection or occasional debugging but not recommended for viewing the current data accurately. Godot loading a resource into the inspector is really a Godot issue that has little to do with Terrain3D anyway. Inspector reading is not something we should change how we handle caching throughout the entire system for.

stan4dbunny: Can I ask why you have chosen CACHE_MODE_IGNORE?

From my understanding, what's seen in the inspector is what's in the cache, so why wouldn't we want to update it when the scene is saved? 

I've figured out that it's the CACHE_MODE_REPLACE in Terrain3DData::load_directory on line 375 that seems to make the difference. I tried changing it back to CACHE_MODE_IGNORE , which has following effects: 
-The editor has to be restarted for the changes in the region inspector to appear (added/removed compressed color map)
-When using CACHE_MODE_IGNORE there is a bug where when switching between compressed textures and normal textures one region resource retains its old value for the compressed texture which causes errors due to later code mixing compressed and uncompressed textures in the texture array. This bug is fixed by using CACHE_MODE_REPLACE.

I understand that the problem might be with the caching in Godot and not Terrain3D, but using CACHE_MODE_REPLACE fixes this problem, so what would be the downside of using it?

stan4dbunny: This is where I read about the fix https://github.com/godotengine/godot/issues/104966#issuecomment-2865270420

src/terrain_3d_region.cpp
stan4dbunny: I've started working on this in the latest commit. Currently, I'm not sure how to handle certain formats not being available on the used hardware. For example, ETC on desktop. 

At the moment, you can freely select ETC. When you try to view the compressed color map in the region inspector, this error will show up:

image

image

And when I run the application using RenderDoc, the normal color map is used instead. 

What sort of error message/constraint do you think would be appropriate here?

I also haven't fixed using RGB/RGBA channels correctly yet for each format, but I'll do that as well!

This is what the option looks like now

image

src/terrain_3d_data.cpp
TokisanGames: Showing the right thumbnails when viewing a res file is a nice to have feature, but not essential. Being able to load regions and textures properly is essential. Changing cache modes could break the terrain under certain circumstances. It requires care and testing and cannot be done casually.

You changed cache handling in 3 places:
* assets under free_editor_textures - no reason to change it in this PR, and probably has broken a feature
* Util::load_image - no reason to change it in this PR
* region loading - up for debate

When loading regions in the editor, we want a non-cached version. If swapping scenes, the regions should stay in memory. When loading in game when we don't intend to edit, we could load cached.

We've had problems with cached region files in the past. Replace mode might fix the thumbnails, but I don't care about that if it breaks the terrain saving or loading. Changing to replace mode is an option, but it needs to be thoroughly tested to ensure it is reliable in all circumstances. e.g. changing data directories, replacing the res files externally by hand and with git, and under git-lfs, changing between editor scenes with terrains, saving compressed color maps, deleting and saving many res files such as after changing region sizes. The loading system needs to be abused to be tested.

src/terrain_3d_region.cpp
```diff
@@ -273,7 +284,15 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
...
 	set_version(Terrain3DData::CURRENT_VERSION);
+	_compressed_color_map.unref();
 	Error err = OK;
+
+	if (p_compression_mode != Image::COMPRESS_MAX) {
+		_compressed_color_map = Image::create_empty(_color_map->get_width(), _color_map->get_height(), _color_map->has_mipmaps(), _color_map->get_format());
+		_compressed_color_map->copy_from(_color_map);
+		//_compressed_color_map->compress(Image::COMPRESS_BPTC, Image::COMPRESS_SOURCE_SRGB);
+		_compressed_color_map->compress_from_channels(p_compression_mode, Image::USED_CHANNELS_RGBA);
+	}
```
stan4dbunny: Here I should correctly handle the channels used, a well as add ASTC format. 

Not sure how to handle not allowing users to set compression formats that are incompatible with the used hardware. Currently, the normal color map will be used instead when the application is run, but an error pops up in the editor.

510864682-e0d13e2d-078c-4791-b263-db160cedb626

src/terrain_3d_data.cpp
stan4dbunny: I've removed the other instances, so let's only talk about the cache mode in region loading.

The thumbnail updating in the inspector wasn't my main issue with CACHE_MODE_IGNORE. The main problem was that with CACHE_MODE_IGNORE there seemed to be some weird caching behaviour going on with the regions, when switching between compressed textures and normal textures. When restarting the editor the changes would be reflected, but sometimes one region would still have its compressed color map in the inspector even though it wasn't supposed to. When pressing play in the editor, the terrain would be black and the error "using initialized RID" pops up. I'm pretty sure that this is because the texture array will have the normal color maps as well as one compressed color map. 

When changing the region loading to CACHE_MODE_REPLACE, yes, the thumbnail updates correctly, but I also never had that issue again with one region not correctly caching+saving/uncaching+unsaving its compressed color map.

> When loading regions in the editor, we want a non-cached version. If swapping scenes, the regions should stay in memory. When loading in game when we don't intend to edit, we could load cached.

Why wouldn't we want the cached version when loading the regions in the editor, or when swapping scenes with the same terrain?

> We've had problems with cached region files in the past.

What type of problems did you have then?

> Changing to replace mode is an option, but it needs to be thoroughly tested to ensure it is reliable in all circumstances. e.g. changing data directories, replacing the res files externally by hand and with git, and under git-lfs, changing between editor scenes with terrains, saving compressed color maps, deleting and saving many res files such as after changing region sizes. The loading system needs to be abused to be tested.

I haven't tested extensively, but from what I've seen the loading and saving seems to work as it should. I've tested doing sculpting/painting and creating new regions, and this seems to work and the compressed color map is updated when the scene is saved. But of course this needs to be tested a lot like you said.

TokisanGames: > there seemed to be some weird caching behaviour going on with the regions, when switching between compressed textures and normal textures. When restarting the editor the changes would be reflected, but sometimes one region would still have its compressed color map in the inspector even though it wasn't supposed to. When pressing play in the editor, the terrain would be black and the error "using initialized RID" pops up. I'm pretty sure that this is because the texture array will have the normal color maps as well as one compressed color map.

This doesn't seem like something changing the cache mode would cause or fix. It seems like the logic in generating the texture array, and which map to choose is or was incorrect. It bears looking closely again at that code flow before we merge.


> Why wouldn't we want the cached version when loading the regions in the editor, 

If the region has been modified on disk say via git, we want that version, not whatever is in memory. Godot doesn't detect that all files have been modified externally. It may for certain types like textures, but not every file. It certainly has no facility for it to reload our regions, as that is done only by calling our load functions.

> or when swapping scenes with the same terrain?

If the terrain is loaded in two scenes, and edited in one, it shouldn't be updated in the other until it is saved. It can get dicey. They shouldn't be editing the same data in two scenes, but might load it for reference. It's debatable if it should use the version in memory, but that's out of scope for this PR.

We could discuss and look at loading in a cached mode when running in game, which might improve scene changing in game. But some people edit in game. Again this is out of scope for this PR.

> What type of problems did you have then?
It's been years of problems we have been working on. I don't recall specifically, but generally it would have been displaying unintended data, maybe old versions, or worse, corruption.

However, as I understand it REPLACE mode is nearly the same as IGNORE mode in that it does not read from the cache, but it does add the data to the cache. So that is why I'm open to it in this PR if thoroughly tested. It must be tested with external changes; that's the primary source of problems with caching.

src/terrain_3d_region.cpp
```diff
@@ -141,6 +144,14 @@ void Terrain3DRegion::set_color_map(const Ref<Image> &p_map) {
...
 }
 
+void Terrain3DRegion::set_compressed_color_map(const Ref<Image> &p_map) {
+	LOG(INFO, "Setting compressed color map for region: ", (_location.x != INT32_MAX) ? String(_location) : "(new)");
+	if (_region_size == 0 && p_map.is_valid()) {
+		set_region_size(p_map->get_width());
+	}
+	_compressed_color_map = p_map;
+}
+
```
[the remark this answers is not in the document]
TokisanGames: If you want the compressed color map saved into the region and written to disk (you do), then the region needs to be marked modified just as the other setter functions do. You had it before, but you removed it. My direction wasn't to remove the whole block, it was to remove the map condition, since you're not santizing it. Look again at set_color_map()

src/terrain_3d_region.cpp
TokisanGames: > Currently, I'm not sure how to handle certain formats not being available on the used hardware. For example, ETC on desktop.

People develop on desktop and export to mobile. The idea of the compressed color map is that it is usable on export (and maybe in game mode, but that is less important than export).

> When you try to view the compressed color map in the region inspector, this error will show up:

That demonstrates that perhaps it is properly converting the map, so it can be used on export. That's what we want.

We don't really care if the mobile formats work on desktop, as long as they work on mobile.

Do the errors occur only on ETC? Not on ETC2/ASTC?
Do they occur any other time except when opening in the inspector?

Also I just realized ETC 1 doesn't support an alpha channel, so no wetness in that format. Maybe we just make our own enum and cut out ETC. 

Looking at your screenshot, `compression_mode` makes it look like all maps are compressed with that. Maybe change it to `color_compress_mode`.

When retrieving the color map for texture array generation, you're already detecting if we use the compressed or regular. We may need to extend that logic to detect if the map is in a mobile format and we're running in game mode on desktop, we use the regular color map.

So if compressed map is enabled and valid:
* Editor on Desktop - uncompressed
* Editor on mobile - uncompressed
* Export/Game on desktop - compressed if desktop compatible format, else uncompressed
* Export/Game on mobile - compressed if mobile compatible format, else uncompressed

https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-get-name

And on startup in export/game mode, an option to free the uncompressed map. Adjacent to free_editor_textures.

stan4dbunny: > Do the errors occur only on ETC? Not on ETC2/ASTC?
Do they occur any other time except when opening in the inspector?

The error happens for ETC, ETC2 and ASTC. I have only seen it appear when opening the region in the inspector. You're probably right that it will work on export though.

> Also I just realized ETC 1 doesn't support an alpha channel, so no wetness in that format. Maybe we just make our own enum and cut out ETC.

I was thinking, some users may not want to use wetness, and in that case, using an RGB texture could reduce memory consumption notably in some compression formats, compared to RGBA. I don't know exactly how to structure this UX-wise, though. 

Maybe we could have an option to use RGB when compressing, letting the user know that wetness won't work? And, we could have ETC1 as an option but let the user know that wetness will be ignored. What do you think?

src/terrain_3d_data.cpp
stan4dbunny: How can I help with the testing? What scenarios should I test?:)

src/terrain_3d_region.cpp
stan4dbunny: In order to not export the uncompressed color map upon export, I would have to do something with EditorExportPlugin, right?

src/terrain_3d_data.cpp
TokisanGames: Test replacing the res files with git. Open a scene. Make identifiable changes in a scene. Commit them. Close the scene but keep Godot open. Change HEAD to the previous commit. Open the scene.

Do something similar but moving the files manually on your harddrive. Keep Godot open so it has files in the cache.

Add and remove multiple regions and save. Ensure the changes are accurate.

Make sure changing the region size handles everything properly. Easiest is probably to just erase the compressed colormap.

Open two scenes w/ compressed color maps and swap between them. Ensure data stays consistent.

Draw on the color map, undo, save and restart to check. Do several levels of undo and redo, and ensure the right one gets saved. 

Do several rounds of check compressed, save, uncompress, save, while Godot is open and check the files. Do it restarting Godot in between.

src/terrain_3d_region.cpp
TokisanGames: > Maybe we could have an option to use RGB when compressing, letting the user know that wetness won't work? And, we could have ETC1 as an option but let the user know that wetness will be ignored. What do you think?

I think it's too many options right now. Lets just keep the enum as is, and perhaps drop some or adjust later.

> In order to not export the uncompressed color map upon export, I would have to do something with EditorExportPlugin, right?

I don't know, I've never used it. I see you wrote a class.

src/terrain_3d_data.cpp
```diff
@@ -372,11 +372,16 @@ void Terrain3DData::load_directory(const String &p_dir) {
...
 		if (region.is_null()) {
 			LOG(ERROR, "Cannot load region at ", path);
 			continue;
 		}
+		if (region->get_compressed_color_map().is_valid() && !IS_EDITOR) {
+			Ref<Image> color_map = region->get_color_map();
+			RS->free_rid(color_map->get_rid());
```
TokisanGames: `free_rid()` is only to be used when you have created the RID from one of the servers. It's dangerous to use in any other case.

src/terrain_3d.cpp
```diff
@@ -1190,6 +1202,7 @@ void Terrain3D::_bind_methods() {
 	ADD_GROUP("Regions", "");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "region_size", PROPERTY_HINT_ENUM, "64:64,128:128,256:256,512:512,1024:1024,2048:2048", PROPERTY_USAGE_EDITOR), "change_region_size", "get_region_size");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "save_16_bit"), "set_save_16_bit", "get_save_16_bit");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "color_compression_mode", PROPERTY_HINT_ENUM, "COMPRESS_S3TC,COMPRESS_ETC,COMPRESS_ETC2,COMPRESS_BPTC,COMPRESS_ASTC,COMPRESS_MAX"), "set_color_compression_mode", "get_color_compression_mode");
```
TokisanGames: These text values are for us.
Change COMPRESS_MAX to None. Remove COMPRESS_ from the others.

src/terrain_3d.h
```diff
@@ -161,6 +164,9 @@ class Terrain3D : public Node3D {
 	void change_region_size(const RegionSize p_size) { _data ? _data->change_region_size(p_size) : void(); }
 	void set_save_16_bit(const bool p_enabled);
 	bool get_save_16_bit() const { return _save_16_bit; }
+
```
TokisanGames: No blank line

src/terrain_3d_region.cpp
```diff
@@ -273,7 +284,14 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
 	}
 	LOG(MESG, "Writing", (p_16_bit) ? " 16-bit" : "", " region ", _location, " to ", get_path());
 	set_version(Terrain3DData::CURRENT_VERSION);
+	_compressed_color_map.unref();
 	Error err = OK;
+
+	if (p_color_compression_mode != Image::COMPRESS_MAX) {
```
TokisanGames: The original function should work fine in game mode, but this new one will not because of compress. add `&& IS_EDITOR` to prevent attempting to compress in game.

src/terrain_3d_region.cpp
```diff
@@ -273,7 +284,14 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
...
+		_compressed_color_map = Image::create_empty(_color_map->get_width(), _color_map->get_height(), _color_map->has_mipmaps(), _color_map->get_format());
+		_compressed_color_map->copy_from(_color_map);
```
TokisanGames: create_empty() probably allocates the memory and writes 0s to every pixel. We use it when generating pixels. But in this case, it's probably faster to use `create_from_data()`
`		_compressed_color_map = Image::create_from_data(_color_map->get_width(), _color_map->get_height(), _color_map->has_mipmaps(), _color_map->get_format(), _color_map->get_data());`

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -42,7 +44,6 @@ func _init() -> void:
 	# Get the Godot Editor window. Structure is root:Window/EditorNode/Base Control
 	godot_editor_window = EditorInterface.get_base_control().get_parent().get_parent()
 	godot_editor_window.focus_entered.connect(_on_godot_focus_entered)
-
```
TokisanGames, line 50: Add this space. [GDScript code style](https://terrain3d.readthedocs.io/en/latest/docs/contributing.html#gdscript)

project/addons/terrain_3d/src/export_plugin.gd
```diff
@@ -0,0 +1,27 @@
+@tool
+extends EditorExportPlugin
+var terrain: Terrain3D
+var terrain_data: Terrain3DData
+var region: Terrain3DRegion
+var hashing: String
```
TokisanGames: Remove the unused variables.
The remaining ones are private, preface them with an underscore.
Other than hashing, I don't see any need for the others to be member variables. Can't region be local?
Blank line above and below this block.

project/addons/terrain_3d/src/export_plugin.gd
```diff
@@ -0,0 +1,27 @@
...
+func _get_name() -> String:
+	return "ExportPlugin"
```
TokisanGames: Terrain3DExportPlugin

src/terrain_3d_data.cpp
TokisanGames: I don't really like this here. It is us automatically deciding for the user to remove it without a choice. And you didn't add it to load_region, so it's inconsistent. I had written this:

> And on startup in export/game mode, an option to free the uncompressed map. Adjacent to free_editor_textures.

I still think that's what we should do. A separate option just like free_editor_textures called `free_uncompressed_color_map` maybe. And the code runs right next to the former.

TokisanGames:

src/terrain_3d.cpp
```diff
@@ -547,6 +547,16 @@ void Terrain3D::set_save_16_bit(const bool p_enabled) {
...
 }
 
+void Terrain3D::set_color_compression_mode(const Image::CompressMode p_color_compression_mode) {
+	LOG(INFO, p_color_compression_mode);
+	_color_compression_mode = p_color_compression_mode;
+	TypedArray<Terrain3DRegion> regions = _data->get_regions_active();
+	for (int i = 0; i < regions.size(); i++) {
+		Ref<Terrain3DRegion> region = regions[i];
+		region->set_modified(true);
+	}
```
TokisanGames: After rebase, this can be:
```
void Terrain3D::set_color_compression_mode(const Image::CompressMode p_color_compression_mode) {
	SET_IF_DIFF(_color_compression_mode, p_color_compression_mode);
	LOG(INFO, p_color_compression_mode);
	TypedArray<Terrain3DRegion> regions = _data->get_regions_active();
	for (int i = 0; i < regions.size(); i++) {
		Ref<Terrain3DRegion> region = regions[i];
		region->set_modified(true);
	}
}
```

src/terrain_3d_region.cpp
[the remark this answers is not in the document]
TokisanGames: After rebase:
```
void Terrain3DRegion::set_compressed_color_map(const Ref<Image> &p_map) {
	SET_IF_DIFF(_compressed_color_map, p_map);
	LOG(INFO, "Setting compressed color map for region: ", (_location.x != INT32_MAX) ? String(_location) : "(new)");
	if (_region_size == 0 && p_map.is_valid()) {
		set_region_size(p_map->get_width());
	}
	// If already initialized and receiving a new map
	if (_compressed_color_map.is_valid() && _compressed_color_map != p_map) {
		_modified = true;
	}
}
```

src/terrain_3d_region.cpp
```diff
@@ -164,6 +186,10 @@ void Terrain3DRegion::sanitize_maps() {
 		_modified = true;
 	}
 	_control_map = map;
+	if (p_free_uncompressed_color_maps) {
```
stan4dbunny: I have now implemented that the uncompressed color maps are freed upon request:

1. When not running the game in the editor to save RAM

2. AND when exporting the game, to also save disk space

For 2. I added `export_plugin.gd`, which will free the uncompressed color map when the game is being exported. When I was testing this by debugging in Visual Studio, I noticed that `set_color_map` wouldn't happen, but then `sanitize_maps` would create an empty image if `color_map` is null. So I had to make a change here to return before sanitizing the color_map if `free_uncompressed_color_maps` is set. 

To make 1. work as well, I simply had to do `color_map.unref()`  before returning.

For 1. would it be possible to free the color map already in `set_color_map()`?

src/terrain_3d_data.cpp
stan4dbunny: > Test replacing the res files with git. Open a scene. Make identifiable changes in a scene. Commit them. Close the scene but keep Godot open. Change HEAD to the previous commit. Open the scene.

- I tried sculpting + color a region and commiting without closing Godot, then using `git reset --hard` to go back to the commit before.
- The changes were reverted in the scene.
- Even if the scene wasn't closed upon using `git reset --hard`, Godot prompts the user to reload from disk.

> Do something similar but moving the files manually on your harddrive. Keep Godot open so it has files in the cache.

- I created a copy of the Terrain3D demo project with the demo scene, let's call this `Copy_Scene`.
- I sculped the terrain in `Copy_Scene` and copied the `data` folder + `Demo.tscn`.
- I closed my original scene (but not Godot) and replaced `data`+ `Demo.tscn`.
- I reoponed the scene, and the regions looked like the copied ones, and the settings were updated correctly.
- The scene file `Data.tscn` is required in the copy, because otherwise the regions will update but the settings will not update accordingly after the regions. For example, if `save_16_bit` is enabled or not. I'm guessing this is expected behaviour?
- Separately, I also tested changing `compression_mode` in `Copy_Scene` and pasting `data`+ `Demo.tcsn` into the original scene, with the same results as above.

> Add and remove multiple regions and save. Ensure the changes are accurate.

- I tried adding and removing multiple regions, sculpting + painting wetness.
- The color map and compressed color map update correctly when saving.

> Make sure changing the region size handles everything properly. Easiest is probably to just erase the compressed colormap.

- I changed the region size from 1024 to 512, dividing my terrain into smaller regions.
When saving, these had correctly updated color maps + compressed color maps.
- I changed back and saved to 1024 and everything updated correctly.
- Sometimes, more regions were created than before. For example, if I had the default 3 regions with 1024 size in the demo project and switched to 512 and then back to 1024, I noticed that a seemingly empty region was added, but I assume this is expected behaviour.

> Open two scenes w/ compressed color maps and swap between them. Ensure data stays consistent.

- I created two scenes with different Terrain3D nodes and different `data` folders.
- Changing the compression settings in one scene doesn't affect the settings or regions in the other scene if different `data` folders are used

> Draw on the color map, undo, save and restart to check. Do several levels of undo and redo, and ensure the right one gets saved.

- I tried painting wetness and undoing/redoing.
- Upon save, the color maps and uncompressed color maps are updated correctly after what can be seen on the region.

> Do several rounds of check compressed, save, uncompress, save, while Godot is open and check the files. Do it restarting Godot in between.

- Upon saving the scene, I can see in the inspector thumbail that the compressed color map is added/updated, and in the file explorer I can see that the res files change in size, for example become bigger when compressed color maps are enabled vs not.
- Settings and regions will look the same before and after closing the scene/project.

project/addons/terrain_3d/src/export_plugin.gd
```diff
@@ -0,0 +1,49 @@
...
+		_hash += feat
+	_hash += platform.to_string()
+	return true
+	
+	
+func _customize_resource(resource: Resource, path: String) -> Resource:
+	var _region: Terrain3DRegion
+	if resource is Terrain3DRegion:
+		_region = resource
+		if _color_compression_mode != Image.COMPRESS_MAX and _free_uncompressed_color_maps and _region.compressed_color_map != null:
```
stan4dbunny: I have tested exporting the project + testing the exported project (stepping through in VS + capturing frame in renderdoc), and everything seems to be working as it should.

src/terrain_3d.h
```diff
@@ -70,6 +70,7 @@ class Terrain3D : public Node3D {
 	// Regions
 	RegionSize _region_size = SIZE_256;
 	bool _save_16_bit = false;
+	Image::CompressMode _color_compression_mode = Image::COMPRESS_MAX;
```
TokisanGames: I don't like using the Godot enum. They have a weird order. Let's make our own.

In the exposed property change it like this:
`ADD_PROPERTY(
        PropertyInfo(Variant::INT, "color_compression_mode",
                     PROPERTY_HINT_ENUM, "None,DXT5 (LQ Desktop),BPTC (HQ Desktop),ETC1 (Mobile),ETC2 (Mobile),ASTC (Mobile)"),
        "set_color_compression_mode", "get_color_compression_mode"
    );`

Then make our own enum. Make sure to expose them in _bind_methods with `BIND_ENUM_CONSTANT`
```
enum ColorCompressMode {
 COMPRESS_NONE,
 COMPRESS_DXT5,
 COMPRESS_BPTC,
 COMPRESS_ETC,
 COMPRESS_ETC2,
 COMPRESS_ASTC,
}
```

Then make a translation function, which could just be part of set_color_compression_mode():
```
void Terrain3D::get_image_compression_mode(int index) {
    switch (index) {
        case COMPRESS_DXT5:
            return Image::COMPRESS_S3TC;
        case COMPRESS_BPTC:
            return Image::COMPRESS_BPTC;
        case COMPRESS_ETC:
            return Image::COMPRESS_ETC;
        case COMPRESS_ETC2:
            return Image::COMPRESS_ETC2;
        case COMPRESS_ASTC:
            return Image::COMPRESS_ASTC;
        default: 
            return Image::COMPRESS_MAX;
    }
}
```

src/terrain_3d_region.h
```diff
@@ -95,6 +100,9 @@ class Terrain3DRegion : public Resource {
 	void set_vertex_spacing(const real_t p_vertex_spacing) { _vertex_spacing = CLAMP(p_vertex_spacing, 0.25f, 100.f); }
 	real_t get_vertex_spacing() const { return _vertex_spacing; }
 
+	// File I/O
+	Error save(const String &p_path = "", const bool p_16_bit = false, const Image::CompressMode p_color_compression_mode = Image::COMPRESS_MAX);
```
TokisanGames: Header and body files should be sorted in the same order. Here you moved this out of order. It can go back where it was, below the working data.

Please also double check the other functions you added to ensure the bodies are in the same order as the declarations.

src/terrain_3d.cpp
```diff
@@ -1206,6 +1220,7 @@ void Terrain3D::_bind_methods() {
 	ADD_GROUP("Regions", "");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "region_size", PROPERTY_HINT_ENUM, "64:64,128:128,256:256,512:512,1024:1024,2048:2048", PROPERTY_USAGE_EDITOR), "change_region_size", "get_region_size");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "save_16_bit"), "set_save_16_bit", "get_save_16_bit");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "color_compression_mode", PROPERTY_HINT_ENUM, "S3TC,ETC,ETC2,BPTC,ASTC,NONE"), "set_color_compression_mode", "get_color_compression_mode");
```
TokisanGames: Let's make a subgroup under regions:
	ADD_SUBGROUP("Advanced", "");

Then place save_16_bit, color_compression_mode, and free_uncompressed_color_maps underneath it, in that order.

src/terrain_3d_region.cpp
TokisanGames: That sounds like a good idea to provide the option to free them on export. 

> For 1. would it be possible to free the color map already in set_color_map()?

I'm not sure what you mean. You are freeing the color map now.

TokisanGames (changes requested):
Nice job so far with the code and the testing. Running the demo shows the right format change and VRAM reduction.

Image

Image

Image

I have noticed one issue so far:

If compression mode is set and free_uncompressed_color_maps is false, it still frees the uncompressed colormap, when it shouldn't.

```
var region: Terrain3DRegion = $Terrain3D.data.get_region(Vector2i(0,0))
print("Color map: ", region.color_map)
print("Compressed color map: ", region.compressed_color_map)

color_compression = bptc
free_uncompressed_color_maps = false (same if true)
Color map: <Object#null>
Compressed color map: <Image#-9223371995985476204>

color_compression = none
free_uncompressed_color_maps = false (same if true)
Color map: <Image#-9223371996035807859>
Compressed color map: <Object#null>
```

You can also rebase and squash all of your commits.

stan4dbunny:
Hello! I have now fixed the changes you requested. Let me know if anything should be changed. 

Just wondering, but regarding squashing would it not be easier for you to squash the commits through the "squash and merge" GitHub option?

TokisanGames:
> Hello! I have now fixed the changes you requested. Let me know if anything should be changed.

Great, thank you. 

> Just wondering, but regarding squashing would it not be easier for you to squash the commits through the "squash and merge" GitHub option?

No, it needs to be squashed and rebased. Otherwise it will make a mess of the history. Github does one or other other, not both. And it doesn't rebase if there are conflicts, which is the case. 

There are currently conflicts and 25 commits. Please squash your commits down to two: code changes in one, and docs in the other. Then rebase onto the latest main and resolve any conflicts. The docs explain how, and I can help on discord. 

When I prepare it for merging I will tweak the code and docs, generate the rst files, and csharp bindings. It's helpful if I don't also have to squash, rebase, and fix conflicts myself.

src/terrain_3d.cpp
```diff
@@ -610,6 +610,35 @@ void Terrain3D::set_save_16_bit(const bool p_enabled) {
 	}
 }
 
+void Terrain3D::set_color_compression_mode(const ColorCompressMode p_color_compression_mode) {
+	SET_IF_DIFF(_color_compression_mode, p_color_compression_mode);
+	switch (_color_compression_mode) {
```
TokisanGames: Better to convert this on the very rare get godot version call, then we don't need to store the extra variable.

src/terrain_3d.cpp
```diff
@@ -610,6 +610,35 @@ void Terrain3D::set_save_16_bit(const bool p_enabled) {
...
+			break;
+		case COMPRESS_ASTC:
+			_image_color_compression_mode = Image::COMPRESS_ASTC;
+			break;
+		default:
+			_image_color_compression_mode = Image::COMPRESS_MAX;
+	}
+	LOG(INFO, "Setting compression mode for color maps: ", _image_color_compression_mode);
+	TypedArray<Terrain3DRegion> regions = _data->get_regions_active();
+	for (int i = 0; i < regions.size(); i++) {
```
TokisanGames: We now use ranged for loops
```
	for (Ref<Terrain3DRegion> region : regions) {
		region->set_modified(true);
	}
```

project/addons/terrain_3d/src/export_plugin.gd
```diff
@@ -0,0 +1,47 @@
...
+func _get_name() -> String:
+	return "Terrain3DExportPlugin"
+
+
+func _begin_customize_scenes(platform: EditorExportPlatform, features: PackedStringArray) -> bool:
+	return true
+	
+	
+func _customize_scene(scene: Node, path: String) -> Node:
+	var _terrain: Terrain3D = scene.find_child("Terrain3D", true)
```
TokisanGames: _ is our convention for class private variables like _hash.
terrain is local, so does not use _.

project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -7,6 +7,7 @@ extends EditorPlugin
 # Includes
 const Terrain3DUI: Script = preload("res://addons/terrain_3d/src/ui.gd")
 const RegionGizmo: Script = preload("res://addons/terrain_3d/src/region_gizmo.gd")
+const ExportPlugin = preload("res://addons/terrain_3d/src/export_plugin.gd")
```
TokisanGames: : Script

src/terrain_3d_data.h
```diff
@@ -122,7 +122,7 @@ class Terrain3DData : public Object {
 
 	// File I/O
 	void save_directory(const String &p_dir);
-	void save_region(const Vector2i &p_region_loc, const String &p_dir, const bool p_16_bit = false);
+	void save_region(const Vector2i &p_region_loc, const String &p_dir, const bool p_16_bit = false, const Image::CompressMode p_color_compression_mode = Image::COMPRESS_MAX);
```
TokisanGames: This should use our color compression mode.

src/terrain_3d.h
```diff
@@ -45,6 +45,15 @@ class Terrain3D : public Node3D {
 		SIZE_2048 = 2048,
 	};
 
+	enum ColorCompressMode {
```
TokisanGames: This is a property of the maps, so should be in Terrain3DRegion, along with the Image Format.

src/terrain_3d_data.cpp
```diff
@@ -449,7 +453,7 @@ void Terrain3DData::update_maps(const MapType p_map_type, const bool p_all_regio
 			Terrain3DRegion *region = get_region_ptr(region_loc);
 			// Generate all or only those marked edited
 			if (region && (p_all_regions || region->is_edited())) {
-				region->get_color_map()->generate_mipmaps();
+				region->get_map(TYPE_COLOR)->generate_mipmaps();
```
TokisanGames: Hmm, this cannot be done on a compressed map. So, it must be done on the uncompressed if valid or not at all.

src/terrain_3d_region.h
```diff
@@ -79,7 +81,10 @@ class Terrain3DRegion : public Resource {
 	Ref<Image> get_control_map() const { return _control_map; }
 	void set_color_map(const Ref<Image> &p_map);
 	Ref<Image> get_color_map() const { return _color_map; }
-	void sanitize_maps();
+	void set_compressed_color_map(const Ref<Image> &p_map);
+	Ref<Image> get_compressed_color_map() const { return _compressed_color_map; }
+	void free_uncompressed_color_map();
+	void sanitize_maps(const bool p_free_uncompressed_color_maps);
```
TokisanGames: This should have a default, rather than changing all usage cases of it.

TokisanGames (commented):
Looking better, thank you. I have now taken over this PR. Comments are for your edification only.

Pending: 
* Review Terrain3DRegion:
  * free_uncompressed_color_map still frees when disabled.
  * rework santisize_maps()

src/terrain_3d_region.cpp
```diff
@@ -164,6 +185,10 @@ void Terrain3DRegion::sanitize_maps() {
...
+		free_uncompressed_color_map();
+		return;
```
TokisanGames: I don't think this is the right function to clear the uncompressed maps. And returning here I'm pretty sure is a bug. The purpose of this function is to ensure the existing data is clean and safe to use, not to free data or optimize memory.

src/terrain_3d_region.cpp
```diff
@@ -79,6 +83,9 @@ Image *Terrain3DRegion::get_map_ptr(const MapType p_map_type) const {
 		case TYPE_CONTROL:
 			return *_control_map;
 		case TYPE_COLOR:
+			if (!IS_EDITOR && _compressed_color_map.is_valid()) {
+				return *_compressed_color_map;
+			}
```
TokisanGames: This isn't the right place for this either. The expectation should be if you ask for a color map you get the editable color map. We need a new function to get the active one which implies the one you get may be compressed and uneditable.

Other functions like set_maps(), get_maps(), get_data() don't even consider the compressed map. It needs to be all or nothing. I think it's cleaner to be nothing - so the compressed map is only created on save, freed in the editor, only accessed by get_compressed_color_map() or get_active_color_map(), everything else gives you the editable color map, though it may be null.

TokisanGames:
I have reorganized some of the code. Now:
* All functions that get the colormap, are unchanged, returning the editable colormap().
* Maps and functions all refer to `color map` and `compressed color map`. No more "uncompressed".
* get_active_color_map() returns the compressed map if not editor and valid, which update_maps() uses as read only.
* compressed maps are freed in the editor. Viewing a res file after it has been loaded will show an empty compressed map. Closing a scene and viewing the res will show the compressed and uncompressed map.
* color maps are freed in game if the option is chosen and the compressed map is valid.
* compress_color_map() is now available

Next, I will pull a commit from my dev mode PR to entirely disable the color map.

Then I'll remove free_color_map and instead introduce a color map mode with these options:
* Disabled - saved data is unchanged; editing is disabled; new regions create maps; compressed and uncompressed are freed in editor and game; 
* Cleared - disabled, and maps are cleared on save
* Editable - current - no compressed map saved or used, editable in game
* Compressed - editable in editor; compressed freed in editor; uncompressed freed in game and stripped on export. This will also enable the compression options.

Does this look like the full list of desired possibilities?

Xtarsia:
Have exported projects to web/mobile been tested for this yet? Ommiting to test those has caught me out before.

TokisanGames:
There might be an issue with compressed maps on linux.

https://discord.com/channels/691957978680786944/1065519581013229578/1463464825316180061

TokisanGames:
I have rebased and squashed this. I tried expanding the scope of this unnecessarily, including my work to outright disable the color map, making a dev mode, etc and then got hung up in my personal life. So I'm cutting all of that and going to merge in the original scope. I'll disable it by default and mark it experimental. If there's a problem with linux then we need more testers to confirm and troubleshoot. So this will go into 1.1.

Logging reverted to INFO.

Pending:
* [ ] More testing (Editing, game, exported)
* [ ] Final review
* [x] Complete docs
* [x] C# generation

TokisanGames:
I've been experimenting with the export plugin but have found it doesn't change the resource at all. The export always has the compressed and uncompressed maps. It's probably pulling from the file, regardless of what resource it receives from the plugin. So I've tuned up the export plugin, documented it, and commented it out as it has no effect if enabled.

Next, I'm removing the free_color_map option. Now at runtime it frees the uncompressed map if the compressed exists. I've adjusted some of the generic get color map options to return compressed if available.

I've also extracted the CompressMode enum into a global and it's much easier to work with. Now its values exactly match Godot's enum.

This just needs a little more testing and review and I think it will be ready for merging.

aidandavey:
I'm getting a crash when painting the color map, tested in 4.5.2 and 4.7.1 on Windows 11 

Load the demo scene 
Select Terrain3D node
Select color map tool 
Click on the terrain.

image

image

When I change the compression mode and run the game it plays fine and the texture memory is reduced as expected. ~12mb in the demo (3 regions) checks out.

Exported project also plays fine.

TokisanGames:
Thanks, I fixed that crash and found another, and then design issues.

I had a pretty solid design where the functions seamlessly switched between compressed and uncompressed. But then I learned neither Image::set_pixel() nor get_pixel() work on compressed images and both cause error spam. I thought get_pixel() worked and set_pixel() silently failed. Since that's not the case, I'm going to restructure a bit.

Most of the functions that retrieve the map all derive down to getting or setting a pixel, but now automatically choosing doesn't make sense anymore. With a compressed map, there's little you can do with it, except get it's size or format, so all of the info functions can't do anything with it. So I'm going to revert them all so they only work with the uncompressed map (if available). Then I'm going to add an option `keep_uncompressed_color` or something that avoids clearing it at runtime. That way a user could keep the color map in RAM to read data from, and the compressed map is in vram. They could even write to the color map (and maybe even display it if I allow a toggle). They just can't recompress, or read/write the compressed one. Back to the drawing board for this portion.

TokisanGames:
@aidandavey This can use another round of testing.

Now how it should work is:
* only one map, compressed or uncompressed is ever in VRAM, placed there by update_maps() calling get_active_color_map(). 
* set_pixel, get_pixel always use the uncompressed map, if available
* In editor mode the compressed is always freed from RAM, and uncompressed used. On save it's compressed to disk. 
* In game mode, it frees the unused color map from RAM, unless compress and keep_uncompressed_color are enabled, in which case it keeps both and sends compressed to VRAM. This allows reading color pixels with compressed vram, though both are in RAM. 
* The user could free either or both maps at runtime via the API, as long as they never call update_maps() when both have been freed. Full disabling is not supported yet. 

```
	//					Editor		Game
	//	No compression	uc -co		uc -co
	//	Compressed		uc -co		-uc co
	//	Comp+keep		uc -co		uc co
```

**I need to remove the printed output block on load before merging.**

PR #843, opened 2025-11-05
Bug fix: Toggling option save_16_bit sets all regions to modified

stan4dbunny:
The regions are now marked as "modified" when the option to save_16_bit is toggled.

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -545,6 +545,11 @@ void Terrain3D::set_region_size(const RegionSize p_size) {
 void Terrain3D::set_save_16_bit(const bool p_enabled) {
 	LOG(INFO, p_enabled);
 	_save_16_bit = p_enabled;
+	TypedArray<Terrain3DRegion> regions = _data->get_regions_active();
+	for (int i = 0; i < regions.size(); i++) {
+		Ref<Terrain3DRegion> region = regions[i];
+		region->set_modified(true);
+	}
 }
 
 void Terrain3D::set_label_distance(const real_t p_distance) {


src/terrain_3d.cpp
```diff
@@ -545,6 +545,13 @@ void Terrain3D::set_region_size(const RegionSize p_size) {
 void Terrain3D::set_save_16_bit(const bool p_enabled) {
 	LOG(INFO, p_enabled);
 	_save_16_bit = p_enabled;
+
+	Dictionary regions = _data->get_regions_all();
+	Array locs = regions.keys();
+	for (int i = 0; i < locs.size(); i++) {
+		const Terrain3DRegion *region = _data->get_region_ptr((Vector2i)locs[i]);
+		_data->set_region_modified(region->get_location(), true);
```
TokisanGames: You can just do `region->set_modified()`.

src/terrain_3d.cpp
```diff
@@ -545,6 +545,13 @@ void Terrain3D::set_region_size(const RegionSize p_size) {
 void Terrain3D::set_save_16_bit(const bool p_enabled) {
 	LOG(INFO, p_enabled);
 	_save_16_bit = p_enabled;
+
+	Dictionary regions = _data->get_regions_all();
```
TokisanGames: You can use `get_regions_active()`, which does the same thing as `keys()`, with region error checking, and only addresses non-deleted regions. Deleted regions won't be saved.

TokisanGames:
This can stay separate. It was a good choice to issue them as separate smaller PRs. I can reopen this if you cannot. Make the changes I specified, and also correct your commit user if you wish to have it tied to the right account. It has the same issue as the other. Right now it's not tied to a github account.

TokisanGames:
Thank you.

PR #845, opened 2025-11-07
Fix leaking CanvasItems on exit

aidandavey:
Fixes these errors on shutdown:
```
WARNING: 3 RIDs of type "CanvasItem" were leaked.
     at: _free_rids (servers/rendering/renderer_canvas_cull.cpp:2679)
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:2378)
ERROR: Cannot get path of node as it is not in a scene tree.
   at: (scene/main/node.cpp:2345)
Leaked instance: TextureButton:2323594170415 - Node path:
Leaked instance: Object:2323610947632
ERROR: Cannot get path of node as it is not in a scene tree.
   at: (scene/main/node.cpp:2345)
Leaked instance: TextureButton:2323929714755 - Node path:
Leaked instance: Object:2323946491972
ERROR: Cannot get path of node as it is not in a scene tree.
   at: (scene/main/node.cpp:2345)
Leaked instance: TextureButton:2324248481878 - Node path:
Leaked instance: Object:2324265259095
Hint: Leaked instances typically happen when nodes are removed from the scene tree (with `remove_child()`) but not freed (with `free()` or `queue_free()`).
```

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -735,16 +735,17 @@ class ListEntry extends MarginContainer:
 	var is_selected: bool = false
 	var is_highlighted: bool = false
 	
-	@onready var focus_style: StyleBox = get_theme_stylebox("focus", "Button").duplicate()
-	@onready var background: StyleBox = get_theme_stylebox("pressed", "Button")
 	var name_label: Label
 	var count_label: Label
-	@onready var button_row := FlowContainer.new()
-	@onready var button_clear := TextureButton.new()
-	@onready var button_edit := TextureButton.new()
-	@onready var spacer := Control.new()
-	@onready var button_enabled := TextureButton.new()
-	@onready var button_highlight := TextureButton.new()
+	var button_row: FlowContainer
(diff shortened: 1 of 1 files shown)


TokisanGames (changes requested):
Nice find.

However there is an issue here. ListContainer is recreated every time you switch between Textures and Meshes. So when you click Meshes and hover over an icon it prints:
```
SCRIPT ERROR: Cannot call method 'set_pressed_no_signal' on a previously freed instance.
          at: ListEntry._notification (res://addons/terrain_3d/src/asset_dock.gd:916)
```

We probably need to move setup_buttons() to enter_tree, as well as some or all of the @ onready vars. Maybe a setup function for all of that stuff.

aidandavey:
> We probably need to move setup_buttons() to enter_tree, as well as some or all of the @ onready vars. Maybe a setup function for all of that stuff.

Yes this fixed the error you found and I think it is only the Control nodes which need to be created and destroyed in this way. Everything else can stay in _ready()

TokisanGames:
Great! thank you.

PR #846, opened 2025-11-08
Changing init functions to be custom module friendly

Jackerty:
Custom module build expects that init functions end with module.

Also wrapping the GDExtension starter function to preprocessing to:
* avoid errors
* been unnessary for custom module build.


TokisanGames:
Thank you. I like that these are small and digestible PRs.

PR #865, opened 2025-11-16
Fix-#862-Regression-Texture-Highlighting-error

### LINKED ISSUE #862: Regression Texture Highlighting error
### Description

Caused by and 6bb2eb019efe86696076e9d8bdb1966168087f4b or 94f0b13200c635807aef14da02754eda735af68a

When changing the texture hightlight, this error is produced:

```
ERROR: Object '<Object#2808271163670>' was freed or unreferenced while a signal is being emitted from it. Try connecting to the signal using 'CONNECT_DEFERRED' flag, or use queue_free() to free the object (if this object is a Node) to avoid this error and potential crashes.
   at: ~Object (core/object/object.cpp:2167)
```

@aidandavey Could you take this on please?

aidandavey:
Fixes #862

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -922,9 +922,7 @@ class ListEntry extends MarginContainer:
 
 	func _notification(p_what) -> void:
 		match p_what:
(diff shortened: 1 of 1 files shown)


TokisanGames:
Great, thank you.

Issue #757, opened 2025-07-05
PBR Channel debug views


TokisanGames:
### Description

`Texture normal` isn't very useful placed on albedo. It would be much more useful just showing albedo grey with normals on NORMAL_MAP.

Review the debug views and group some of them into individual PBR channels that can be toggled on or off together:
ALBEDO - normal/grey
NORMAL_MAP
AO
ROUGHNESS

I think we have three categories of debug views now:
Overlays - grids, navigation, custom user data?
PBR channels
Debugging - jaggedness, control maps

Related #173

aidandavey:
Maybe as part of this work we could also tackle https://github.com/TokisanGames/Terrain3D/issues/173

TokisanGames:
I'm not sold on implementing it. Checkboxes are half as many clicks as a dropdown selection. Some can be shown simultaneously requiring checkboxes: overlays and PBR.

PR #868, opened 2025-11-19
Add Ao packed normals

Xtarsia:
Channel packer:

- Updates the channel packer to have tabs instead of 1 single pane.
- Adds the ability to *optionally* "pack" an AO texture into the normal map vectors.

Terrain:

- Removed old AO approximation.
- AO is now always extracted from the normal map, before it is re-normalized. Even when no AO is specifically packed, the nature of normal maps can somewhat align with the packing technique, and give AO-esq results anyways.
- AO strength is now applied directly to the extracted AO value, per ID, rather than accumulated.
- Added seperate "AO Light Effect" parameter.
- Demo textures updated to include AO packed normals.

image

Raw AO value as a temp debug view ( ALBEDO = vec3(mat.ao) ):
G6IWR_dWMAM0zr2

technical explanation:
AO is converted from linear to non-linear space, to boost precision at where it is most perceptible, then packed into a 0.5-1.0 range, so that when the normals get scaled, they are not crushed all the way to 0.0 length.

The entire process is inverted when AO is extracted, resulting in the same Linear AO values.

Normals are perceptually un-touched. AO has some precision loss where the effect is weakest.


src/shaders/main.glsl
```diff
@@ -578,11 +589,8 @@ void fragment() {
 	// Repack final normal map value.
 	NORMAL_MAP = fma(normalize(mat.normal_rough.xzy), vec3(0.5), vec3(0.5));
 	NORMAL_MAP_DEPTH = mat.normal_map_depth;
-
-	// Higher and/or facing up, less occluded.
-	float ao = (1. - (mat.albedo_height.a * log(2.1 - mat.ao_strength))) * (1. - mat.normal_rough.y);
-	AO = clamp(1. - ao * mat.ao_strength, mat.albedo_height.a, 1.0);
-	AO_LIGHT_AFFECT = (1.0 - mat.albedo_height.a) * clamp(mat.normal_rough.y, 0., 1.);
+	AO = clamp(mat.ao, 0., 1.);
+	AO_LIGHT_AFFECT = mat.ao_effect;
```
TokisanGames: mat.ao_effect should be mat.ao_affect.

src/terrain_3d_texture_asset.h
```diff
@@ -63,6 +64,9 @@ class Terrain3DTextureAsset : public Terrain3DAssetResource {
 	void set_ao_strength(const real_t p_ao_strength);
 	real_t get_ao_strength() const { return _ao_strength; }
 
+	void set_ao_light_effect(const real_t p_ao_light_effect);
+	real_t get_ao_light_effect() const { return _ao_light_effect; }
```
TokisanGames: ao_light_affect()

src/terrain_3d_util.h
```diff
@@ -46,10 +46,12 @@ class Terrain3DUtil : public Object {
 			const Vector2 &p_r16_height_range = Vector2(0.f, 255.f), const Vector2i &p_r16_size = V2I_ZERO);
 	static Ref<Image> pack_image(const Ref<Image> &p_src_rgb,
 			const Ref<Image> &p_src_a,
+			const Ref<Image> &p_src_ao,
 			const bool p_invert_green = false,
 			const bool p_invert_alpha = false,
 			const bool p_normalize_alpha = false,
-			const int p_alpha_channel = 0);
+			const int p_alpha_channel = 0,
+			const int p_occlusion_channel = 0);
```
TokisanGames: Maybe the bools should turn into a bitfield and moved to the end?

project/addons/terrain_3d/extras/shaders/lightweight.gdshader
```diff
@@ -45,6 +45,7 @@ uniform int _region_map[1024];
 uniform vec2 _region_locations[1024];
 uniform float _texture_normal_depth_array[32];
 uniform float _texture_ao_strength_array[32];
+uniform float _texture_ao_effect_array[32];
```
TokisanGames: affect

TokisanGames:
> Raw AO value as a temp debug view ( ALBEDO = vec3(mat.ao) ):

This can be a debug view. Eventually we want to have all of the PBR channels as debug views. #757

Xtarsia:
> It seems that both ao_scale and ao_affect need to be up high to do anything, and having either one down neutralizes the effect. Do they do different things? Do we need both settings instead of a combined one?

ao_scale adjusts the AO value directly, which is always applied to ambient light.
light affect, is how much the AO value is also applied to other lights (direct/spot/omni).

Scale will affect both results, but light affect will only apply to non-ambient light.

> Also it affects the distance. I wonder if distance scaled AO will be too strong with this now.

Yeah the mipmaping of the normals has the result of blending the AO as well, so applying a manual distance scale to AO wont be needed imo.

project/addons/terrain_3d/menu/channel_packer.tscn
```diff
@@ -532,4 +659,6 @@ alignment = 1
 layout_mode = 2
 text = "Close"
 
+[connection signal="ready" from="PanelContainer/MarginContainer/VBoxContainer/TabContainer" to="PanelContainer/MarginContainer/VBoxContainer/TabContainer" method="set_tab_title" flags=6 binds= [1, "Normal Roughness AO"]]
+[connection signal="tree_entered" from="PanelContainer/MarginContainer/VBoxContainer/TabContainer" to="PanelContainer/MarginContainer/VBoxContainer/TabContainer" method="set_tab_title" flags=6 binds= [0, "Albedo Height"]]
```
TokisanGames: I don't like using scene based signals, nor code in the scene files, which is very hard to see in git. I want all code in GDscript files, including signal connections so it can be easily found by git.
{B3DE1DF3-EF0E-4542-A932-FE4709915F96}

project/addons/terrain_3d/menu/channel_packer.gd
```diff
@@ -89,6 +92,12 @@ func pack_textures_popup() -> void:
 		window.find_child("RoughnessChannelB") as Button,
 		window.find_child("RoughnessChannelA") as Button
 	]
+	occlusion_channel = [
+		window.find_child("RoughnessChannelR") as Button,
+		window.find_child("RoughnessChannelG") as Button,
+		window.find_child("RoughnessChannelB") as Button,
+		window.find_child("RoughnessChannelA") as Button
```
TokisanGames: Should be OcclusionChannelR/G/B/A

project/addons/terrain_3d/menu/channel_packer.gd
```diff
@@ -207,7 +227,7 @@ func _init_texture_picker(p_parent: Node, p_image_index: int) -> void:
 			_set_wh_labels(p_image_index, image.get_width(), image.get_height())
 			if p_image_index == IMAGE_NORMAL:
 				_set_normal_vector(image)
-			if p_image_index == IMAGE_HEIGHT or p_image_index == IMAGE_ROUGHNESS:
+			if p_image_index == IMAGE_HEIGHT or p_image_index == IMAGE_ROUGHNESS or p_image_index == IMAGE_AO:
```
TokisanGames: `if p_image_index in [ IMAGE_HEIGHT, IMAGE_ROUGHNESS, IMAGE_AO]:`

TokisanGames (commented):
Great, thanks for this. 

I appreciate the organizational effort. Though I don't think tabs works here. The packing button will respond to either albedo or normal sets or both, so it's awkward to have it visible below and act on the invisible tab. I reworked the panel to include a scroll container and a flow container. Now the three segments are responsive and scroll for small displays.

I've taken this over. Almost ready to merge.

TokisanGames:
I'll merge in the PBR as is, but I don't think it achieves #757. Let's discuss if this is useful as is, or my alternative.

I was thinking PBR checkboxes would all default to be on. Then you could disable AO, Normal, Roughness, Albedo (to .2 grey), and pick and choose which channels you're viewing.

TokisanGames:
Great, thank you!

PR #870, opened 2025-11-20
Scale normal depth by distance

### LINKED ISSUE #857: Scale Normals, AO by distance
### Description

Scale the normals and AO by distance to reduce the look of shadows or details fading in the distance.

MopanionStudios:
Admin edit

This PR adds:
* Normal scaling, default on set to 2
* Light direction/color to the terrain shader (using the same light target as the ocean shader)
* [Specular suppression](https://github.com/TokisanGames/Terrain3D/pull/870#issuecomment-4968418686) for terrain faces that aren't facing the sun light

Normal scaling:

Before
{6FDAB0DF-CB85-430C-88AB-9869AA07D58C}

After set to 10 (2-4 is best)
{EBB6C664-3C96-4CF8-91AB-FE6BB5147179}

-------

Fixes #857 

You can close this pull request if you like, as I understand I am not assigned to the issue, but I implemented a lightweight method using Xstarsia's plan to use dFdx/dFdy space for scaling the AO and normals intensity depending on their distance to the camera.

I allow for the effect's intensity to be adjusted by a shader parameter, and clamp it to 1.0 at minimum so that it doesn't go below the normal depth / ao original values. It adds a subtle but noticeable difference that improves the visuals, as seen here:

Screenshot 2025-11-19 230509
Screenshot 2025-11-19 230729

### DIFF
--- a/src/shaders/main.glsl
+++ b/src/shaders/main.glsl
@@ -72,14 +72,16 @@ uniform highp sampler2DArray _control_maps : repeat_disable;
 uniform highp sampler2DArray _color_maps : source_color, FILTER_METHOD, repeat_disable;
 uniform highp sampler2DArray _texture_array_albedo : source_color, FILTER_METHOD, repeat_enable;
 uniform highp sampler2DArray _texture_array_normal : hint_normal, FILTER_METHOD, repeat_enable;
+uniform vec3 _light_direction = vec3(0., 0., 0);
+uniform vec3 _light_color : source_color = vec3(1.0, 1.0, .735);
 group_uniforms;
 
 // Public uniforms
 group_uniforms shader_uniforms.general;
 //INSERT: FLAT_UNIFORMS
 uniform bool flat_terrain_normals = false;
+uniform float distant_normal_scale : hint_range(1.0, 10.0, 0.1) = 2.0;
 uniform float blend_sharpness : hint_range(0, 1) = 0.5;
-//INSERT: PROJECTION_UNIFORMS
 group_uniforms;
 
 //INSERT: AUTO_SHADER_UNIFORMS
@@ -90,7 +92,6 @@ group_uniforms;
 group_uniforms shader_uniforms.mipmaps;
 uniform float bias_distance : hint_range(0.0, 16384.0, 0.1) = 512.0;
 uniform float mipmap_bias : hint_range(0.5, 1.5, 0.01) = 1.0;
-uniform float distant_normal_scale : hint_range(1.0, 10.0, 0.1) = 2.0;
 uniform float depth_blur : hint_range(0.0, 35.0, 0.1) = 0.0;
 group_uniforms;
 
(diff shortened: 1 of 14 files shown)


MopanionStudios:
Just updated/fixed the distance scaling, before it was utilizing UV space which goes from 0-1 in regions. To make the scaling more reliable, it now uses v_vertex, I also changed the name of the uniform float controlling the effect to make its purpose more clear. Heres an updated example of what the effect does:

https://github.com/user-attachments/assets/89ee6bad-4f39-47da-b287-a8814a7a59fc

src/shaders/main.glsl
```diff
@@ -570,17 +571,20 @@ void fragment() {
...
+	float vdF = clamp(length(vec2(dFdx(v_vertex.x), dFdx(v_vertex.z))), 1.0, distant_detail_amplifier);
 	
 	// Apply PBR
 	ALBEDO = mat.albedo_height.rgb * color_map.rgb * macrov;
 	ROUGHNESS = roughness;
 	SPECULAR = 1. - mat.normal_rough.a;
 	// Repack final normal map value.
 	NORMAL_MAP = fma(normalize(mat.normal_rough.xzy), vec3(0.5), vec3(0.5));
-	NORMAL_MAP_DEPTH = mat.normal_map_depth;
+	NORMAL_MAP_DEPTH = mat.normal_map_depth * vdF;
```
Xtarsia: as a note, the reason this is needed, is due to mipmaps flattening normals during the mipmap generation process.

however im not sure we actually need AO boosting, as the mipmap generation process, will actually make the AO boost how we want anyways - and with the pending PR for texture based AO rather than the current rough estimate, shows that remains the case with the normal packing method when I did some quick testing of that.

src/shaders/main.glsl
```diff
@@ -570,17 +571,20 @@ void fragment() {
 	
 	// Wetness/roughness modifier, converting 0 - 1 range to -1 to 1 range, clamped to Godot roughness values 
 	float roughness = clamp(fma(color_map.a - 0.5, 2.0, mat.normal_rough.a), 0., 1.);
+
+	// vertex in dFdx space to compare distance to camera
+	float vdF = clamp(length(vec2(dFdx(v_vertex.x), dFdx(v_vertex.z))), 1.0, distant_detail_amplifier);
```
Xtarsia: Full vec3 derivatives for v_vertex are called at the start of the shader, and stored in base_ddx and base_ddy.

You can use those rather than calling dFdx() again on single components.

also, using both screenspace x and y derivatives of xz, rather than each component seperatley will give a more stable result, which can seen by testing, and moving the camera around, especially rotation whilst looking down.

``float vdF = clamp(max(length(base_ddx.xz), length(base_ddy.xz)), 1.0, distant_detail_amplifier);``

rename vdF to something more descriptive too, normal_boost_factor for example.

src/shaders/main.glsl
```diff
@@ -74,6 +74,7 @@ uniform highp sampler2DArray _control_maps : repeat_disable;
 group_uniforms general;
 uniform bool flat_terrain_normals = false;
 uniform float blend_sharpness : hint_range(0, 1) = 0.5;
+uniform float distant_detail_amplifier : hint_range(1.0, 5.0) = 1.5;
```
Xtarsia: This can go into the ``group_uniforms mipmaps;`` section.

Xtarsia (changes requested):

src/shaders/main.glsl
MopanionStudios: May I ask, why is max() added in the clamp() you put in the comment? Wouldn't we want to boost it by both ddx and ddy, regardless of which has a larger change in value?

edit: Nevermind, looking more into mipmap generation answered that for me. Mipmaps choose the greater change of the two axes for the mipmapping, so it would make sense to use max here to combat which is being more affected by the mipmaps.

MopanionStudios: For instance, doing something like this instead to get the median of the two rather than whichever has a larger change in value:
`float vdF = clamp((length(base_ddx.xz) + length(base_ddy.xz)) / 2., 1.0, distant_detail_amplifier);`

src/shaders/main.glsl
TokisanGames: What this does is scale AO/Normal strength by distance, rather than amplify distant details, which is done more by mipmap_bias. It should be named something like `distant_normal_scale`.

MopanionStudios:
@Xtarsia Hey, sorry it took me so long. Cleaned up & rebased this branch. Should be ready for merging.

TokisanGames:
I tested this today. While it looks good in the light, unfortunately the normals cause specular to reflect light where it should be occluded. This artifact was already present, but this PR makes it significantly worse. 

image

Unfortunately since the land will occlude the sun a significant time, this is unusable as is. Fortunately, we've already accepted a light target for the ocean shader, so it's not a big deal to also add it to the terrain shader. Then we can filter the effect and other existing artifacts based on lit faces.

So comparing the shader before this PR:

{3275CE92-DEFC-419C-95A3-71AB810E413F}

And now with damping specular based on light direction:

{546EB260-E51A-4EA6-9314-F1C8A1C9C906}


@MopanionStudios Please [enable this option](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) so I can write to this PR. It was set on by default so you would have had to turn it off.

{D0DA6EF6-1EDB-4AA3-9EE4-4A0FE4B172D5}

MopanionStudios:
@TokisanGames Alright, enabled it so you should have access to editing the PR! Didn't even notice that issue w/ specular, good catch

TokisanGames:
Nice work @MopanionStudios. Thank you.

PR #872, opened 2025-11-20
Add ground_level, negative world noise height

### LINKED ISSUE #776: When using World Noise, allow negative heights
In my game, there is a lot of underwater landscapes, and some islands. I'm using the world background to fill the ocean floor for generic areas, but the default shader [doesn't allow](https://github.com/TokisanGames/Terrain3D/blob/8307e97e957e6adb103a846b105651978968d268/src/shaders/world_noise.glsl#L13) negative height values. I tried to offset the noise but the result wasn't good, it creates some plateaus. I modified the shader to allow negative heights and the result is now perfect.

Also the [world_noise_region_blend](https://github.com/TokisanGames/Terrain3D/blob/8307e97e957e6adb103a846b105651978968d268/src/shaders/world_noise.glsl#L8) could allow the value from range 0.0 to 1.0

TokisanGames:
Fixes #776 

Prepares the way for an ocean mesh.

* Adds `ground_level` and `region_blend` to flat background and lightweight.
* Allows negative `world_noise_height`. Ground level is already adjustable in world_noise_height or offset.

### DIFF
--- a/Terrain3D.vcxproj
+++ b/Terrain3D.vcxproj
@@ -224,7 +224,7 @@
     <None Include="src\shaders\main.glsl" />
     <None Include="src\shaders\overlays.glsl" />
     <None Include="src\shaders\samplers.glsl" />
-    <None Include="src\shaders\world_noise.glsl" />
+    <None Include="src\shaders\backgrounds.glsl" />
     <None Include="src\shaders\editor_functions.glsl" />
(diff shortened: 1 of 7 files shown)


TokisanGames:
Great idea. Thanks.

Background conditionals aren't necessary for backgrounds.glsl since it's only injected if true.

PR #873, opened 2025-11-23
Fix #856 - Update Transforms isn't working

### LINKED ISSUE #856: Update Transforms isn't working
Painting instances then sculpting should immediately update the instance position as the terrain moves.
In the demo even in one sculpting stroke, across (0, -2) and (0, -1), it appears to update instances in the former immediately, while in the latter they are not updated until the mouse moves to a distant cell, then they snap. This suggests the transforms are being updated, but not being redrawn for that region. 

It sometimes is limited per asset, as well as per region. So one cell with several types of meshes might update transforms of every other mesh type and ignore one.

I suspect it is accurately updating the transforms in the `instances` dictionary, and that the problem is downstream of update_mmis(), destroy_*, _process_updates(). 

Or perhaps the caller of update_mmis() is not passing along the correct ID and/or region to be updated, which would be a UI/Terrain3DEditor issue.
`update_mmis(const int p_mesh_id, const Vector2i &p_region_loc, const bool p_rebuild)`

----
~~Around this time painting instances then sculpting won't update the transforms. Sometimes it's per region or per asset.~~
~~Adding another asset is clearing the asset buttons and not redisplaying them.~~

aidandavey:
Fixes #856

When updating the instancer following sculpting, we were passing the index of the current position in the mesh array, not the actual mesh id.

### DIFF
--- a/project/project.godot
+++ b/project/project.godot
@@ -12,7 +12,7 @@ config_version=5
 
 config/name="Terrain3D"
 run/main_scene="res://demo/Demo.tscn"
-config/features=PackedStringArray("4.4", "C#")
+config/features=PackedStringArray("4.4")
 config/icon="res://icon.png"
 
 [display]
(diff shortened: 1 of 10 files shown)


TokisanGames:
Nice find. Thinking along the same lines, I also found a couple other bugs and style inconsistencies as I converted most for loops to range based.

PR #881, opened 2025-12-03
Ocean mesh

aidandavey:
This PR adds a second clipmap and exposes the material, so that it can be used as an ocean mesh. The Mesher has been reworked so that it is easier to add further clipmaps if desired (shadow imposters, ceilings?)

I have included an example ocean shader - which needs improvement. The LOD edges are very visible from above.

### DIFF
--- a/src/terrain_3d.h
+++ b/src/terrain_3d.h
@@ -79,6 +79,14 @@ private:
 	int _tessellation_level = 0;
 	real_t _vertex_spacing = 1.0f;
 
+	// Ocean meshes
+	int _ocean_mesh_lods = 7;
+	int _ocean_mesh_size = 48;
+	int _ocean_tessellation_level = 0;
+
+	// Ocean material
+	Ref<Material> _ocean_material;
+
 	// Rendering
 	uint32_t _render_layers = 1u | (1u << 31u); // Bit 1 and 32 for the cursor
 	RenderingServer::ShadowCastingSetting _cast_shadows = RenderingServer::SHADOW_CASTING_SETTING_ON;
@@ -194,6 +202,19 @@ public:
 	bool is_buffer_shader_override_enabled() const { return _material.is_valid() ? _material->is_buffer_shader_override_enabled() : false; }
 	void set_buffer_shader_override(const Ref<Shader> &p_shader) { return _material.is_valid() ? _material->set_buffer_shader_override(p_shader) : void(); }
 	Ref<Shader> get_buffer_shader_override() const { return _material.is_valid() ? _material->get_buffer_shader_override() : Ref<Shader>(); }
+
+	// Ocean Mesh
+	void set_ocean_enabled(const bool p_enabled);
+	bool is_ocean_enabled() const { return _mesher ? _mesher->is_ocean_enabled() : false; }
+	void set_ocean_mesh_lods(const int p_count);
+	int get_ocean_mesh_lods() const { return _ocean_mesh_lods; }
+	void set_ocean_mesh_size(const int p_size);
+	int get_ocean_mesh_size() const { return _ocean_mesh_size; }
+	void set_ocean_tessellation_level(const int p_level);
+	int get_ocean_tessellation_level() const { return _ocean_tessellation_level; }
+	Ref<Material> get_ocean_material() const { return _ocean_material; }
+	void set_ocean_material(const Ref<Material> &p_material);
+	RID get_ocean_material_rid() const { return _ocean_material.is_valid() ? _ocean_material->get_rid() : RID(); }
 
 	// Rendering
 	void set_render_layers(const uint32_t p_layers);
--- a/src/terrain_3d_mesher.cpp
+++ b/src/terrain_3d_mesher.cpp
@@ -12,34 +12,33 @@
 // Private Functions
 ///////////////////////////
 
-void Terrain3DMesher::_generate_mesh_types(const int p_size) {
-	_clear_mesh_types();
+void Terrain3DMesher::_generate_mesh_types(Array &p_mesh_rids_array, const int p_size, const RID &p_mat) {
 	LOG(INFO, "Generating all Mesh segments for clipmap of size ", p_size);
 	// Create initial set of Mesh blocks to build the clipmap
 	// # 0 TILE - mesh_size x mesh_size tiles
-	_mesh_rids.push_back(_generate_mesh(V2I(p_size)));
+	p_mesh_rids_array.push_back(_generate_mesh(V2I(p_size), false, p_mat));
 	// # 1 EDGE_A - 2 by (mesh_size * 4 + 8) strips to bridge LOD transitions along +-Z axis
-	_mesh_rids.push_back(_generate_mesh(Vector2i(2, p_size * 4 + 8)));
+	p_mesh_rids_array.push_back(_generate_mesh(Vector2i(2, p_size * 4 + 8), false, p_mat));
 	// # 2 EDGE_B - (mesh_size * 4 + 4) by 2 strips to bridge LOD transitions along +-X axis
-	_mesh_rids.push_back(_generate_mesh(Vector2i(p_size * 4 + 4, 2)));
+	p_mesh_rids_array.push_back(_generate_mesh(Vector2i(p_size * 4 + 4, 2), false, p_mat));
 	// # 3 FILL_A - 4 by mesh_size
-	_mesh_rids.push_back(_generate_mesh(Vector2i(4, p_size)));
+	p_mesh_rids_array.push_back(_generate_mesh(Vector2i(4, p_size), false, p_mat));
 	// # 4 FILL_B - mesh_size by 4
-	_mesh_rids.push_back(_generate_mesh(Vector2i(p_size, 4)));
+	p_mesh_rids_array.push_back(_generate_mesh(Vector2i(p_size, 4), false, p_mat));
 	// # 5 STANDARD_TRIM_A - 2 by (mesh_size * 4 + 2) strips for LOD0 +-Z axis edge
-	_mesh_rids.push_back(_generate_mesh(Vector2i(2, p_size * 4 + 2), true));
+	p_mesh_rids_array.push_back(_generate_mesh(Vector2i(2, p_size * 4 + 2), true, p_mat));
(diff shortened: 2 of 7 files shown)


src/terrain_3d_mesher.h
```diff
@@ -63,10 +68,15 @@ class Terrain3DMesher {
 	void initialize(Terrain3D *p_terrain);
 	void destroy();
 
+	bool get_ocean_enabled() const { return _ocean_enabled; }
```
TokisanGames: Can be is_ocean_enabled(), and sorted below the setter.

src/terrain_3d_mesher.h
```diff
@@ -63,10 +68,15 @@ class Terrain3DMesher {
...
+
 	void snap();
+	void _update_mesh_transforms(Array _instance_rids, real_t vertex_spacing, Vector3 snapped_pos);
 	void reset_target_position() { _last_target_position = V2_MAX; }
 	void update();
 	void update_aabbs();
+	void update_ocean_material_override(Ref<Material> p_mat);
```
TokisanGames: Use `const Ref<> &`
Rename to `set_ocean_material()`. In this case we aren't overriding.

src/terrain_3d_mesher.h
```diff
@@ -63,10 +68,15 @@ class Terrain3DMesher {
 	void initialize(Terrain3D *p_terrain);
 	void destroy();
 
+	bool get_ocean_enabled() const { return _ocean_enabled; }
+
 	void snap();
+	void _update_mesh_transforms(Array _instance_rids, real_t vertex_spacing, Vector3 snapped_pos);
```
TokisanGames: Use `(const Array &p_instance_rids, const real_t p_vertex_spacing, const Vector3 &p_snapped_pos)`

src/terrain_3d.h
```diff
@@ -195,6 +198,13 @@ class Terrain3D : public Node3D {
 	void set_buffer_shader_override(const Ref<Shader> &p_shader) { return _material.is_valid() ? _material->set_buffer_shader_override(p_shader) : void(); }
 	Ref<Shader> get_buffer_shader_override() const { return _material.is_valid() ? _material->get_buffer_shader_override() : Ref<Shader>(); }
 
+	// Ocean
+	void set_ocean_enabled(const bool p_enabled) { _mesher ? _mesher->set_ocean_enabled(p_enabled) : void(); }
+	bool get_ocean_enabled() const { return _mesher ? _mesher->get_ocean_enabled() : false; }
+	Ref<Material> get_ocean_material() const { return _ocean_material; }
+	void set_ocean_material(const Ref<Material> &p_material);
+	RID get_ocean_material_rid() const { return _ocean_material.is_valid() ? _ocean_material->get_rid() : RID(); }
```
TokisanGames, line 217: How about an option to set the ocean mesh tessellation level? You said tessellation was important for the ocean? Maybe they want different values for land and sea.

Maybe also options to specify size and lods since the options are there in the function.

src/terrain_3d_mesher.h
```diff
@@ -47,14 +51,15 @@ class Terrain3DMesher {
 	real_t _offset_c = 0.f;
 	PackedVector3Array _edge_pos;
 
-	void _generate_mesh_types(const int p_mesh_size);
-	RID _generate_mesh(const Vector2i &p_size, const bool p_standard_grid = false);
-	RID _instantiate_mesh(const PackedVector3Array &p_vertices, const PackedInt32Array &p_indices, const AABB &p_aabb);
-	void _generate_clipmap(const int p_size, const int p_lods, const RID &scenario);
+	void _generate_mesh_types(const int p_size, Array &p_mesh_rids_array, const RID mat);
+	RID _generate_mesh(const Vector2i &p_size, const bool p_standard_grid = false, const RID mat = RID());
+	RID _instantiate_mesh(const PackedVector3Array &p_vertices, const PackedInt32Array &p_indices, const AABB &p_aabb, const RID mat);
```
TokisanGames: `const RID &p_mat` on all.
Use & on anything larger than a pointer, or any object. RID is 8 byte, same as a pointer, but it doesn't hurt to maintain the consistent style.

src/terrain_3d_mesher.cpp
```diff
@@ -12,34 +12,33 @@
 // Private Functions
 ///////////////////////////
 
-void Terrain3DMesher::_generate_mesh_types(const int p_size) {
-	_clear_mesh_types();
+void Terrain3DMesher::_generate_mesh_types(const int p_size, Array &p_mesh_rids_array, const RID mat) {
```
TokisanGames: p_mat

src/terrain_3d_mesher.h
```diff
@@ -47,14 +51,15 @@ class Terrain3DMesher {
...
+	void _generate_clipmap(const int p_size, const int p_lods, const RID &p_scenario, Array &p_instance_rids, Array &p_mesh_rids, RID p_material_rid);
```
TokisanGames: If we're going to have modified array parameters, let's move them to the beginning of the list. So the format would be:

`func(&p_insertion_array1, &p_insertion2, const p_param1, const p_param2, const p_default_param1 = 1, const p_defparam2 = 2)`

src/terrain_3d_mesher.cpp
```diff
@@ -251,28 +255,30 @@ void Terrain3DMesher::_generate_offset_data(const int p_size) {
...
 
 // Frees all clipmap instance RIDs. Mesh rids must be freed seperatley.
-void Terrain3DMesher::_clear_clipmap() {
+void Terrain3DMesher::_clear_clipmap(Array &p_instance_rids) {
 	LOG(INFO, "Freeing all clipmap instances");
-	for (Array lod_array : _clipmap_rids) {
-		for (Array mesh_array : lod_array) {
-			for (const RID &rid : mesh_array) {
-				RS->free_rid(rid);
+	for (int lod = 0; lod < p_instance_rids.size(); lod++) {
```
TokisanGames: Why can't you use the previous range style for loop here?

src/terrain_3d_mesher.cpp
```diff
@@ -329,24 +340,33 @@ void Terrain3DMesher::snap() {
 	// Recenter terrain on the target
 	_last_target_position = target_pos_2d;
 	Vector3 snapped_pos = (target_pos / vertex_spacing).floor() * vertex_spacing;
+	RS->material_set_param(_terrain->get_material()->get_material_rid(), "_camera_pos", snapped_pos);
```
TokisanGames: Is this old? I think _camera_pos was removed.

src/terrain_3d_mesher.cpp
```diff
@@ -329,24 +340,33 @@ void Terrain3DMesher::snap() {
...
+	_update_mesh_transforms(_clipmap_rids, vertex_spacing, snapped_pos);
+	if (_ocean_enabled && _terrain->get_ocean_material_rid().is_valid()) {
+		RS->material_set_param(_terrain->get_ocean_material_rid(), "_camera_pos", snapped_pos);
```
TokisanGames: Probably should be _target_pos now.

src/terrain_3d_mesher.cpp
```diff
@@ -329,24 +340,33 @@ void Terrain3DMesher::snap() {
...
+		_update_mesh_transforms(_ocean_clipmap_rids, vertex_spacing, snapped_pos);
+	}
+}
+
+void Terrain3DMesher::_update_mesh_transforms(Array p_instance_rids, real_t p_vertex_spacing, Vector3 p_snapped_pos) {
```
TokisanGames: All can be const.

src/terrain_3d_mesher.cpp
```diff
@@ -463,5 +496,32 @@ void Terrain3DMesher::update_aabbs() {
...
+	}
 	return;
 }
+
+void Terrain3DMesher::set_ocean_enabled(const bool p_enabled) {
+	SET_IF_DIFF(_ocean_enabled, p_enabled);
+	destroy();
+	initialize(_terrain);
+	_last_target_position = V2_MAX;
+	snap();
```
TokisanGames: These last two are already done in initialize.

src/terrain_3d_mesher.cpp
```diff
@@ -251,28 +255,30 @@ void Terrain3DMesher::_generate_offset_data(const int p_size) {
...
 	return;
 }
 
 // Frees all Mesh RIDs use for clipmap instances.
-void Terrain3DMesher::_clear_mesh_types() {
+void Terrain3DMesher::_clear_mesh_types(Array &p_mesh_rids) {
 	LOG(INFO, "Freeing all clipmap meshes");
-	for (const RID &rid : _mesh_rids) {
-		RS->free_rid(rid);
+	for (const RID mesh_rid : p_mesh_rids) {
```
TokisanGames: Can you use &?

src/terrain_3d.h
```diff
@@ -195,6 +198,13 @@ class Terrain3D : public Node3D {
 	void set_buffer_shader_override(const Ref<Shader> &p_shader) { return _material.is_valid() ? _material->set_buffer_shader_override(p_shader) : void(); }
 	Ref<Shader> get_buffer_shader_override() const { return _material.is_valid() ? _material->get_buffer_shader_override() : Ref<Shader>(); }
 
+	// Ocean
+	void set_ocean_enabled(const bool p_enabled) { _mesher ? _mesher->set_ocean_enabled(p_enabled) : void(); }
+	bool get_ocean_enabled() const { return _mesher ? _mesher->get_ocean_enabled() : false; }
```
TokisanGames: is_ocean_enabled()

src/terrain_3d.h
aidandavey: Tessellation is definitely important as it looked pretty terrible before. I didn't do anything to make use of the new tesselation and am not sure if and how I could control it independently for this clipmap. I'll see if I can work it out.

TokisanGames: You don't need the displacement buffer or anything fancy. All I'm suggesting is separate mesh size, lods, and tessellation parameters, which will be fed into _generate_clipmap(). Maybe even vertex_spacing, fed into update_mesh_transforms, but not sure for now.

aidandavey:
New PR incoming

PR #886, opened 2025-12-04
Better Grass Shader implementation

rpgshooter:
Pretty traditional implementation to add and improve to the existing grass shader with a subscatter feature making the grass have a more realistic look


rpgshooter:
If wanted, i can push the rest of the shader stuff, specifically the normal bend, better wind, better size clustering and better aso. before the merge of just the ss

rpgshooter:
updated name, Going to implement full grass shader system then push for review

TokisanGames:
Did you intend to close this? I see you pushed this from `main` which causes many problems.

rpgshooter:
> Did you intend to close this? I see you pushed this from `main` which causes many problems.

Yes, swapped from main to sub branch to prevent said issues will re-pr once more progress is done

Issue #298, opened 2024-01-14
Water System


anandbarai09:
## Update
Admin edit

Future updates and a full water system could be made up of these components:

* [x] Ocean clipmap mesh #887 
* [ ] River system (Use waterways, perhaps we incorporate it)
* [ ] Custom map #565 to paint where the ocean/rivers are within regions

The current ocean shader could be improved with:
* [ ] Detiling in the wave pattern in the middle distance, without costing more
* [ ] Add Caustics
* [ ] Fix refraction artifacts
* [ ] Geomorph code injection

### Original

We can create a shader for ocean and then set it in WorldBackground property in Terrian3DMaterial section. And make thing float with some lines of code as shown in the video. https://youtu.be/_R2KDcAp1YQ?si=A2o_jcG4XJmZJEx4

OR

Make a way to create custom WorldBackground shader so user can implement this by their own.

This would be helpful to create an Islands.

TokisanGames:
An ocean can currently be implemented separately and independent of Terrain3D.

World background is just a shader, so a user could already implement it on their own using the Terrain3D mesh.

If we implement an ocean shader in this repo, water should be a separate clipmap layer, and its location paintable so it can be a lake, but that's a ways off.

anandbarai09:
> An ocean can currently be implemented separately and independent of Terrain3D.
> 
> World background is just a shader, so a user could already implement it on their own using the Terrain3D mesh.
> 
> If we implement an ocean shader in this repo, water should be a separate clipmap layer, and its location paintable so it can be a lake, but that's a ways off.

I would like to make a clipmap ocean system, so any idea how to implement??

TokisanGames:
> I would like to make a ocean system addon, so any idea how to implement??

Do you want to add an ocean to your game or do you want to make a PR for Terrain3D to build in an ocean?

If the first, you don't need to modify terrain3d or the shader at all. Just make a large plane node with the water shader that overlaps Terrain3D, then you can set our worldbackground to none. The water node will intersect with the terrain at every point on every curve. There will be some redundant meshes but it's easy to implement.

If you want to make a PR, then you should do what I said in my first comment. 

Putting the oceanshader on the worldbackground flat mesh is going to be challenging to do. You'll run into problems blending with the rest of the world smoothly along the curves of the coast line.

Here's what should happen in phase 1:
* A second clipmap mesh is generated at a certain height above the terrain. This height is user settable
* The water clipmap gets the water shader
[...]

PR #887, opened 2025-12-04
Add ocean mesh

aidandavey:
Adds a secondary clipmap dedicated to the ocean. The ocean now uses an independent mesher/clipmap so ocean LOD, mesh size and tessellation can be configured and updated separately from terrain.

--
Partially addresses #298

### DIFF
--- a/project/addons/terrain_3d/csharp/Terrain3D.cs
+++ b/project/addons/terrain_3d/csharp/Terrain3D.cs
@@ -157,18 +157,29 @@ public partial class Terrain3D : Node3D
 		public new static readonly StringName PhysicsMaterial = "physics_material";
 		public new static readonly StringName ClipmapTarget = "clipmap_target";
 		public new static readonly StringName MeshLods = "mesh_lods";
+		public new static readonly StringName TessellationLevel = "tessellation_level";
 		public new static readonly StringName MeshSize = "mesh_size";
 		public new static readonly StringName VertexSpacing = "vertex_spacing";
-		public new static readonly StringName TessellationLevel = "tessellation_level";
+		public new static readonly StringName CullMargin = "cull_margin";
+		public new static readonly StringName CastShadows = "cast_shadows";
+		public new static readonly StringName GiMode = "gi_mode";
+		public new static readonly StringName RenderLayers = "render_layers";
 		public new static readonly StringName DisplacementScale = "displacement_scale";
 		public new static readonly StringName DisplacementSharpness = "displacement_sharpness";
 		public new static readonly StringName BufferShaderOverrideEnabled = "buffer_shader_override_enabled";
 		public new static readonly StringName BufferShaderOverride = "buffer_shader_override";
-		public new static readonly StringName RenderLayers = "render_layers";
+		public new static readonly StringName OceanEnabled = "ocean_enabled";
+		public new static readonly StringName OceanMeshLods = "ocean_mesh_lods";
+		public new static readonly StringName OceanTessellationLevel = "ocean_tessellation_level";
+		public new static readonly StringName OceanMeshSize = "ocean_mesh_size";
+		public new static readonly StringName OceanVertexSpacing = "ocean_vertex_spacing";
+		public new static readonly StringName OceanCullMargin = "ocean_cull_margin";
+		public new static readonly StringName OceanCastShadows = "ocean_cast_shadows";
+		public new static readonly StringName OceanGiMode = "ocean_gi_mode";
+		public new static readonly StringName OceanRenderLayers = "ocean_render_layers";
+		public new static readonly StringName OceanMaterial = "ocean_material";
+		public new static readonly StringName OceanLightTarget = "ocean_light_target";
 		public new static readonly StringName MouseLayer = "mouse_layer";
-		public new static readonly StringName CastShadows = "cast_shadows";
-		public new static readonly StringName GiMode = "gi_mode";
-		public new static readonly StringName CullMargin = "cull_margin";
 		public new static readonly StringName FreeEditorTextures = "free_editor_textures";
 		public new static readonly StringName InstancerMode = "instancer_mode";
 		public new static readonly StringName ShowRegionGrid = "show_region_grid";
@@ -329,6 +340,12 @@ public partial class Terrain3D : Node3D
 		set => Set(GDExtensionPropertyName.MeshLods, value);
 	}
 
+	public new long TessellationLevel
+	{
+		get => Get(GDExtensionPropertyName.TessellationLevel).As<long>();
+		set => Set(GDExtensionPropertyName.TessellationLevel, value);
+	}
+
 	public new long MeshSize
 	{
 		get => Get(GDExtensionPropertyName.MeshSize).As<long>();
@@ -341,10 +358,28 @@ public partial class Terrain3D : Node3D
 		set => Set(GDExtensionPropertyName.VertexSpacing, value);
 	}
 
-	public new long TessellationLevel
+	public new double CullMargin
 	{
-		get => Get(GDExtensionPropertyName.TessellationLevel).As<long>();
-		set => Set(GDExtensionPropertyName.TessellationLevel, value);
+		get => Get(GDExtensionPropertyName.CullMargin).As<double>();
+		set => Set(GDExtensionPropertyName.CullMargin, value);
+	}
+
+	public new long CastShadows
+	{
+		get => Get(GDExtensionPropertyName.CastShadows).As<long>();
+		set => Set(GDExtensionPropertyName.CastShadows, value);
+	}
+
+	public new long GiMode
+	{
+		get => Get(GDExtensionPropertyName.GiMode).As<long>();
+		set => Set(GDExtensionPropertyName.GiMode, value);
+	}
+
+	public new long RenderLayers
+	{
+		get => Get(GDExtensionPropertyName.RenderLayers).As<long>();
+		set => Set(GDExtensionPropertyName.RenderLayers, value);
 	}
 
 	public new double DisplacementScale
@@ -371,34 +406,76 @@ public partial class Terrain3D : Node3D
 		set => Set(GDExtensionPropertyName.BufferShaderOverride, value);
 	}
 
-	public new long RenderLayers
+	public new bool OceanEnabled
 	{
-		get => Get(GDExtensionPropertyName.RenderLayers).As<long>();
-		set => Set(GDExtensionPropertyName.RenderLayers, value);
+		get => Get(GDExtensionPropertyName.OceanEnabled).As<bool>();
+		set => Set(GDExtensionPropertyName.OceanEnabled, value);
+	}
+
+	public new long OceanMeshLods
+	{
+		get => Get(GDExtensionPropertyName.OceanMeshLods).As<long>();
+		set => Set(GDExtensionPropertyName.OceanMeshLods, value);
+	}
+
+	public new long OceanTessellationLevel
+	{
+		get => Get(GDExtensionPropertyName.OceanTessellationLevel).As<long>();
+		set => Set(GDExtensionPropertyName.OceanTessellationLevel, value);
+	}
+
+	public new long OceanMeshSize
+	{
+		get => Get(GDExtensionPropertyName.OceanMeshSize).As<long>();
+		set => Set(GDExtensionPropertyName.OceanMeshSize, value);
+	}
+
+	public new double OceanVertexSpacing
+	{
+		get => Get(GDExtensionPropertyName.OceanVertexSpacing).As<double>();
+		set => Set(GDExtensionPropertyName.OceanVertexSpacing, value);
+	}
+
+	public new double OceanCullMargin
+	{
+		get => Get(GDExtensionPropertyName.OceanCullMargin).As<double>();
+		set => Set(GDExtensionPropertyName.OceanCullMargin, value);
+	}
+
+	public new long OceanCastShadows
+	{
+		get => Get(GDExtensionPropertyName.OceanCastShadows).As<long>();
+		set => Set(GDExtensionPropertyName.OceanCastShadows, value);
+	}
+
+	public new long OceanGiMode
+	{
+		get => Get(GDExtensionPropertyName.OceanGiMode).As<long>();
+		set => Set(GDExtensionPropertyName.OceanGiMode, value);
+	}
+
+	public new long OceanRenderLayers
+	{
+		get => Get(GDExtensionPropertyName.OceanRenderLayers).As<long>();
+		set => Set(GDExtensionPropertyName.OceanRenderLayers, value);
+	}
+
+	public new Material OceanMaterial
+	{
+		get => Get(GDExtensionPropertyName.OceanMaterial).As<Material>();
+		set => Set(GDExtensionPropertyName.OceanMaterial, value);
+	}
+
+	public new Node3D OceanLightTarget
+	{
+		get => Get(GDExtensionPropertyName.OceanLightTarget).As<Node3D>();
+		set => Set(GDExtensionPropertyName.OceanLightTarget, value);
 	}
 
 	public new long MouseLayer
 	{
 		get => Get(GDExtensionPropertyName.MouseLayer).As<long>();
 		set => Set(GDExtensionPropertyName.MouseLayer, value);
-	}
-
-	public new long CastShadows
-	{
-		get => Get(GDExtensionPropertyName.CastShadows).As<long>();
-		set => Set(GDExtensionPropertyName.CastShadows, value);
-	}
-
-	public new long GiMode
-	{
-		get => Get(GDExtensionPropertyName.GiMode).As<long>();
-		set => Set(GDExtensionPropertyName.GiMode, value);
-	}
-
-	public new double CullMargin
-	{
-		get => Get(GDExtensionPropertyName.CullMargin).As<double>();
-		set => Set(GDExtensionPropertyName.CullMargin, value);
 	}
 
 	public new bool FreeEditorTextures


TokisanGames:
The other PR could have taken this new branch had you just changed the target and wanted to keep the old.

Anyway, I spent some time playing with it and the shader today. I think the functionality and organization are good. I'm not convinced we need tessellation for the ocean, but I suppose it's free to offer the option. We may wish to add vertex spacing.

This shader seems a bit expensive. One view showed cutting the FPS in half. 300fps to 150fps. I have a lot of tweaks for the shader to improve the look a lot, but we need to look at the performance first. Looking up a noise texture might be a lot cheaper than calculating it.

We should test some other popular ocean shaders and make sure they work just by dropping them in, though we may need to add the code snippet for geomorphing the lods that Xtarsia gave. That should be added here.
https://discord.com/channels/691957978680786944/1065519581013229578/1445856250523947070

src/terrain_3d.cpp
```diff
@@ -1338,6 +1429,13 @@ void Terrain3D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "buffer_shader_override_enabled"), "set_buffer_shader_override_enabled", "is_buffer_shader_override_enabled");
 	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "buffer_shader_override", PROPERTY_HINT_RESOURCE_TYPE, "Shader"), "set_buffer_shader_override", "get_buffer_shader_override");
 
+	ADD_GROUP("Ocean Mesh", "");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "ocean_enabled"), "set_ocean_enabled", "is_ocean_enabled");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "ocean_mesh_lods", PROPERTY_HINT_RANGE, "1,10,1"), "set_ocean_mesh_lods", "get_ocean_mesh_lods");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "ocean_mesh_size", PROPERTY_HINT_RANGE, "8,256,2"), "set_ocean_mesh_size", "get_ocean_mesh_size");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "ocean_tessellation_level", PROPERTY_HINT_RANGE, "0,6,1"), "set_ocean_tessellation_level", "get_ocean_tessellation_level");
+	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "ocean_material", PROPERTY_HINT_RESOURCE_TYPE, "Material"), "set_ocean_material", "get_ocean_material");
```
TokisanGames: The hint should be "ShaderMaterial,BaseMaterial3D". We don't need sky, fog, canvas item available in the inspector.

src/terrain_3d_mesher.cpp
```diff
@@ -106,40 +106,40 @@ RID Terrain3DMesher::_instantiate_mesh(const PackedVector3Array &p_vertices, con
...
 	_clear_clipmap();
-	_generate_mesh_types(p_size);
-	_generate_offset_data(p_size);
-	LOG(DEBUG, "Creating instances for all mesh segments for clipmap of size ", p_size, " for ", p_lods, " LODs");
-
-	for (int level = 0; level < p_lods; level++) {
+	_generate_mesh_types();
+	_generate_offset_data();
+	LOG(DEBUG, "Creating instances for all mesh segments for clipmap of size ", _mesh_size, " for ", _lods, " LODs");
+	const RID scenario = _terrain->get_world_3d()->get_scenario();
```
TokisanGames: You need to test that _terrain is valid or use IS_INIT(VOID);, or set the scenario in initialize(), where I think it was before.

TokisanGames:
I added a fix that I'll cherry-pick out. It will go away after a rebase. I also added some updates to the shader and demo that we'll continue to refine or replace.

aidandavey:
OK thanks, I have some more updates based on the review I'm about to push

TokisanGames:
Oh, also we should pass the heightmap into the ocean shader, so that vertex() can discard itself if it is lower than the terrain height minus a margin.

aidandavey:
> Oh, also we should pass the heightmap into the ocean shader, so that vertex() can discard itself if it is lower than the terrain height minus a margin.

Yes, i cant help but think this could also be used for shoreline effects, breaking waves rolling towards the land mass. Etc...

Saul2022:
The shadee works well in the mobile renderer, with some optimizations ( size and lod) it doesnt cut that much fps  ,though there might be some room of improvement there.

I would also add some way to tint the color of water and  shoreline might be good , but cost a lot of fps from the shadwr i tested.

Xtarsia:
I pushed some changes to the shader to get it working (as long as the mesh_size / vertex_spacing etc, are correct)

src/terrain_3d_material.h
```diff
@@ -88,7 +89,7 @@ class Terrain3DMaterial : public Resource {
 	String _strip_comments(const String &p_shader) const;
 	String _inject_editor_code(const String &p_shader) const;
 	void _update_shader();
-	void _update_uniforms(const RID &p_material);
+	void _update_uniforms(const RID &p_material, const Terrain3DMesher &p_mesher);
```
TokisanGames: This should be `const Terrain3DMesher *p_mesher`. Inside the functions can revert to ptr-> instead of *ptr.

src/terrain_3d_material.cpp
```diff
@@ -590,21 +593,28 @@ void Terrain3DMaterial::_update_uniforms(const RID &p_material) {
 	LOG(EXTREME, "Control map RID: ", data->get_control_maps_rid());
 	LOG(EXTREME, "Color map RID: ", data->get_color_maps_rid());
 
-	real_t spacing = _terrain->get_vertex_spacing();
+	RS->material_set_param(p_material, "_background_mode", _world_background);
+
+	real_t spacing = p_mesher->get_vertex_spacing();
```
TokisanGames: This will crash if p_mesher is invalid.
It's also inconsistent with how you get the meshers from _terrain in update().

project/addons/terrain_3d/extras/shaders/ocean_shader.gdshader
```diff
@@ -0,0 +1,134 @@
...
+
+    float h;
+    h = sea_level + get_height(UV, TIME * 0.5) * height_scale;
+    vertex.y = h;
+
+    VERTEX = (VIEW_MATRIX * vec4(vertex, 1.0)).xyz;
+}
+
+void fragment() {
+	// not sure about these normals.
```
TokisanGames: Who is not sure about them? @Xtarsia

project/addons/terrain_3d/extras/shaders/ocean_shader_with_heightmap_reads.gdshader
```diff
@@ -0,0 +1,204 @@
...
+    vertex.y = sea_level + get_height(UV, TIME * 0.5) * height_scale;
+
+	if(vertex.y + 5.0 < get_terrain_height(scale, vertex_lerp, start_pos, end_pos, vertex_distance)){
+		vertex.x = 0. / 0.;
+	}
+    VERTEX = (VIEW_MATRIX * vec4(vertex, 1.0)).xyz;
+}
+
+void fragment() {
+	// not sure about these normals.
```
TokisanGames: @Xtarsia

TokisanGames (changes requested):
Setting _mesh_size in the ocean shader is a bit broken. First the shader should default to 48 which is the C++ default. But that isn't the cause of this issue.

The ocean handles LODs seamlessly in the editor:

Image

However in game, LODs break apart

Image

And this is weird, in game:

Image

It happens with both meshes, it's just if tessellation is enabled it's really small.

Image

In editor it looks fine:

Image

If in game, you continuously print the _mesh_size shader parameter it shows 48.0. Change the shader parameter to 47.9, the LODs seams and the vertices around 0 are fixed. Reset it back to 48 and it remains fixed.

----

Every shader we make is every one we have to maintain.

src/terrain_3d_material.cpp
aidandavey: I see the unsafe pointer usage - but I'm not sure what you mean about an inconsistency. 
update() gets the mesher pointers from terrain and then passes them into this function.

project/addons/terrain_3d/extras/shaders/ocean_shader.gdshader
Xtarsia: yeah the normals were a bit wonky, I updated the shader right before I went to sleep for 30+hrs.

the correct normals:
```glsl
	float h, u, v;
	h = get_height(UV, TIME * 0.5) * height_scale;
	u = get_height(UV + vec2(.01, 0.), TIME * 0.5) * height_scale;
	v = get_height(UV + vec2(0., .01), TIME * 0.5) * height_scale;
	NORMAL = mat3(VIEW_MATRIX) * normalize(vec3(h - u, 0.1, h - v));
```

project/addons/terrain_3d/extras/shaders/ocean_shader_with_heightmap_reads.gdshader
Xtarsia: as above.

Xtarsia:
@TokisanGames @aidandavey , I have some thoughts:
I feel like this would benefit from its own seperate "Terrain3DOceanMaterial" class? currently the ocean mesh feels a bit invasive/convoluted reading through some of the changes to get it working.

Or even, a potentially hot take: this would be better suited as a seperate plugin, giving us Ocean3D, Terrain3D, & Sky3D.

Outside of the clipmap, there isnt much linking terrain3D to the ocean mesh. The culling based on terrain height is slower from some quick testing too, simply due to extra bandwidth incured by reading the heightmaps for each ocean vertex.

Also it seems in this PRs current state, displacement is non-functional, though I didnt look into which change was breaking yet.

Maybe some further discusion needed. This is a pretty big feature in its own right and I feel it shouldnt be rushed without abit more planning & thought.

TokisanGames:
It's not just the ocean mesh that has seam issues in game. I'm sure it's just this refactor missed a step that needs to push some shader parameters at startup. Changing them in the remote debugger fixes the issue.

{53D3D6C4-130C-44AC-A162-664FDA8FFA44}

When switching scenes the ocean mesh appears above the camera in the sky for a second before snapping in place. I think that's  a clue.


I spent some time reworking the shader. It now reflects light more accurately, fades the water's edge, smooths out normals over distance, and adds refraction.

{CC9B1971-E786-4FA2-A3FB-19EB5327047A}

{1243EC6B-6EEF-4B4E-A729-3BC934487513}

{E6ECE30C-D5EA-49C7-B2F5-169E51D4D19D}

mikest:
> Anyway, I spent some time playing with it and the shader today. I think the functionality and organization are good. I'm not convinced we need tessellation for the ocean, but I suppose it's free to offer the option. We may wish to add vertex spacing.

This is something I would definitely make use of. I have an ocean shader which uses decaling to produce water effects like waves and prop wash from boats and it has a relatively high density mesh around the camera location to accomplish that.

Seen here, compared to the density of the Terrain3D mesh, I'm using 1cm resolution vertex spacing (radial, which uh... isn't great) in comparison to 1m with the terrain mesh.

image

mikest:
Did some testing and this looks pretty good! I get that the water shader is a demo, and that quite a bit will want to be done on it in the fullness of time, but I did see one problem that might want to be addressed.

There's some height tearing at the mesh LOD edges that is quite noticeable for certain parameter ranges:
image

I got this with Ocean Shader "Height Scale" > 10 and "UV scale" < 30, and ocean mesh at default settings.

Xtarsia:
Adding depth_draw_always to the render mode flags should fix that.

mikest:
> Adding depth_draw_always to the render mode flags should fix that.

I added the flag to the shader and it appears to change the problem into a different problem :)

I'm not enough of a glsl wizard to be able to know what the root of this problem is, but it does seem like the blending tessellations between the LODs having something off about their normals?

Regardless, this artifact is preferrable to the previous artifact.

image

Xtarsia:
> I added the flag to the shader and it appears to change the problem into a different problem :)

``cull_disabled`` -> ``cull_back`` should fix that, though for an ocean, this may be undesirable if the camera is ever to be under water.

src/terrain_3d_mesher.h
```diff
@@ -60,13 +67,24 @@ class Terrain3DMesher {
...
+
+	void set_material(const RID &p_material) { SET_IF_DIFF(_material, p_material); }
+	RID get_material() const { return _material; }
+	void set_tessellation_level(const int p_level) { SET_IF_DIFF(_tessellation_level, p_level); }
+	int get_tessellation_level() const { return _tessellation_level; }
+	void set_mesh_size(const int p_size) { SET_IF_DIFF(_mesh_size, p_size); }
+	int get_mesh_size() const { return _mesh_size; }
+	void set_lods(const int p_lods) { SET_IF_DIFF(_lods, p_lods); }
+	int get_lods() const { return _lods; }
+	void set_vertex_spacing(const real_t p_spacing) { SET_IF_DIFF(_vertex_spacing, p_spacing); }
```
TokisanGames: Here we can just assign it directly. SET_IF_DIFF(x, y) means `set x = y if x != y else exit`.

We only need to use it if we're going to do more like send signals because it early exits.

DeepCanionStudio:
i copied the current water shader and modified it to add varius features from my own water shader
image

the shader adds a light scattering approximation
and depth based alpha on top of the fresnel alpha that was already in the shader

the only thing that could be a problem currently is that you need to pass the light direction to the shader

here's the shader code:

```
shader_type spatial;
render_mode cull_back, depth_draw_always, diffuse_burley, specular_schlick_ggx, skip_vertex_transform;

// Shader adapted from https://docs.godotengine.org/en/4.4/tutorials/shaders/your_first_shader/your_second_3d_shader.html

//////////////////////////////
// Constants
//////////////////////////////
const int ITERATIONS_VERTEX = 3;
const int ITERATIONS_FRAGMENT = 4;

//////////////////////////////
// Uniforms
//////////////////////////////
group_uniforms water;
uniform float time_scale : hint_range(0.0, 16.0) = 3.0;
uniform float height_scale : hint_range(0.0, 16.0) = 3.0;
uniform float ocean_scale : hint_range(0.1, 64.0, 0.1) = 4.0;
uniform float sea_level : hint_range(-50.0, 50.0, 0.1) = 12.0;
uniform vec3 albedo_color : source_color = vec3(.102, .212, .318);
uniform float visible_depth : hint_range(0., 10000., .1) = 256.;
uniform float depth_curve : hint_range(0., 1.) = .2;
uniform float edge_fade_distance : hint_range(0.0, 10.0, 0.01) = 1.;
uniform float refraction_strength : hint_range(0.0, 0.1, 0.001) = 0.07;

group_uniforms light_scattering;

uniform vec3 light_direction;
uniform vec3 light_color : source_color = vec3(1.0, 1.0, .735);
uniform float height_scattering : hint_range(0.0, 15.0) = 4.0;
uniform float height_threshold : hint_range(0.0, 1.0) = 0.1;
uniform float height_angle_threshold : hint_range(0.0, 1.0) = 0.21;

uniform float fersnel_scattering : hint_range(0.0, 70.0) = 15.0;
uniform float direct_scattering : hint_range(0.0, 70.0) = 10.0;

uniform float depth_absorbtion : hint_range(0.0, 10.0) = 1.0;

group_uniforms;
group_uniforms reflections;
uniform float fresnel_scale : hint_range(0., 1.) = .065;
uniform float fresnel_power = 1.;
uniform float specular : hint_range(0., 1.) = .5;
group_uniforms;

// Uniforms set by C++
group_uniforms private_dont_touch;
uniform float _mesh_size = 32.;
uniform float _subdiv = 1.;
uniform float _vertex_spacing = 1.;
uniform float _vertex_density = 1.;
uniform vec3 _target_pos = vec3(0.,0.,0.);
group_uniforms;

uniform sampler2D depth_texture : hint_depth_texture, repeat_disable, filter_nearest;
uniform sampler2D screen_texture : hint_screen_texture, repeat_disable, filter_nearest;

varying float vertex_distance;

//////////////////////////////
// Ocean Noise
//////////////////////////////

float hash(vec2 p) {
    return fract(sin(dot(p, vec2(127.1, 311.7))) * 43758.5453123);
}

float noise(vec2 p) {
    vec2 i = floor(p);
    vec2 f = fract(p);
    vec2 u = smoothstep(0., 1., f);
    vec2 i1 = i + 1.0;

    float a = hash(i);
    float b = hash(vec2(i1.x, i.y));
    float c = hash(vec2(i.x, i1.y));
    float d = hash(i1);

    return 2.0 * mix(mix(a, b, u.x), mix(c, d, u.x), u.y) - 1.0;
}

float octave(vec2 uv) {
	uv += noise(uv);
	vec2 wave = 1.0 - abs(sin(uv));
	// Shapeing
	float v = wave.x * wave.y + 1e-6;
	float s = v * inversesqrt(v);
	return 1.0 - 2.0 * s + v;
}

float ocean_height(vec2 pos, int iterations) {
	float freq = 0.03 * ocean_scale;
	float amp = 4.0;
	vec2 uv = pos;
	uv.x *= 0.7;
	float h = 0.0;
	float t = TIME * time_scale;

	for (int i = 0; i < iterations; i++) {
	    h += (octave((uv + t) * freq) + octave((uv - t) * freq)) * amp;
	    uv *= mat2(vec2(1.8, 0.8), vec2(-0.8, 1.8));
	    freq *= 1.9;
	    amp *= 0.2;
	}
    return h * 0.1 * height_scale;
}

//////////////////////////////
// Vertex
//////////////////////////////

void vertex() {
    // Get vertex of flat plane in world coordinates and set world UV
    vec3 vertex = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
    float scale = MODEL_MATRIX[0][0];
    float inv_scale = 1.0 / scale;
    float vertex_xz_distance = max(abs(vertex.x - _target_pos.x), abs(vertex.z - _target_pos.z));
    float vertex_lerp = smoothstep(0.0, 1.0, (vertex_xz_distance * inv_scale - _mesh_size - 4.0) / (_mesh_size - 4.0));
    vec2 v_fract = fract(VERTEX.xz * 0.5) * 2.0;
    // For LOD0 morph from a regular grid to an alternating grid to align with LOD1+
    vec2 shift = (scale < _vertex_spacing / _subdiv + 1e-6) ? // LOD0 or not
        // Shift from regular to symmetric
        mix(v_fract, vec2(v_fract.x, -v_fract.y),
            round(fract(round(mod(vertex.z * inv_scale, 4.0)) *
            round(mod(vertex.x * inv_scale, 4.0)) * 0.25))) :
        // symmetric shift
        v_fract * round((fract(vertex.xz * 0.25 * inv_scale) - 0.5) * 4.0);
    vertex.xz -= shift * scale * vertex_lerp;

    UV = (vertex.xz);

    vertex.y = sea_level + ocean_height(UV, ITERATIONS_VERTEX);

    VERTEX = (VIEW_MATRIX * vec4(vertex, 1.0)).xyz;
	vertex_distance = length(vertex - _target_pos);
}

//////////////////////////////
// Fragment
//////////////////////////////

void fragment() {
	float h, u, v;
	// adjust normal domain as screen space distance increases to reduce high frequency values across pixels
	float dv = max(0.25, length(fwidth(VERTEX)));
	h = ocean_height(UV, ITERATIONS_FRAGMENT);
	u = ocean_height(UV + vec2(dv, 0.0), ITERATIONS_FRAGMENT);
	v = ocean_height(UV + vec2(0.0, dv), ITERATIONS_FRAGMENT);
	vec3 world_normal = normalize(vec3(h - u, dv * _vertex_density, h - v));
	NORMAL = mat3(VIEW_MATRIX) * world_normal;
	TANGENT = normalize(cross(NORMAL, VIEW_MATRIX[2].xyz));
	BINORMAL = normalize(cross(NORMAL, TANGENT));

	float fresnel = clamp(fresnel_scale * pow(1.0 + dot(VIEW, NORMAL), fresnel_power), 0., 1.);
	ROUGHNESS = fresnel;
	SPECULAR = specular;

	// Depth fade
	float alpha = 1. - depth_curve + depth_curve * smoothstep(.0, 1., vertex_distance/visible_depth);

	// Edge fade
	float proximity_depth_tex = textureLod(depth_texture, SCREEN_UV, 0.0).r;
	vec4 proximity_view_pos = INV_PROJECTION_MATRIX * vec4(SCREEN_UV * 2.0 - 1.0, proximity_depth_tex, 1.0);
	proximity_view_pos.xyz /= proximity_view_pos.w;
	float edge_mask = 1.0 - clamp(1.0 - smoothstep(proximity_view_pos.z + 2. * edge_fade_distance, proximity_view_pos.z, VERTEX.z), 0.0, 1.0);
	alpha = clamp(alpha - edge_mask, 0., 1.);
	
	// increase alpha with depth as more water means more light absorbed
	alpha = clamp(alpha+ min((VERTEX.z - proximity_view_pos.z) * 0.001 * depth_absorbtion, 1.0), 0.0, 1.0);
	
	// Refraction: Distort screen UVs using the normal (tangent-space projection)
	vec2 refract_offset = NORMAL.xz * refraction_strength;
	// limit refraction at the bottom edge of the screen as normals push away in view direction.
	float y_limit = 1.0 - smoothstep(.0, .25, 1.-SCREEN_UV.y);
	refract_offset = mix(refract_offset, vec2(0.), max(edge_mask, y_limit));
	vec2 refract_uv = SCREEN_UV + refract_offset;
	
	// Invert refract if target screenspace pixel is above ocean surface, replaces major artifact with minor one.
	proximity_depth_tex = textureLod(depth_texture, refract_uv, 0.0).r;
	proximity_view_pos = INV_PROJECTION_MATRIX * vec4(refract_uv * 2.0 - 1.0, proximity_depth_tex, 1.0);
	
	proximity_view_pos.xyz /= proximity_view_pos.w;
	
	refract_uv = SCREEN_UV + mix(refract_offset * (float(proximity_view_pos.z < VERTEX.z) * 2.0 - 1.0), vec2(0.), max(edge_mask, y_limit));
	
	vec3 background = texture(screen_texture, refract_uv).rgb;
	
	// Approximate light scattering 
	vec3 ray_dir = normalize(mat3(INV_VIEW_MATRIX) * VIEW);
	// light scattering based on height and sun angle
	vec3 light_factor = height_scattering * max(h, height_threshold) * max(pow(dot(normalize(light_direction), -ray_dir), 4.0), height_angle_threshold)* light_color * pow(0.5 - 0.5 * dot(normalize(light_direction), world_normal), 3.0);
	// light scattering based on view angle
	light_factor += fersnel_scattering *pow(dot(ray_dir, world_normal), 2.0) * light_color * albedo_color;
	// direct light scattering
	light_factor += direct_scattering * dot(normalize(light_direction), world_normal) * albedo_color* light_color;
	
	ALBEDO = mix(background, albedo_color * light_color, alpha * alpha);
	BACKLIGHT = light_factor * 4.0 * clamp(alpha, 0.4, 1.0);
	ALPHA = clamp(1.0 - edge_mask, 0., 1.);

}
```

src/terrain_3d_mesher.h
```diff
@@ -60,13 +67,24 @@ class Terrain3DMesher {
...
-	void initialize(Terrain3D *p_terrain);
+	void initialize(Terrain3D *p_terrain, const int p_mesh_size, const int p_lods, const int p_tessellation_level, const real_t p_vertex_spacing, const RID &p_material);
 	void destroy();
 
 	void snap();
 	void reset_target_position() { _last_target_position = V2_MAX; }
 	void update();
 	void update_aabbs();
+
+	void set_material(const RID &p_material) { _material = p_material; }
```
TokisanGames: SET_IF_DIFF() was removed from these because it was unnecessary. SET_IF_DIFF(x, y) means `set x = y if x != y else return`. Because it's an early exit on false, we only need it if:
* We're doing other things after the line like sending a signal or updating.
* Assignment is a heavy operation but checking equality is not.

DeepCanionStudio:
added edge and height foam, made a new height function that almost removes tiling
image

i'll paste only the parts of the code that changed

changed some defaults in the uniforms and constants and added foam controls:
```
//////////////////////////////
// Constants
//////////////////////////////
const int ITERATIONS_VERTEX = 1;
const int ITERATIONS_FRAGMENT =3;

//////////////////////////////
// Uniforms
//////////////////////////////
group_uniforms water;
uniform float time_scale : hint_range(0.0, 16.0) = 3.0;
uniform float height_scale : hint_range(0.0, 16.0) = 3.0;
uniform float ocean_scale : hint_range(0.1, 64.0, 0.1) = 4.0;
uniform float sea_level : hint_range(-50.0, 50.0, 0.1) = 12.0;
uniform vec3 albedo_color : source_color = vec3(.102, .212, .318);
uniform float visible_depth : hint_range(0., 10000., .1) = 256.;
uniform float depth_curve : hint_range(0., 1.) = .15;
uniform float edge_fade_distance : hint_range(0.0, 10.0, 0.01) = 2.;
uniform float refraction_strength : hint_range(0.0, 0.1, 0.001) = 0.07;
uniform sampler2D water_noise : filter_linear_mipmap, repeat_enable;

group_uniforms light_scattering;

uniform bool _light_scattering_enabled = true;

uniform vec3 _light_direction;
uniform vec3 _light_color : source_color = vec3(1.0, 1.0, .735);
uniform float height_scattering : hint_range(0.0, 15.0) = 3.0;
uniform float height_threshold : hint_range(0.0, 1.0) = 0.1;
uniform float height_angle_threshold : hint_range(0.0, 1.0) = 0.21;

uniform float fersnel_scattering : hint_range(0.0, 70.0) = 20.0;
uniform float direct_scattering : hint_range(0.0, 70.0) = 10.0;

uniform float depth_absorbtion : hint_range(0.0, 10.0) = 1.0;

group_uniforms;
group_uniforms reflections;
uniform float fresnel_scale : hint_range(0., 1.) = .065;
uniform float fresnel_power = 1.;
uniform float specular : hint_range(0., 1.) = .5;
group_uniforms;

group_uniforms foam;
uniform bool foam_enabled = true;
uniform bool edge_foam_enabled = true;
uniform sampler2D foam_alpha;
uniform float foam_height : hint_range(-2.0, 5.0) = 2.5;
uniform float foam_scale : hint_range(0.1, 5.0) = 1.5;
```



ocean height function:
```
//////////////////////////////
// Ocean Noise
//////////////////////////////

float octave(vec2 uv) {
	float n = texture(water_noise, uv).r;
	return 1.0 - 2.0 * n;
}

vec2 rotateUV(vec2 uv, float rotation) {
	float mid = 0.5;
	return vec2(
		cos(rotation) * (uv.x - mid) + sin(rotation) * (uv.y - mid) + mid,
		cos(rotation) * (uv.y - mid) - sin(rotation) * (uv.x - mid) + mid
	);
}

float get_base_height(vec2 pos, float multiplier, int iterations) {
	float freq = multiplier * ocean_scale;
	float amp = 4.0;
	vec2 uv = pos;
	uv.x *= 0.6;
	float h = 0.0;
	float t = TIME * time_scale / freq * 0.004;

	for (int i = 0; i < iterations; i++) {
	    h += (octave((uv + t) * freq) + octave((uv - t) * freq * 0.89)) * amp;
	   // rotate uv to avoid repetition
	 	uv = rotateUV(uv, 0.2);
		
		// offset smaller waves by the movement of bigger ones
		uv -= h / freq * 0.003;
		
	    freq *= 2.0;
	    amp *= 0.15;
	}
    return h * 0.1 * height_scale;
}
float ocean_height(vec2 pos, int iterations) {
	float giant_wave = get_base_height(pos, 0.00007, iterations );
	float big_wave = get_base_height(pos, 0.0008, iterations);
	float detail_wave = get_base_height(pos, 0.02, iterations);
	return big_wave + detail_wave * 0.02 + giant_wave * 2.0;
}

//////////////////////////////

```
in the vertex shader, i changed the offset to be on all 3 axis instead of only Y for more interesting displacement:

```
    UV = (vertex.xz);
	
	// move on every axis for more complex displacement
    vertex += sea_level * vec3(0.0, 1.0, 0.0) + ocean_height(UV, ITERATIONS_VERTEX) * vec3(0.5, 1.0, 0.5);

    VERTEX = (VIEW_MATRIX * vec4(vertex, 1.0)).xyz;

```
and i changed the last part of the fragment shader to add foam:

```
vec3 final_color = albedo_color * _light_color;
	float foam_factor = 0.0;
	if (foam_enabled) {
		float detail_wave = get_base_height(UV, 0.005 * foam_scale, 2);
		float foam_texture_alpha = texture(foam_alpha, UV * 0.09 * foam_scale).r;
		foam_factor = clamp(h * 0.5 + (detail_wave * 2.0 +  edge_mask * 100.0 * float(edge_foam_enabled)) * foam_texture_alpha * 1.5 - foam_height,0.0, 1.0);
		
		final_color = mix(final_color, vec3(0.8), foam_factor);
	}
	if (_light_scattering_enabled) {
		// Approximate light scattering
		vec3 ray_dir = normalize(mat3(INV_VIEW_MATRIX) * VIEW);
		// light scattering based on height and sun angle
		vec3 light_factor = height_scattering * max(h, height_threshold) * max(pow(dot(normalize(_light_direction), -ray_dir), 4.0), height_angle_threshold) * _light_color * pow(0.5 - 0.5 * dot(normalize(_light_direction), world_normal), 3.0);
		// light scattering based on view angle
		light_factor += fersnel_scattering *pow(dot(ray_dir, world_normal), 2.0) * _light_color * albedo_color;
		// direct light scattering
		light_factor += direct_scattering * dot(normalize(_light_direction), world_normal) * albedo_color * _light_color;
		BACKLIGHT = light_factor * 4.0 * clamp(alpha, 0.4, 1.0) * max(1.0 - foam_factor, 0.5);
	}
	
	vec3 background = texture(screen_texture, refract_uv).rgb;
	ALBEDO = mix(background, final_color, alpha * alpha);
	ALPHA = clamp(1.0 - edge_mask, 0., 1.);
```

JekSun97:
> added edge and height foam

Great job! Looks great and very realistic!

ChampionX1001:
Great progress, this is something I've wanted in Terrain3D and am glad it is taking shape.

https://github.com/TokisanGames/Terrain3D/issues/298#issuecomment-1893384960 
As mentioned in issue linked above, it would be nice to have this clipmap be paintable. Instead of just having a background option as Ocean, drawing lakes and rivers would be nice too. A part of this would be to be able to draw the clipmap at different heights, so you can have lakes at different heights.

River flow would also be nice to be able to designate within terrain3D, as I see water inside your terrains as a part of the whole "terrain" ecosystem.


> We should test some other popular ocean shaders and make sure they work just by dropping them in, though we may need to add the code snippet for geomorphing the lods that Xtarsia gave. That should be added here.

https://github.com/2Retr0/GodotOceanWaves
Here is a popular (but slept-on) shader for oceans. This shader looks a lot more morphic and realistic than most other I see, and it makes a lot of sense for surrounding islands. It uses the Fast-Fourier-Transform wave simulation developed by Jerry Tessendorf, which was also used in the movie _Titanic_, and implements it into GDshader. The FFT can simulate lunar tides and caustic dispersion at rapid speeds despite how high-fidelity it looks, but it still requires a lot of processing. However, the settings can be tweaked to make it very performant in Godot. I suggest we plug this in as well and see how it performs.

This shader lacks edge foam. However, it has not been updated for a year, so I could probably add edge foam to it. Edge foam makes a lot of sense considering this is Terrain3D, where there is going to be land lol

mikest:
Some of the features I am interested in figuring out:
- heightmap for fading lakes to rivers to ocean.
- water current vector field for shader use.
- water feature painting for rapids/waterfalls/shorelinewavebreaks.
- localized decal displacement map for interactive water behaviors.
- compute shader ripple effects (probably not for large bodies of water...)

TokisanGames:
@ChampionX1001 @mikest Both of these comments could be moved and discussed in #298 . 
This PR won't be adding any of those other features. It's just waiting a final review and will be merged in. New additions can come in additional PRs. That ocean shader looks great. Any ocean shader should be able to be dropped in, with the addition of geomorphing vertex code which needs to be isolated and documented.

ChampionX1001:
> This PR won't be adding any of those other features.

Thanks for informing us 

Could you elaborate on whether this PR will support geomorphing vertex code out of the box (via the shader drop-in), because the FFT ocean shader I shared uses vertex displacement.

TokisanGames:
@ChampionX1001 You can see what's in the PR right now. The sample shader geomorphing in vertex(). I expect every ocean shader uses vertex displacement, so they will all need the vertex code adapted to work with the clipmap.

TokisanGames:
Rebased on top of the region blending changes. Fixed crashes. Next I'll update the shader w/ DeepCanion's changes. Xtarsia wanted to review the shader before merging but he's busy. So I intend to merge in a few days and we can look at it again later.

TokisanGames:
I discovered here and fixed an issue in `main` with gaps appearing between lods when changing vertex spacing. The ocean mesh is working fine now and I'm happy with the inspector organization.

Initialize parameters could be moved to a MesherParams struct. All mesher settings could be stored only in Terrain3DMesher and aliased from Terrain3D like the other objects. This seems like the right thing to do, but a lot of other things occur on updates, like set_vertex_spacing() so that will require care and can be done in another PR.

Pending:
* Polish up the foam edge
* Caustics

@Xtarsia I went back to your commits adding caustics, 7876e6f and don't see what I expect to see.

I'm expecting to look down through the surface of the water and see surface light reflected on the ground:



Instead, the only affect I can see that caustics adds are the red streaks on the surface reflections, not on the ground. I changed them to red to highlight, and I had to increase intensity up to 3200 to even see them. Can you look at the implementation again please? 

Godot_v4 5 1-stable_win64_eUxjXV5FnX

TokisanGames:
The ocean waves look like crap under D3D12. DeepCanion has alternate waves, but they are significantly more expensive. So, I want to stick with what we have, and people can replace their ocean shader with any of the many options.

There might be a precision issue in D3D12.

{DCC364CF-D5A5-4EA6-BAFB-81F06F72FC89}

Grok gave me a new hash function in ee6d86f which fixed it and explains:

```
Sin Precision Loss: For moderately large inputs (e.g., UV > 1000–10,000), sin loses precision due to the 
limited mantissa in 32-bit floats (IEEE 754). In GLSL (Vulkan), implementations might handle large 
angles more gracefully or with slight differences in wrapping/reduction, masking the issue. In 
HLSL (D3D12), the reduction can quantize values more noticeably, leading to repeated or zeroed outputs 
that create grid-like patterns (like "under-populated" edges where noise drops to 0).

Integer hashes (using uint multiplies and XOR) are deterministic, avoid transcendental functions, and 
don't suffer from large-input precision loss. This is a common upgrade for procedural noise in shaders 
(inspired by PCG/Wang hashes, adapted for GLSL).
```

TokisanGames:
Ok, I think this is mergeable. I removed caustics for now. Other elements of the shader can be improved in another PR. Anyone can try it out now and give feedback. I'll look once more tomorrow, squash commits, and merge it in.

{1F95EC0F-A182-4C1C-93E6-BA997DA81C8E}

TokisanGames:
Alright good enough for now. Thanks to all the contributors on this one. It's ready for broader testing. The default shader can be improved if you guys want to open up new PRs, or perhaps we'll incorporate a favored ocean shader from someone else later. Refraction can definitely be improved. Caustics can be readded.

You should be able to use your own ocean shaders provided you use `skip_vertex_transform` and copy the geomorphing code from our `vertex()` shader.

PR #888, opened 2025-12-05
Fix instances appear in other scenes - and instance visibility should mirror terrain visibility

aidandavey:
A fix for the bug reported on Discord. RS instances are persistent and need to be destroyed when Terrain3D exits the scene tree. 

Also made a change so that the instances are destroyed and rebuilt when the Terrain3D node visibility changes.


TokisanGames:
Thank you.

PR #898, opened 2025-12-12
Fix region boundary slicing, negative coordinates, and add UI coordinate conversion

### LINKED ISSUE #770: Vertex spacing factored into import positional values
### Terrain3D version

v1.0.1 (Asset store)

### System information

Godot v4.4.1.stable

Not applicable

### Issue description

When vertex spacing is set to a value different from 1.0 i the import values are affected.
For example if set to 10.0 it requires nonsensical position offsets of 5x the image size:

ERROR: core/variant/variant_utility.cpp:1098 - Terrain3DData#9407:import_images:886: (6061, 6061) image will not fit at (-3031.0, 0.0, -3031.0). Try (-30305, -30305) to center

With smaller values it similarly offsets position in possibly undesirable ways.

Reporting because it wasn't entirely obvious to me why imports of the same data with the seemingly the same import settings were producing different results, so I thought it could be helpful.

### Logs

_No response_

rpgshooter:
Fixes #770 

Fixed region boundary slicing bugs in `import_images()` and improved the importer UI to automatically convert logical/pixel coordinates to world coordinates.

 (`terrain_3d_data.cpp`):

1. Negative coordinate handling:
Added `Math::floor()` when converting positions to integers. Previously, casting negative floats to int truncated toward zero (e.g., -0.5 became 0 instead of -1), causing incorrect region placement for negative positions.

2. Region boundary slicing:
Rewrote the slicing logic to properly calculate which regions an image spans based on position + size. Previously, the code only considered image size:
```cpp
int slices_width = ceil(img_size.x / region_size);  // Always 1 for 256px image
```

This meant an image placed at a position crossing region boundaries would only create 1 slice instead of being correctly split across multiple regions.

The new logic calculates actual region overlap and correctly splits images across boundaries.

3. Region index clamping:
Added bounds clamping for region indices to prevent out-of-bounds regions.

(`importer.gd`):

- Import position now accepts logical/pixel coordinates (matching image dimensions)
- UI automatically multiplies by `vertex_spacing` before calling the API
- Added tooltip explaining the coordinate system

This keeps the API consistent (expecting world coordinates) while making the importer UI more intuitive for users.

### DIFF
--- a/project/addons/terrain_3d/tools/importer.gd
+++ b/project/addons/terrain_3d/tools/importer.gd
@@ -38,21 +38,21 @@ func update_heights() -> void:
 
 
 @export_group("Import File")
-## height file - EXR or R16 are recommended for heightmaps. 16-bit PNGs are down sampled to 8-bit and not recommended.
+## EXR or R16 are recommended for heightmaps. 16-bit PNGs are down sampled to 8-bit and not recommended.
 @export_global_file var height_file_name: String = ""
-## control - Only use EXR, and only use files in our proprietary format
+## Only use EXR files in our proprietary format.
 @export_global_file var control_file_name: String = ""
-## color - Any RGB or RGBA format is fine; PNG or Webp are recommended.
+## Any RGB or RGBA format is fine; PNG or Webp are recommended.
 @export_global_file var color_file_name: String = ""
-## Import position - The top left corner (-X, -Y) position of where to place the imported data. Positions are descaled, so ignore the vertex_spacing setting.
+## The top left (-X, -Y) corner position of where to place the imported data. Positions are descaled and ignore the vertex_spacing setting.
 @export var import_position: Vector2i = Vector2i(0, 0) : set = set_import_position
-## Import scale - This scales the height of imported values.
+## This scales the height of imported values.
 @export var import_scale: float = 1.0
-## Height offset - This vertically offsets the height of imported value.
+## This vertically offsets the height of imported values.
 @export var height_offset: float = 0.0
-## r16 range - The lowest and highest height values of the imported image.
+## The lowest and highest height values of the imported image. Only use for r16 files.
 @export var r16_range: Vector2 = Vector2(0, 1)
-## r16 size - The dimensions of the imported image.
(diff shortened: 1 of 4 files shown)


project/addons/terrain_3d/tools/importer.gd
TokisanGames, line 56: Please add a `## tool tip comment` above each of these:
* height file - EXR or R16 are recommended for heightmaps. 16-bit PNGs are down sampled to 8-bit and not recommended.
* control - Only use EXR, and only use files in our proprietary format
* color - Any RGB or RGBA format is fine; PNG or Webp are recommended.
* Import position - The top left corner (-X, -Y) position of where to place the imported data. Positions are descaled, so ignore the vertex_spacing setting.
* Import scale - This scales the height of imported values.
* Height offset - This vertically offsets the height of imported value.
* r16 range - The lowest and highest height values of the imported image.
* r16 size - The dimensions of the imported image.

rpgshooter: > * Please add a ## tool tip comment above each of these:

Updated, with recent commit

TokisanGames: What I meant was write this in order to produce a tool tip.
```
## The dimensions of the imported image.
@export var r16_size: Vector2i = Vector2i(1024, 1024) : set = set_r16_size
```

rpgshooter: oh my bad i misread it as literal will pull a quick update to fix that

rpgshooter: resolving as fixed

TokisanGames:
Just a few issues I resolved, many from before this PR.
* Regions that have been deleted remained deleted after import, so were not visible.
* Regions are not marked modified after changing, so they won't save.
* Deleted regions that were undeleted had old data, but should be cleared.
* Imported Color maps don't display properly because mipmaps aren't being generated, added.
* Update_maps didn't regenerate mipmaps in this import phase because region_locations is empty for some reason

Otherwise it looks good. Nice work. Thank you.

PR #899, opened 2025-12-12
Improve Region Blend, gizmos, shaders

### LINKED ISSUE #100: Draw region grid in shader instead of gizmos

### LINKED ISSUE #757: PBR Channel debug views

### LINKED ISSUE #850: Projection is permanently on
### Terrain3D version

v1.0.1

### System information

Godot v4.5.1.stable - Windows 10 (build 19044) - Multi-window, 1 monitor - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 960 (NVIDIA; 32.0.15.6636) - 13th Gen Intel(R) Core(TM) i5-13400F (16 threads) - 31.83 GiB memory

Yes

### Issue description

Was looking into a shader to optimize it, found out that **enable_projection** is permanently on, it is not tied to an option in **inspector**.
Practically disabling it results in no obvious visual changes.
Except this option eats a whole lot of performance.

With projection enabled inside shader
Image

With projection code disabled, +30 flat fps on my old GPU
Image

### Logs

_No response_

### LINKED ISSUE #858: Bitflags for Material::update()
### Description

Add a bitflags parameter to Material::update() to separate costly updates.

### LINKED ISSUE #902: Terrain texture issue
### Terrain3D version

v1.0.1

### System information

Godot v4.5.stable.mono - Windows 10 (build 19045) - Multi-window, 2 monitors - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 (NVIDIA; 32.0.15.8180) - Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz (20 threads) - 31.85 GiB memory

Yes

### Issue description

To me it looks like there is some flickering or z-fighting going on with my textures. I've tried adjusting the camera near and far values and that hasn't affected it. I've also tried disabling temporal effects. I'm fairly new to game dev so please excuse my ignorance on the subject.

https://github.com/user-attachments/assets/3f2dcf01-971a-4486-88cf-62b2a6667efe

### Logs

[godot.log](https://github.com/user-attachments/files/24146363/godot.log)
[godot2025-12-13T16.42.58.log](https://github.com/user-attachments/files/24146360/godot2025-12-13T16.42.58.log)
[godot2025-12-13T16.51.48.log](https://github.com/user-attachments/files/24146362/godot2025-12-13T16.51.48.log)
[godot2025-12-13T16.51.58.log](https://github.com/user-attachments/files/24146359/godot2025-12-13T16.51.58.log)
[godot2025-12-13T17.08.53.log](https://github.com/user-attachments/files/24146361/godot2025-12-13T17.08.53.log)

Xtarsia:
- [x] Improve flat mode region blending
- [x] perf gain from skipping geomorph lerp at sub-texel scales
- [x] add region map border to region grid, limit grid to inside border, refactor with AA.
- [x] enable region grid when using editor region tool. closes #100
- [x] disable and remove gizmos
- [x] move macro variation to insert system
- [x] move projection to insert system. closes #850
- [x] insert bg mode non search fix
- [x] material update bitflags. closes #858
- [x] make PBR views cumulative. closes #757
- [x] set default bilerp region mip higher. closes #902
- [x] mimic flat mode region blending in colision calculations

### DIFF
--- a/doc/api/class_terrain3dassets.rst
+++ b/doc/api/class_terrain3dassets.rst
@@ -74,8 +74,6 @@ Methods
    +-----------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
    | ``PackedFloat32Array``                                    | :ref:`get_texture_uv_scales<class_Terrain3DAssets_method_get_texture_uv_scales>`\ (\ ) |const|                                                                                  |
    +-----------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-   | ``int``                                                   | :ref:`get_texture_vertical_projections<class_Terrain3DAssets_method_get_texture_vertical_projections>`\ (\ ) |const|                                                            |
-   +-----------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
    | Error                                                     | :ref:`save<class_Terrain3DAssets_method_save>`\ (\ path\: ``String`` = ""\ )                                                                                                    |
    +-----------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
    | |void|                                                    | :ref:`set_mesh_asset<class_Terrain3DAssets_method_set_mesh_asset>`\ (\ id\: ``int``, mesh\: :ref:`Terrain3DMeshAsset<class_Terrain3DMeshAsset>`\ )                              |
@@ -411,18 +409,6 @@ Returns the array of roughness modification values for each texture asset, index
 ``PackedFloat32Array`` **get_texture_uv_scales**\ (\ ) |const| :ref:`🔗<class_Terrain3DAssets_method_get_texture_uv_scales>`
 
 Returns the array of uv scale values for each texture asset, indexed by asset id.
-
-.. rst-class:: classref-item-separator
-
-----
-
-.. _class_Terrain3DAssets_method_get_texture_vertical_projections:
-
-.. rst-class:: classref-method
-
-``int`` **get_texture_vertical_projections**\ (\ ) |const| :ref:`🔗<class_Terrain3DAssets_method_get_texture_vertical_projections>`
-
-Returns a 32-bit int that identifies which textures are to be projected vertically.
 
 .. rst-class:: classref-item-separator
 
--- a/doc/api/class_terrain3dmaterial.rst
+++ b/doc/api/class_terrain3dmaterial.rst
@@ -23,7 +23,9 @@ This class handles options for both the built-in shader and any custom override 
 
 It is a savable resource, so you can save it to disk and use the same material settings in multiple scenes that use Terrain3D. The amount of data is small, assuming you have saved your shader parameter textures to disk, so it can be saved as a git-friendly, text based .tres file or left within the scene file.
 
-While it does mimic some of the functionality of ShaderMaterial, it does not derive from any of the Godot Material classes. It will not pass any ``is Material`` checks. It is a ``Resource``.
+While it does mimic some of the functionality of ShaderMaterial, it does not derive from any of the Godot Material classes. It will fail any ``is Material`` checks. It is a ``Resource``.
+
+Inspector settings above `Custom Shader` and :ref:`shader_override<class_Terrain3DMaterial_property_shader_override>` are used to determine what code is used in the current shader. Inspector settings in `Shader Uniforms` are the public uniforms (not prefaced with `\_`) available in the current shader.
 
 .. rst-class:: classref-reftable-group
 
@@ -47,6 +49,18 @@ Properties
    | ``float``                                                        | :ref:`displacement_sharpness<class_Terrain3DMaterial_property_displacement_sharpness>`                 | ``0.5``   |
    +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
    | ``bool``                                                         | :ref:`dual_scaling_enabled<class_Terrain3DMaterial_property_dual_scaling_enabled>`                     | ``false`` |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`macro_variation_enabled<class_Terrain3DMaterial_property_macro_variation_enabled>`               | ``false`` |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`output_albedo<class_Terrain3DMaterial_property_output_albedo>`                                   | ``true``  |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`output_ambient_occlusion<class_Terrain3DMaterial_property_output_ambient_occlusion>`             | ``true``  |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`output_normal_map<class_Terrain3DMaterial_property_output_normal_map>`                           | ``true``  |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`output_roughness<class_Terrain3DMaterial_property_output_roughness>`                             | ``true``  |
+   +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
+   | ``bool``                                                         | :ref:`projection_enabled<class_Terrain3DMaterial_property_projection_enabled>`                         | ``false`` |
    +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
    | ``Shader``                                                       | :ref:`shader_override<class_Terrain3DMaterial_property_shader_override>`                               |           |
    +------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+-----------+
@@ -124,7 +138,7 @@ Methods
    +-------------+----------------------------------------------------------------------------------------------------------------------------+
    | |void|      | :ref:`set_shader_param<class_Terrain3DMaterial_method_set_shader_param>`\ (\ name\: ``StringName``, value\: ``Variant``\ ) |
    +-------------+----------------------------------------------------------------------------------------------------------------------------+
-   | |void|      | :ref:`update<class_Terrain3DMaterial_method_update>`\ (\ full\: ``bool`` = false\ )                                        |
+   | |void|      | :ref:`update<class_Terrain3DMaterial_method_update>`\ (\ flags\: ``int`` = 0\ )                                            |
    +-------------+----------------------------------------------------------------------------------------------------------------------------+
 
 .. rst-class:: classref-section-separator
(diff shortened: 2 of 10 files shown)


TokisanGames:
Note, the color map is getting injected in #842.

TokisanGames:
We could optionally generate new regions at ground level; and maybe the world noise offset level. Though I'm not terribly concerned about leaving it at 0.

TokisanGames:
> disable or provide option to use gizmos OR shader version

Do we need this? I'd like to drop the classes. The gizmos are ugly. As long as we can render everywhere we need, is there any reason to keep them?

Xtarsia:
some quick AB test in the demo,
Main:
image

This PR:
image


The blendinging differences are due to the grass now being projected at the fixed 45 degree point as well.

TokisanGames:
The speed boost is great. I don't see any, but are there any artifacts due to vertical projection? What happens with faces exactly at 45? Or when a grass slope starts off <45 and curves up >45? Is there a seam?

Xtarsia:
> The speed boost is great. I don't see any, but are there any artifacts due to vertical projection? What happens with faces exactly at 45? Or when a grass slope starts off <45 and curves up >45? Is there a seam?

The condition is checked against each index normal, (not the smooth interpolated terrain normal) and then the biliniear blend prevents seams.

So there are no severe stretching artifacts like before, if the threshold was too high/low.

src/terrain_3d_util.h
```diff
@@ -198,6 +198,47 @@ inline Rect2 aabb2rect(const AABB &p_aabb) {
...
+
+inline Vector3 smoothstep(const real_t p_low, const real_t p_high, const Vector3 &p_value) {
+	Vector3 t = (p_value - Vector3(p_low, p_low, p_low)) / (p_high - p_low);
+	t.x = CLAMP(t.x, 0.f, 1.f);
+	t.y = CLAMP(t.y, 0.f, 1.f);
+	t.z = CLAMP(t.z, 0.f, 1.f);
+	return t * t * (Vector3(3.f, 3.f, 3.f) - 2.f * t);
+}
+
+inline real_t mix(const real_t p_0, const real_t p_1, const real_t p_w) {
```
TokisanGames: Doesn't Math::lerp() and Vector3.lerp() already handle these?

src/shaders/samplers.glsl
```diff
@@ -4,20 +4,10 @@ R"(
...
-uniform highp sampler2DArray _color_maps : source_color, filter_nearest_mipmap_anisotropic, repeat_disable;
-uniform highp sampler2DArray _texture_array_albedo : source_color, filter_nearest_mipmap_anisotropic, repeat_enable;
-uniform highp sampler2DArray _texture_array_normal : hint_normal, filter_nearest_mipmap_anisotropic, repeat_enable;
-
-//INSERT: NOISE_SAMPLER_LINEAR
-uniform highp sampler2D noise_texture : source_color, filter_linear_mipmap_anisotropic, repeat_enable;
-
-//INSERT: NOISE_SAMPLER_NEAREST
-uniform highp sampler2D noise_texture : source_color, filter_nearest_mipmap_anisotropic, repeat_enable;
+#define FILTER_METHOD filter_nearest_mipmap_anisotropic
```
TokisanGames, line 11: Good, simpler.

src/shaders/main.glsl
```diff
@@ -453,7 +411,7 @@ void fragment() {
 	// Calculate the effective mipmap for regionspace, and when less than 0,
 	// skip all extra lookups required for bilinear blend.
 	float region_mip = log2(max(length(base_ddx.xz), length(base_ddy.xz)) * _vertex_density);
-	bool bilerp = region_mip < 0.0 && region_uv.z > -1.;
+	bool bilerp = region_mip < 4.0 && any(greaterThan(ivec4(index[0].z, index[1].z, index[2].z, index[3].z), ivec4(-1)));
```
TokisanGames, line 422: Do you think that 4.0 should be an adjustable parameter since it directly impacts visual quality and performance?

src/terrain_3d_collision.cpp
```diff
@@ -22,7 +22,43 @@
 Dictionary Terrain3DCollision::_get_shape_data(const Vector2i &p_position, const int p_size) {
 	IS_DATA_INIT_MESG("Terrain not initialized", Dictionary());
 	const Terrain3DData *data = _terrain->get_data();
+
+	Ref<Terrain3DMaterial> material = _terrain->get_material();
```
TokisanGames: Need to check that `material.is_valid()` even though it always(TM) is.

src/terrain_3d_collision.cpp
```diff
@@ -22,7 +22,43 @@
...
+
+		float a = check_region(uv2 + Vector2(0.0f, 1.0f));
+		float b = check_region(uv2 + Vector2(1.0f, 1.0f));
+		float c = check_region(uv2 + Vector2(1.0f, 0.0f));
+		float d = check_region(uv2 + Vector2(0.0f, 0.0f));
+
+		Vector2 blend_factor(2.0f + 126.0f * (1.0f - region_blend),	2.0f + 126.0f * (1.0f - region_blend));
+		Vector2 f(uv2.x - Math::floor(uv2.x), uv2.y - Math::floor(uv2.y));
+		Vector2 w(1.f / (1.f + Math::exp(blend_factor.x * Math::log((1.f - f.x) / f.x))),
+				1.f / (1.f + Math::exp(blend_factor.y * Math::log((1.f - f.y) / f.y))));
```
TokisanGames: Is it possible for `f.x`, `f.y`, or `(1.f + Math::exp(blend_factor.y * Math::log((1.f - f.y) / f.y)))` to be 0.0?

src/terrain_3d_collision.cpp
```diff
@@ -22,7 +22,43 @@
 Dictionary Terrain3DCollision::_get_shape_data(const Vector2i &p_position, const int p_size) {
 	IS_DATA_INIT_MESG("Terrain not initialized", Dictionary());
 	const Terrain3DData *data = _terrain->get_data();
+
+	Ref<Terrain3DMaterial> material = _terrain->get_material();
+	bool is_flat_mode = material->get_world_background() == Terrain3DMaterial::WorldBackground::FLAT;
```
TokisanGames: Why make some of these values const and not the others for consistency?

src/terrain_3d_material.h
```diff
@@ -47,8 +54,16 @@ class Terrain3DMaterial : public Resource {
 	// Material Features
 	WorldBackground _world_background = FLAT;
 	TextureFiltering _texture_filtering = LINEAR;
-	bool _dual_scaling = false;
-	bool _auto_shader = false;
+	bool _dual_scaling_enabled = false;
+	bool _auto_shader_enabled = false;
+	bool _macro_variation_enabled = false;
+	bool _projection_enabled = true;
```
TokisanGames: This should default to false like the others, and let the user enable it.

TokisanGames:
Flat mode works great and will do well with an ocean. In `None` mode, I'm unable to sculpt down to any height I want without a lip. 

{15C7C1E6-975F-4AB1-B1D5-03652D5A098C}

Depending on the camera angle the lip comes all the way up to 0.

{FBFFA5D6-9267-4244-AD3A-B0B8EFB3A3C2}

----
In `Noise` mode the region blend is a little odd. 

In flat mode region_blend goes from the region height to ground_level. This provides great control over blending.

image

Here region_blend=0.25 with Noise, offset lower. Region_blend seems fixed at 0 height. It is more possible to get a usable coast line with Noise than it was before, but it still feels a bit inflexible. Is it possible for Noise to work more like Flat, using world_noise_offset.y as ground_level? Maybe with a multiplier (maybe of about 88).

{7AF9BB1D-9F6C-4074-9690-4C66AF1456C6}

Xtarsia:
Yeah, i can look at making noise blend regions similar to flat.

None is problematic, but now that its a seperate insert, I could investigate it a bit more. I suspect having tesselation enabled doesnt work well with the current fix.

src/terrain_3d_util.h
Xtarsia: Yes they are actually the same now that you mention it, they are still handy in the specific cases of replicating glsl easily, but could be removed.

src/shaders/main.glsl
Xtarsia: I added it in an override shader to test, but its so minor, sliding from 0.0, to 16.0 was effectively un-noticable. I had to replicate the specific case where it can show up to even see it doing anything. I think its fine to leave at a fixed value for now.

src/terrain_3d_collision.cpp
Xtarsia: yeah good catch, added clamps.

Xtarsia:
NOISE background is now purely background (fixed blend, noise always has ZERO effect on in-region heights), and has the same features as FLAT mode.

The PBR debug views, I have found useful on occasion, for checking normals, heights etc are packed correctly. They are tucked away at the bottom of the debug views so I think its fine if they stay.

None mode I think could be a seperate issue, as it might need some investigating again.

src/terrain_3d_material.h
```diff
@@ -26,6 +26,13 @@ class Terrain3DMaterial : public Resource {
 		NEAREST,
 	};
 
+enum UpdateFlags {
+		UNIFORMS = 0u,
+		TEXTURE_ARRAYS = 1u,
+		REGION_ARRAYS = 2u,
+		FULL_REBUILD = TEXTURE_ARRAYS | REGION_ARRAYS,
```
TokisanGames: This structure is a little confusing.
* UNIFORMS implies that it's optional, meaning update(TEXTURE_ARRAYS) won't include UNIFORMS, but it does in the activity.
* 0 isn't a bit and should mean no flags are set. If UNIFORMS was optional it should be a bit. NONE or UNIFORMS_ONLY might be more clear. WorldBackground::NONE already exists and we might have an issue with the same name, especially in C#. Maybe I'll just do UNIFORMS_ONLY.
* The values should look like bits, eg, 1<<1 or 0x1, 0x2, 0x4
* Full Rebuild updates both the arrays, AND rebuilds the shader. There's no option to just do the arrays. I also added a new bit for full_rebuild, allowing doing the arrays only.

src/terrain_3d_material.h
```diff
@@ -84,13 +99,14 @@ class Terrain3DMaterial : public Resource {
...
-	String _generate_buffer_shader_code();
+	String _generate_buffer_shader_code() const;
 	String _strip_comments(const String &p_shader) const;
 	String _inject_editor_code(const String &p_shader) const;
 	void _update_shader();
-	void _update_uniforms(const RID &p_material);
+	void _update_uniforms(const RID &p_material, const uint32_t p_update = UpdateFlags::UNIFORMS);
 	void _set_shader_parameters(const Dictionary &p_dict);
 	Dictionary _get_shader_parameters() const { return _shader_params; }
+	static inline bool _has_update_flag(uint32_t value, UpdateFlags flag) {	return (value & static_cast<uint32_t>(flag)) != 0u; }
```
TokisanGames: I don't think we need this. `if (value & flag)` is fine. If we need more type safety we could look at moving to C++20 and `enum class` and these helper functions. But this is only the first flags enum in the class.

src/terrain_3d_collision.cpp
```diff
@@ -22,13 +22,55 @@
...
+
+		return (1.f - blend) * 2.f;
+	};
 
 	int hshape_size = p_size + 1; // Calculate last vertex at end
 	PackedRealArray map_data = PackedRealArray();
 	map_data.resize(hshape_size * hshape_size);
 	real_t min_height = FLT_MAX;
-	real_t max_height = FLT_MIN;
+	real_t max_height = -FLT_MAX;
```
TokisanGames, line 73: FLT_MIN is not negative!? OMG

TokisanGames (commented):
Really great work, as always. I fixed some broken things in the minimum shader, and adjusted categories in the shader uniforms and inspector parameters.

One pending issue is if the Region tool is selected, it shows the grid. If Terrain3D is deselected, the grid goes away. If reselected, it does not show the grid.

Issue #880, opened 2025-12-03
Update Builds including C++20, GCC14


TokisanGames:
### Description

* Move the build settings to C++20 (already setup in SConstruct, works w/ godot-cpp).
* Replace std::unordered_set/map with Godot's Hashmap, LocalVector, Vector, etc.
* Look at using C++20 modules instead of header files.
* Update to GCC14+ in the build scripts (for module support).
* Test building locally w/ clang instead of msvc.
* Use C++20 coroutines w/ cppcoro, libcopp, co_context

rpgshooter:
I can do a draft PR for the dirty work of:

>     * Replace std::unordered_set/map with Godot's Hashmap, LocalVector, Vector, etc.

TokisanGames:
OK, sounds good.

PR #901, opened 2025-12-13
Updated to Godot HashMap, HashSet, and Pair types

rpgshooter:
Replace std::unordered_set/map with Godot containers:

Addition to #880

### Summary

Replace std::unordered_map and std::unordered_set with Godot's HashMap and HashSet in Terrain3DInstancer
Remove custom hash structs that are no longer needed (Godot's HashMapHasherDefault provides built-in hashing for Vector2i, Vector3, and Pair<F, S>)
This change prepares the codebase for C++20 compatibility

### Changes

constants.h
Removed Vector2iHash, PairVector2iIntHash, and Vector3Hash structs
terrain_3d_instancer.h
Replaced <unordered_map> and <unordered_set> includes with Godot templates
Changed type aliases to use HashMap<Vector2i, Pair<RID, RID>> and HashSet<Pair<Vector2i, int>>
terrain_3d_instancer.cpp
Updated API calls: .empty() → .is_empty(), .count() → .has(), .emplace() → .insert()
Changed std::make_pair to Pair<Vector2i, int>
Updated iteration patterns to use KeyValue<K, V> for HashMap and explicit Pair access for HashSet

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -46,32 +46,6 @@ static const Vector3 V3_ZERO{ 0.f, 0.f, 0.f };
 static const Vector3 V3_MAX{ FLT_MAX, FLT_MAX, FLT_MAX };
 static const Vector3 V3_NAN{ NAN, NAN, NAN };
 static const Vector3 V3_UP{ 0.f, 1.f, 0.f };
-//static const Vector3 V3_DOWN{ 0.f, -1.f, 0.f };
-
-struct Vector2iHash {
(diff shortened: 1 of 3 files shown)


rpgshooter:
Addendum, the constants.h build error due to not having a .cpp and potential build order would be resolved with this merge. Example of one of the build errors:
src/constants.h:1:1: note: 'std::hash' is defined in header '<functional>'; this is probably fixable by adding ""include functional""

PR #903, opened 2025-12-14
Update NDK version to r28b once 4.5

rpgshooter:
updated ndk from r23c to r28b, as r28b is the one used is godot-cpp 4.5 on hold until ready for 4.5

### DIFF
--- a/.github/actions/base-deps/action.yml
+++ b/.github/actions/base-deps/action.yml
@@ -23,7 +23,7 @@ runs:
       if: inputs.platform == 'android'
       uses: nttld/setup-ndk@v1
       with:
-        ndk-version: r23c
+        ndk-version: r28b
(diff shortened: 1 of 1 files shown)


TokisanGames:
For now please edit build.yml so it only tries to build android to reduce the CI load and make testing faster. Then when that's working you can revert that change and ensure all platforms are building.

You can also look at the build scripts in the godot-cpp repo in addition to godot, and godot-jolt.

TokisanGames:
This might need to have godot-cpp updated to the 4.5 API before Android NDK r28c will work.

rpgshooter:
Will keep this as a draft for now until there's consideration for 4.5 then this is available.

TokisanGames:
Great, thanks for updating this. Moving main to 4.5 now.

PR #905, opened 2025-12-16
Add update_region() for per-region collision updates

### LINKED ISSUE #891: Update collision by region
### Description

I'm working on a project with terrain modification (explosions that leave craters in the terrain). The dynamic mode isn't appropriate for what I'm doing, but rebuilding the whole map for each modification is slow and wasteful. Could the update function allow for rebuilding only a single region?

rpgshooter:
Fixes #891 
Updates `update()` method to allow for individual update collision calls for a single region instead of rebuilding all regions.

Currently, runtime terrain modification (craters, explosions, deformation) requires calling `update()` which rebuilds collision for every region. This is expensive for large terrains.

Now update() has p_region_loc, for specific regions

- Only applicable in Full mode (FULL_GAME/FULL_EDITOR)

### DIFF
--- a/src/terrain_3d_collision.cpp
+++ b/src/terrain_3d_collision.cpp
@@ -318,7 +318,7 @@ void Terrain3DCollision::build() {
 	update();
 }
 
-void Terrain3DCollision::update(const bool p_rebuild) {
+void Terrain3DCollision::update(const Vector2i &p_region_loc, const bool p_rebuild) {
 	IS_INIT(VOID);
 	if (!_initialized) {
 		return;
@@ -433,6 +433,9 @@ void Terrain3DCollision::update(const bool p_rebuild) {
 		TypedArray<Vector2i> region_locs = _terrain->get_data()->get_region_locations();
 		for (int i = 0; i < region_locs.size(); i++) {
 			Vector2i region_loc = region_locs[i];
+			if (p_region_loc != V2I_MAX && region_loc != p_region_loc) {
+				continue;
+			}
 			Vector2i shape_pos = region_loc * region_size;
 			Dictionary shape_data = _get_shape_data(shape_pos, region_size);
 			if (shape_data.is_empty()) {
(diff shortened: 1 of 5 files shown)


src/terrain_3d_collision.h
```diff
@@ -65,6 +65,7 @@ class Terrain3DCollision : public Object {
 	void build();
 	void reset_target_position() { _last_snapped_pos = V2I_MAX; }
 	void update(const bool p_rebuild = false);
+	void update_region(const Vector2i &p_region_loc);
```
TokisanGames: Like our other update functions, this location selection should be incorporated into update().
https://github.com/TokisanGames/Terrain3D/blob/main/src/terrain_3d_instancer.h#L97
https://github.com/TokisanGames/Terrain3D/blob/main/src/terrain_3d_data.h#L134

Then it should be able to reuse code in the same function.

It should have a location parameter that defaults to V2I_MAX to indicate all regions.

rpgshooter: Your right, similar to the update_maps overload for the visual terrain, I should do an implementation/integration for the update function that already exist. When I get the chance Ill restructure it

rpgshooter: Is this good? Or is there another approach that would provide the best params
```cpp
void update(const bool p_rebuild = false, const Vector2i &p_region_loc = V2I_MAX);
```

NaiJii:
This would be super useful! Hard agree

src/constants.h
```diff
@@ -3,6 +3,8 @@
 #ifndef CONSTANTS_CLASS_H
 #define CONSTANTS_CLASS_H
 
+#include <functional>
```
rpgshooter: Hm, should it stay or should it go, depends. May yeet it as unrelated to pr in general

src/terrain_3d_collision.cpp
```diff
@@ -8,6 +8,7 @@
 #include <godot_cpp/classes/scene_tree.hpp>
 
 #include "constants.h"
+#include "godot_cpp/core/class_db.hpp"
```
TokisanGames: Why was this included?

rpgshooter: Shouldn't have, will be removed

src/constants.h
TokisanGames: It is unrelated, but I don't mind including it if the functions in this file are referenced in functional. If the used elements are elsewhere, the include should go there. What in this file is using functional?

rpgshooter: I was building against nightly for godot-cpp, it was complaining about, to be removed since main targets 4.5

src/terrain_3d_collision.cpp
```diff
@@ -331,6 +331,10 @@ void Terrain3DCollision::update(const bool p_rebuild) {
 	real_t spacing = _terrain->get_vertex_spacing();
 
 	if (is_dynamic_mode()) {
+		if (p_region_loc != V2I_MAX) {
+			LOG(WARN, "Per-region not available in dynamic mode")
+			return;
```
TokisanGames: I don't think we need to warn, or return. I think we can just silently ignore the region parameter and update as normal. Documentation can detail the behavior.

TokisanGames:
Thank you!

PR #906, opened 2025-12-18
Fix tried to access invalid direct_space_state

aidandavey:
In get_raycast_result we need to check if direct_space_state is valid before accessing it. 

Prevents a crash if for some reason direct_space_state is invalid.


TokisanGames:
This should only occur if using threads and not called within `Node._physics_process in the main thread`:
https://github.com/godotengine/godot/issues/90711

Op might have done something with the physics tick or threading.

TokisanGames:
Thanks.

PR #914, opened 2026-01-03
Fixes and features for grass placement shader and material.

mikest:
In using the example grass shader with my project I found a few issues and added some quality of life features for a better out-of-box experience. The bright flashes are due to some edge condition where the grass tips would blow out the dynamic range which would cause bloom flashes with glow lighting enabled. Clamping it down to 6x range removed the flashes.

I also added the ability to place grass over more than one texture. Three seemed reasonable as a default, and it should be easy for some future person to see how to add more/less.

In this PR:
Features:
- Expose grass color as a uniform
- Allow up to three texture IDs to have grass placed on them

Fixes:
- Fix bright flashing at grass tips by clamping color after pow().
- Fix placement of grass so it is aligned with the painted textures.
Before:
image

After:
image

### DIFF
--- a/project/addons/terrain_3d/extras/particle_example/particles.gdshader
+++ b/project/addons/terrain_3d/extras/particle_example/particles.gdshader
@@ -201,7 +201,6 @@ void start() {
 
 	// Apply scale to offset
 	vec3 offset = position_offset * scale;
-	offset -= vec3(0.5,0,0.5);		// Position the grass between the vertices.
 
 	// Apply normal orientation
 	if (align_to_normal) {
@@ -280,6 +279,7 @@ void start() {
 	TRANSFORM[2] *= scale.z;
 
 	// Apply the position
+	offset -= vec3(0.5,0,0.5);		// Position the grass between the vertices.
 	TRANSFORM[3].xyz = pos.xyz + offset;
 
 	// Save Fixed 2 Random values for Reg/Green color randomness


project/addons/terrain_3d/extras/particle_example/particles.gdshader
```diff
@@ -8,6 +8,10 @@
 shader_type particles;
 render_mode disable_velocity, disable_force;
 
+group_uniforms placement;
+// ESTEE: up to three texture ids that the particles can be placed over. Set to -1 to disable
```
TokisanGames: No need to add your name to comments. Git blame will show us who did what. Change them to just a description.

project/addons/terrain_3d/extras/particle_example/particles.gdshader
```diff
@@ -270,7 +279,8 @@ void start() {
 	TRANSFORM[2] *= scale.z;
 
 	// Apply the position
-	TRANSFORM[3].xyz = pos.xyz + offset;
+	vec3 centered = vec3(-0.5,0,-0.5);		// ESTEE: align the placement over the center of the texture tiles.
```
TokisanGames: This line is for this? `Fix placement of grass so it is aligned with the painted textures.`

Please change to:
`offset -= vec3(0.5, 0, 0.5); // Position the grass between the vertices.`

mikest:
Updated with fixes for nits as well as the suggestion for moving the array out of the shader and into the gdscript.

mikest:
Changing this to a draft. I found that on steep slopes the Y-axis position is above the surface and I'd like to figure that out before this lands. Apologies.

image

When offset is adjusted by a half meter the grass will float over the surface on slopes. Probably because the adjustment needs to compensate for the surface normal.

UPDATE:

image

So, I don't fully understand the way the heightmap is sampled, but it looks to me like its grabbing X+1, XZ+1, and Z+1 vertexes and using those to calculate the normal.

image

Thus, the normal for a cluster of instances centered over the heightmap vertex will always be wrong when calculated from the next vertexes in the grid. The offset, when center, will always have heights that are off using this method.

image

I think it might be better to change how the control buffer is sampled such that the filtering for texture can be closer to how the terrain shader is using it, but I don't know how to do that. I think this would be the correct approach and would have accurate placement.

The misalignment in the XZ plane looks worse than the Y misalignment in my opinion, so if you're comfortable with that I can un-draft this PR. I can also remove the offset, come back to it later, and keep the other features as minor niceties

mikest:
...another approach might be to filter out the left/top edge so there is an even border, but that might also be complicated as one would need a way to know the control borders an unpopulated vertex on the top/left.

PR #915, opened 2026-01-03
Add API for querying point over enabled dynamic collision shapes

mikest:
This adds a new method `is_over_collision_shape()` to `Terrain3DCollision.` It takes a point in global space and checks if it is over any of the _enabled_ collision shapes associated with the terrain.

This is useful for enabled/disabling physics objects.

### DIFF
--- a/src/terrain_3d_collision.cpp
+++ b/src/terrain_3d_collision.cpp
@@ -487,13 +487,10 @@ bool Terrain3DCollision::is_on_collision(const Vector3 &p_world_position) const 
     }
 
     Vector2 pos_descaled = Vector2(p_world_position.x, p_world_position.z) / _terrain->get_vertex_spacing();
-	// Determine which shape cell the position falls into using the same math as `update()`:
-	// shape_offset = _shape_size / 2 (meters), shape_pos = snap(pos - shape_offset)
+	// Determine which shape cell the position falls into using the same math as `update()`
 	Vector2i shape_offset = V2I(_shape_size / 2);
 	Vector2i shape_pos = _snap_to_grid(pos_descaled - Vector2(shape_offset));
-	// Grid origin (top-left) was stored as: _last_snapped_pos + _last_grid_offset * _shape_size
 	Vector2i grid_pos = _last_snapped_pos + _last_grid_offset * _shape_size;
-	// grid_loc is an index in cells from the top-left of the grid
 	Vector2i grid_loc = (shape_pos - grid_pos) / _shape_size;
 
     // Early out if clearly outside the grid


doc/doc_classes/Terrain3DCollision.xml
```diff
@@ -1,5 +1,7 @@
-<?xml version="1.0" encoding="UTF-8" ?>
-<class name="Terrain3DCollision" inherits="Object" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<class name="Terrain3DCollision" inherits="Object"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
```
TokisanGames: I think your editor changed the formatting of this document. Please revert and only commit the changes to add the new function text.

src/terrain_3d_collision.h
```diff
@@ -67,6 +70,8 @@ class Terrain3DCollision : public Object {
 	void update(const bool p_rebuild = false);
 	void destroy();
 
+	bool is_over_collision_shape(const Vector3 &p_world_pos) const;
```
TokisanGames: Please rename to `is_on_collision()`

src/terrain_3d_collision.cpp
```diff
@@ -448,6 +466,50 @@ void Terrain3DCollision::destroy() {
...
+// Check if a point in global cooridinates is over one of our collision shapes.
+// The Y coordinate is ignored.
+bool Terrain3DCollision::is_over_collision_shape(const Vector3 &p_world_pos) const {
+	IS_INIT(false);
+	if (!_initialized) {
+		return false;
+	}
+
+	// assume always over a shape if not dynamic mode
+	if (!is_dynamic_mode()) {
```
TokisanGames: This will be wrong if collision is disabled.

src/terrain_3d_collision.h
```diff
@@ -42,6 +42,7 @@ class Terrain3DCollision : public Object {
 	RID _static_body_rid; // Physics Server Static Body
 	StaticBody3D *_static_body = nullptr; // Editor mode StaticBody3D
 	std::vector<CollisionShape3D *> _shapes; // All CollisionShape3Ds
+	std::set<int> _disabled_shapes; // All disabled shape indexes
```
TokisanGames: I'm 99% sure you don't need this set.

doc/doc_classes/Terrain3DCollision.xml
```diff
@@ -52,6 +54,13 @@
 				- If [member mode] is Dynamic, repositions collision shapes around the camera and recalculates ones that moved. Set [code skip-lint]rebuild[/code] to true to recalculate all shapes within [member radius]. This is very fast, and can be updated at 60fps for little cost.
 			</description>
 		</method>
+		<method name="is_over_collision_shape" qualifiers="const">
+			<return type="bool" />
+			<param index="0" name="position" type="Vector3" />
+			<description>
+				Returns true if the given world position is over any of the collision shapes associated with the terrain. Height (Y) is ignored.
```
TokisanGames: `Returns true if the given world position is coincident with any active terrain collision shapes. Y is ignored, so it checks if X and Z are within the boundaries of enabled collision shapes, regardless of height. You can run a separate check to determine if height or a hole exist at this location. This function is intended to help determine if collision is worth checking in this area.`

doc/doc_classes/Terrain3DCollision.xml
mikest: it did, and I missed it. sorry about that.

src/terrain_3d_collision.h
mikest: i chose "over" because it's not a raycast and doesn't report a contact.

i will change.

src/terrain_3d_collision.h
mikest: I couldn't find a way to get the disabled state of a collision shape from the PhysicsServer. Do you know of a way? `body_is_shape_disabled` doesn't exist.

mikest: is no longer necessary with check against cached grid, which brings confidence up to %100 that this is not needed.

mikest:
Updated per your direction towards using cached grid to test for presence over collision grid. Added profiling check using pattern seen elsewhere in file.

Previous method would average around 20-30us per call.
Current implementation averages around 0-1us per call.

So, "a little bit faster."

mikest:
Speed is there. Doing some testing for correctness.

mikest:
Here's a video of 200 RigidBody3Ds being checked every frame and enabled/disabled as the collision shapes move running at 60fps. The video, sadly, does not run at 60fps.

[demo video](https://www.youtube.com/watch?v=6CpKmvLZrbY)

TokisanGames:
I tested this today. I wanted to compare this function vs running a raycast. In our demo I added this to the end of `Player._physics_process()`:

```
	var time: int = Time.get_ticks_usec()
	_ray_query.from = global_position + Vector3(0, 1., 0)
	_ray_query.to = global_position + Vector3(0, -100., 0)
	var results: Dictionary = _space_state.intersect_ray(_ray_query)
	raycast_time = Time.get_ticks_usec() - time
	var rayed: bool = "collider" in results
		
	time = Time.get_ticks_usec()
	var is_on: bool = _terrain.collision.is_on_collision(global_position)
	is_on_collision_time = Time.get_ticks_usec() - time

	print("Raycast hit: %s in %dus" % [ rayed, raycast_time], " is on collision: %s in %dus " % [ is_on, is_on_collision_time ])
	

var raycast_time: int
var is_on_collision_time: int
@onready var _space_state := get_world_3d().direct_space_state
@onready var _terrain: Terrain3D = $"../Terrain3D"
@onready var _ray_query := PhysicsRayQueryParameters3D.create(Vector3.ZERO, Vector3.DOWN, 0xffffffff, [get_rid()])

```

As I moved the player, I received output like this:
```
Raycast hit: true in 12us is on collision: false in 11us 
Raycast hit: true in 12us is on collision: false in 12us 
Raycast hit: true in 12us is on collision: true in 16us 
Raycast hit: true in 12us is on collision: false in 12us 
Raycast hit: true in 12us is on collision: false in 11us 
Raycast hit: true in 12us is on collision: false in 13us 
Raycast hit: true in 13us is on collision: false in 13us 
Raycast hit: true in 11us is on collision: false in 10us 
Raycast hit: true in 13us is on collision: false in 11us 
Raycast hit: true in 12us is on collision: false in 12us 
Raycast hit: true in 15us is on collision: false in 14us 
Raycast hit: true in 14us is on collision: false in 12us 
Raycast hit: true in 16us is on collision: false in 13us 
Raycast hit: true in 14us is on collision: true in 17us 
Raycast hit: true in 16us is on collision: false in 12us 
Raycast hit: true in 13us is on collision: false in 12us 
Raycast hit: true in 14us is on collision: false in 11us 
Raycast hit: true in 14us is on collision: false in 12us 
Raycast hit: true in 16us is on collision: false in 12us 
Raycast hit: true in 13us is on collision: false in 10us 
Raycast hit: true in 12us is on collision: false in 11us 
Raycast hit: true in 13us is on collision: false in 12us 
Raycast hit: true in 14us is on collision: true in 20us 
Raycast hit: true in 13us is on collision: false in 11us 
Raycast hit: true in 13us is on collision: false in 10us 
Raycast hit: true in 13us is on collision: false in 12us 
Raycast hit: true in 12us is on collision: false in 10us 
Raycast hit: true in 14us is on collision: false in 12us 
```

Most of the time is_on_collision reports false, so something isn't quite right in the math. However, even if fixed, they are about the same speed. 

Given these results, what's the use case for this again? You want to disable the physics sim on objects unless they are on top of collision, which makes sense, but why not use a raycast?

mikest:
Hmm. I'm seeing different results. I put a ratio calculation in your print string to account for machine differences.

```
print("Raycast hit: %s in %dus" % [ rayed, raycast_time], " is on collision: %s in %dus " % [ is_on, is_on_collision_time ], " ratio: %f" % [float(raycast_time)/is_on_collision_time])
```

...and the `is_on_collision` is between 1.5 and 3x faster than the ray casting, especially when moving the character faster. So, that's the use case. That, and the "ray" is infinite for the direct query.

This was run against 4.5.1, so it might be an engine difference.

I am also not seeing any misses for is_on_collision, nor have I seen any during testing so I'm not sure where those are coming from. What engine version did you run your test on? I'm running on Windows11/Godot 4.5.1/Vulkan/Forward+

```
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 10us is on collision: true in 4us  ratio: 2.500000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 13us is on collision: true in 5us  ratio: 2.600000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 13us is on collision: true in 7us  ratio: 1.857143
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 17us is on collision: true in 6us  ratio: 2.833333
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 10us is on collision: true in 4us  ratio: 2.500000
Raycast hit: true in 11us is on collision: true in 4us  ratio: 2.750000
Raycast hit: true in 9us is on collision: true in 4us  ratio: 2.250000
Raycast hit: true in 16us is on collision: true in 5us  ratio: 3.200000
Raycast hit: true in 11us is on collision: true in 6us  ratio: 1.833333
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 9us is on collision: true in 5us  ratio: 1.800000
Raycast hit: true in 9us is on collision: true in 4us  ratio: 2.250000
Raycast hit: true in 12us is on collision: true in 5us  ratio: 2.400000
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 10us is on collision: true in 6us  ratio: 1.666667
Raycast hit: true in 11us is on collision: true in 5us  ratio: 2.200000
Raycast hit: true in 12us is on collision: true in 6us  ratio: 2.000000
Raycast hit: true in 10us is on collision: true in 4us  ratio: 2.500000
Raycast hit: true in 28us is on collision: true in 8us  ratio: 3.500000
```

mikest:
Hey, so if you're not feeling this one and "The Juice isn't worth the Squeeze" that's understandable.

1.5-3x isn't likely to break the bank for most use cases and the overhead of keeping this code working over the life of the engine isn't Zero either. I can just use raycast and find other ways to keep large numbers of objects idling out of the collision region.

The original motivation was to bypass the physics engine's raycast machinery, which is famously not that performant compared to some other engines. I enjoyed the exercise of climbing into your library and wrenching on some of the internals so I got something out of it regardless.

mikest:
What's the hash of the git commit you tested against?

TokisanGames:
I looked again. From commit cc98694, in 4.5.1, vulkan, forward. I added the script to player.gd with your ratio. Inexplicably, this time is_on_collision registered properly as true, so I'll say the failures before was a fluke.

On my system, raycasting was consistently 20+% faster.  Then I further limited the raycast collision mask to 0x01, where the terrain is, and cut off another few us:

```
Raycast hit: true in 11us is on collision: true in 17us  ratio: 0.647059
Raycast hit: true in 11us is on collision: true in 15us  ratio: 0.733333
Raycast hit: true in 11us is on collision: true in 13us  ratio: 0.846154
Raycast hit: true in 11us is on collision: true in 17us  ratio: 0.647059
Raycast hit: true in 11us is on collision: true in 16us  ratio: 0.687500
Raycast hit: true in 10us is on collision: true in 17us  ratio: 0.588235
Raycast hit: true in 12us is on collision: true in 15us  ratio: 0.800000
Raycast hit: true in 12us is on collision: true in 18us  ratio: 0.666667
Raycast hit: true in 10us is on collision: true in 14us  ratio: 0.714286
Raycast hit: true in 13us is on collision: true in 17us  ratio: 0.764706
Raycast hit: true in 11us is on collision: true in 17us  ratio: 0.647059
Raycast hit: true in 12us is on collision: true in 17us  ratio: 0.705882
Raycast hit: true in 12us is on collision: true in 16us  ratio: 0.750000
Raycast hit: true in 10us is on collision: true in 16us  ratio: 0.625000
Raycast hit: true in 11us is on collision: true in 15us  ratio: 0.733333
Raycast hit: true in 12us is on collision: true in 18us  ratio: 0.666667
Raycast hit: true in 11us is on collision: true in 15us  ratio: 0.733333
```

Considering just performance: my old test are the same speed. Your test, they differ by 5-20us. My new test they differ by 5-8us but in the other way. 

I understand the idea of your test results showing 2-3x faster relatively, but relative is the keyword. We're comparing microseconds. A whole physics tick is 16.7ms. There are 16,700 microseconds in there. I think we're in the realm of it's too small to measure.

Next, I removed the above test and put this in DemoScene.gd:_ready():

```
	var time: int = 0
	var raycast_time: int = 0
	var is_on_collision_time: int = 0
	var space_state: PhysicsDirectSpaceState3D = $Terrain3D.get_world_3d().direct_space_state
	var ray_query := PhysicsRayQueryParameters3D.create(Vector3.ZERO, Vector3.DOWN, 0x01, [])
	ray_query.from = Vector3(10., 1., 10.)
	ray_query.to = Vector3(10., -10., 10.)

	for i in 10_000_000:
		time = Time.get_ticks_msec()
		space_state.intersect_ray(ray_query)
		raycast_time += Time.get_ticks_msec() - time

		time = Time.get_ticks_msec()
		terrain.collision.is_on_collision(ray_query.from)
		is_on_collision_time += Time.get_ticks_msec() - time

	print("Raycast time: %dms" % [ raycast_time], " is on collision time: %dms " % [ is_on_collision_time ], " ratio: %f" % [float(raycast_time)/is_on_collision_time])	

Output:
Raycast time: 5216ms is on collision time: 3689ms  ratio: 1.413933
```

> Hey, so if you're not feeling this one and "The Juice isn't worth the Squeeze" that's understandable.

With these results, I'm comfortable saying that yes, technically the new function is 40% faster when run under benchmark conditions. 

However, in practice, running even a few hundred per physics tick, the difference is marginal, negligible, and inconsistent. Depending on the user's system, OS scheduling, other activities happening in your game during the physics tick, and gamma rays, raycasts may end up being faster than the function, short term. The new function is only faster over time and many iterations.

Thank you for this process. It was a good exercise for us to learn some things. I didn't know raycasts were so inexpensive or that they could be sped up a hair by limiting collision masks.

Ultimately, I don't think the results are worth the code that we have to maintain.

mikest:
Reasonable. On to the next one.

PR #918, opened 2026-01-16
Fixed uv scale problem in lightweight shader when using the control texture debug view

stan4dbunny:
When using the debug view to look at the control texture, I noticed that the control texture didn't match up to the actual terrain/color map after using the newest version of the lightweight shader. 

I figured that it's because the uv variable was scaled with _vertex_spacing on line 377, and somehow used for the control texture in the debug_view shader.

I also noticed problems with other debug views, such as the PBR views because the shader expects the material struct (mat).

But maybe the shader is not finished yet, and everyone already knows this, but I thought I would say it here just in case!

### DIFF
--- a/project/addons/terrain_3d/extras/shaders/lightweight.gdshader
+++ b/project/addons/terrain_3d/extras/shaders/lightweight.gdshader
@@ -266,12 +266,14 @@ void fragment() {
 	// Recover UVs
 	vec2 uv = UV;
 	vec2 uv2 = UV2;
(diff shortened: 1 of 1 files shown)


TokisanGames:
Thanks for pointing this out. Sorry that this fell through the cracks and didn't get reviewed. Superseded by #954. 
I've checked all of the overlays and debug views between main and the lightweight shader. It looks like they're all consistent.

PR #920, opened 2026-01-26
fix: make region save messages respect debug_level

ogulcancelik:
Changed `LOG(MESG, ...)` to `LOG(INFO, ...)` for the 'Writing region' message.

When saving terrain with many regions (e.g. 256), the console gets flooded with save messages. Since `MESG` bypasses `debug_level`, users can't silence these even with `debug_level = 0`.

This makes save messages respect `debug_level` like other informational logs.

### DIFF
--- a/src/terrain_3d_region.cpp
+++ b/src/terrain_3d_region.cpp
@@ -312,7 +312,7 @@ Error Terrain3DRegion::save(const String &p_path, const bool p_16_bit) {
 		// Set region path and take over the path from any other cached resources,
 		// incuding those in the undo queue
 	}
-	LOG(MESG, "Writing", (p_16_bit) ? " 16-bit" : "", " region ", _location, " to ", get_path());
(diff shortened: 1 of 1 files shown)


TokisanGames:
This is intentional. There are a few functions that print important info to the console always: exporting and saving. A one time dump from importing is not a big deal as it's rare. Most artists and devs work on only a few regions at a time, so only get a few messages. The saved confirmation feedback is important. I would consider a PR for an alternate means of confirmation, but not removing it entirely, or forcing users to enable all info debug messages.

ogulcancelik:
> This is intentional. There are a few functions that print important info to the console always: exporting and saving. A one time dump from importing is not a big deal as it's rare. Most artists and devs work on only a few regions at a time, so only get a few messages. The saved confirmation feedback is important. I would consider a PR for an alternate means of confirmation, but not removing it entirely, or forcing users to enable all info debug messages.

ah i see. okay i could suggest maybe summarizing those into something like ` Saved 256 regions to res://terrain_data/` but it seems that its unnecessary for most use cases. i'm closing it, thanks anways!

PR #952, opened 2026-03-08
update particle grass

Xtarsia:
Updates particle grass to match ground level region blending when background mode is flat or noise.


TokisanGames:
Great that it respects the region blend. However it continues outside of the region boundary which is only good for Flat and odd for None/Noise.

TokisanGames:
Adjusted the texture filter to be a uniform with a positive selection rather than negative filter.

project/addons/terrain_3d/extras/particle_example/process_material.tres
```diff
@@ -32,9 +32,12 @@ shader_parameter/clod_min_threshold = 0.2
 shader_parameter/clod_max_threshold = 0.8
 shader_parameter/patch_min_threshold = 0.025
 shader_parameter/patch_max_threshold = 0.2
+shader_parameter/texture_filter = -1
 shader_parameter/condition_dither_range = 0.15
 shader_parameter/surface_slope_min = 0.87
 shader_parameter/distance_fade_ammount = 0.66
+shader_parameter/ground_level = -618.9519819002201
```
TokisanGames: How does this and region_blend get updated and saved in this file?

I changed ground level in the terrain, saved, and restarted it didn't change in the particle material or file.
Because you're updating with the render server instead of the material, the changes don't get pushed or updated in the inspector, or apparently saved in the file.

TokisanGames:
Are these unnecessary?

// Defined Constants
#define SKIP_PASS 0
#define VERTEX_PASS 1
#define FRAGMENT_PASS 2

---
The items in the gdscript info group and shadow mode should be moved up above the process & particle material. The last items should be:
* process material
* mesh
* mesh material 

I see info is read only, fine, it can go at the very top and marked read only like Terrain3D version so it greys out. Though setting distance by meters is much more intuitive than instance_spacing * cell_width*grid_width or whatever.
Maybe renaming min_draw_distance is in order. It looks more like max draw distance. But if it's read only, maybe calculated_distance or draw distance or calculated_radius.

project/addons/terrain_3d/extras/particle_example/process_material.tres
Xtarsia: it should be synced by _update_process_parameters() interrain_3D_particles.gd

TokisanGames: That probably updates it in the renderer, but not in the resource in the inspector or the file from what I've seen. So I don't know how you got it into the file so it could be pushed in this PR. I couldn't get it to update on mine. And if someone looks in the private folder the settings will appear wrong.

pl4ttenbau:
how does it look, any progress towards merge?

TokisanGames:
All updates will be listed here.

project/addons/terrain_3d/extras/particle_example/process_material.tres
TokisanGames: `_update_process_parameters()` needs to update the material not just the rendering server, otherwise it won't save the values to disk properly, which could present problems loading in addition to git.

```
-RenderingServer.material_set_param(process_rid, "region_blend", terrain.material.region_blend)
+process_material.set_shader_parameter("region_blend", terrain.material.region_blend)
```

TokisanGames:
> Are these unnecessary?
> // Defined Constants
> #define SKIP_PASS 0
> #define VERTEX_PASS 1
> #define FRAGMENT_PASS 2

These were removed from the main shader in #899, but were left in process_material and displacement_buffer.glsl, both removed now.

project/addons/terrain_3d/extras/particle_example/process_material.tres
TokisanGames: On second thought, I'm going to leave it talking to the rendering server, but manually set all of the saved material settings to 0 or 1 neutral values so they don't look wierd.

TokisanGames:
Ok looks good, thanks @Xtarsia.

PR #953, opened 2026-03-09
Improve ocean shader

TokisanGames:
* Provides top level foam intensity, not just edge
* Correctly measures true depth with a max and curve. Obscures both infinity and meshes below the ocean
* Drops depth_absorption from light scattering as above works fine
* Rewrites refraction to reduce artifacts, and stops refracting meshes above the surface of the water
* Renames many of the variables and organize some code
* Disables shadows on the ocean surface - they appear on the ground underneath


JekSun97:
@TokisanGames Wasn't the ocean supposed to be in version 1.0.2 stable? Is something still unfinished? The community would have tested it to identify any issues.

TokisanGames:
No, it's in 1.1. You can use it in nightly builds.

PR #954, opened 2026-03-11
Fix decal position and macro variation when using textures_enabled and _vertex_spacing is not 1

Radivarig:
The code was mutating the uv variable in fragment based on textures_enabled condition, which broke macro variation in the next block and the painting decal which uses that uv after the fragment.

Cilinder is at (0,0,0), _vertex_spacing 0.5
image


TokisanGames:
In main.glsl, UV is scaled by vertex_spacing, which affects macro variation. The macrov noise texture appears fixed to the vertices as they scale up and down.

In lightweight shader, I see that the macro variation is a fixed scale independent of the vertices and the decal is broken.

vertex spacing is applied to UV twice. Once in vertex (as with main.glsl), and again in fragment. Just removing the second `uv *= _vertex_spacing;` fixes the problems with the decal and macro variation scale.

Next, you replace `uv` with `v_vertex.xz`, which keeps the texture scale fixed, regardless of vertex spacing, and consistent with `main`.

Ok looks good. Nice work. Thank you.

PR #957, opened 2026-03-18
Fix get_index_coord in debug_views

imdongye:
When I enabled features like “auto shader” in the debug views, the display turned white and a shader error occurred. Unknown identifier in expression: ‘SKIP_PASS’

I'm not entirely familiar with the history of `get_index_coord`, but I removed the second argument and it now works properly.


TokisanGames:
The history of main.glsl shows https://github.com/TokisanGames/Terrain3D/commit/ee3a03f50034438fd59a7937d59427fbbd2c6afa#diff-ddd98455d1d9f905df13f0b57cc27f6eb11f8ad776c54d742d114cccfbcdab83

Thanks for the fix.

PR #960, opened 2026-03-24
add specific options for anisotropic filtering

Xtarsia:
adds anisotropy to the filter methods, allowing to select linear / nearest without it if desired. (it was on by default)

This can be used even when anisotropy is on as the project default, to disable it just for the terrain for the benefit of mobile performance etc.

image


TokisanGames:
Thank you.

PR #963, opened 2026-03-31
Fix save crashes from Terrain3DMaterial

### LINKED ISSUE #945: [4.6] Crash in _get_property_list() on save
### Terrain3D version

v1.0.1, Nightly Build #2253

### System information

Godot v4.6.1.stable.mono - Devuan GNU/Linux 6 (excalibur) 6 on X11 - X11 display driver, Multi-window, 2 monitors - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1650 SUPER (nvidia) - AMD Ryzen 5 2600 Six-Core Processor (12 threads) - 15.55 GiB memory

Yes

### Issue description

During saving, the engine crashes with a fairly high probability. The issue was not reproduced when running on Windows 10 under the same conditions.
[This pull request](https://github.com/godotengine/godot-cpp/pull/1904) may be related to the crash.

The most common console output at the time of the crash is:
```
Terrain3D#1629:_notification:1060: NOTIFICATION_EDITOR_PRE_SAVE
Terrain3DData#5449:save_directory:304: Saving data files to res://demo/data
Terrain3DRegion#1195:save:302: Region (0, -1) not modified. Skipping res://demo/data/terrain3d_00-01.res
Terrain3DRegion#8217:save:302: Region (0, -2) not modified. Skipping res://demo/data/terrain3d_00-02.res
Terrain3DRegion#0805:save:302: Region (0, 0) not modified. Skipping res://demo/data/terrain3d_00_00.res
Terrain3DMaterial#5003:save:1006: Generating parameter list from shaders
Terrain3DAssets#4520:save:704: Attempting to save external file: res://demo/data/assets.tres
Terrain3DAssets#4520:save:707: File saved successfully: res://demo/data/assets.tres
ERROR: Internal error, property list was not freed by engine!
   at: get_property_list_bind (src/terrain_3d_material.h:14)
```
Additionally, the following messages were each observed only once:
```
malloc(): invalid size (unsorted)
```
and
```
================================================================
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.6.1.stable.mono.official (14d19694e0c88a3f9e82d899a0400f27a24c176e)
Dumping the backtrace. Please include this when reporting the bug on: https://github.com/godotengine/godot/issues
[1] /usr/share/dotnet/shared/Microsoft.NETCore.App/9.0.12/libcoreclr.so(+0x648acc) [0x7efbc1448acc] (??:0)
[2] /usr/share/dotnet/shared/Microsoft.NETCore.App/9.0.12/libcoreclr.so(+0x647ff5) [0x7efbc1447ff5] (??:0)
[3] /lib/x86_64-linux-gnu/libc.so.6(+0x3fdf0) [0x7efc388abdf0] (??:0)
[4] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x4c07be] (??:0)
[5] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x4794309] (??:0)
[6] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x4542e11] (??:0)
[7] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x1d384d5] (??:0)
[8] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x1d39304] (??:0)
[9] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x4363eb5] (??:0)
[10] /home/user/.../Godot_v4.6.1-stable_mono_linux.x86_64() [0x5219da3] (??:0)
[11] /lib/x86_64-linux-gnu/libc.so.6(+0x92b7b) [0x7efc388feb7b] (??:0)
[12] /lib/x86_64-linux-gnu/libc.so.6(+0x1107b8) [0x7efc3897c7b8] (??:0)
-- END OF C++ BACKTRACE --
================================================================
```

### Logs

_No response_

Hrothdel:
Fixes #945

The `_active_params` variable of the `Terrain3DMaterial` class was being modified during `_get_property_list`, which could happen while other threads are trying to read from the array. Because the array was cleared on each `_get_property_list` call and each resize of the array causes the invalidation of its internal pointer, there was a race condition that could lead to segmentation faults, double frees and the like.

The first commit of this PR fixes the crashing in most practical cases, it just makes it so the class variable is only modified when there are actual differences between the new set of parameters and the old one. This doesn't completely fix the race condition, it's still theoretically possible to have a thread read the variable just as it gets swapped and freed, but I'm not sure if this could actually happen with normal usage.

~The second commit is my attempt at completely eliminating the race condition. It uses `SafeNumeric` to make reads and writes to the `_active_params` variable atomic. This might be a bit convoluted, and it might also be a misuse of the `SafeNumeric` class (using it for pointers like this), but I'm not familiar enough with this code base to suggest any other solution. It might best to have the `_get_property_list` method not change any state, and only update `_active_params` when really necessary using some other trigger.~ Removed from PR.

As a side note, I had to add the include for `functional` to the `constants.h` file in order to build this, I'm not sure what's up with that. I can remove it from the PR if needed.

### DIFF
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -1114,7 +1114,7 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
 		}
 	}
 
-	_active_params.clear();
+	TypedArray<StringName> new_active_params;
 	Dictionary grouped_params;
 	StringName current_group = StringName("shader_uniforms.general");
 	grouped_params[current_group] = Array();
@@ -1138,13 +1138,13 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
 				dict["usage"] = name.contains("::") ? PROPERTY_USAGE_SUBGROUP : PROPERTY_USAGE_GROUP;
 			} else {
 				// Filter out duplicate non-groups entries from displacement buffer shader
-				if (_active_params.has(name)) {
+				if (new_active_params.has(name)) {
 					continue;
 				}
 				dict["usage"] = PROPERTY_USAGE_EDITOR;
 			}
 			// Filter out extraneous parameters from the inspector if they are not present in the terrain shader.
-			if (i >= buffer_param && (!_active_params.has(name) && use != PROPERTY_USAGE_GROUP) && !current_group.contains("Displacement")) {
+			if (i >= buffer_param && (!new_active_params.has(name) && use != PROPERTY_USAGE_GROUP) && !current_group.contains("Displacement")) {
 				LOG(INFO, "Displacement buffer has active parameter: ", name, " not present in terrain shader.");
 				continue;
 			}
@@ -1160,7 +1160,7 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
 			grouped_params[current_group] = group;
 
 			// Populate list of public parameters for current shader
-			_active_params.push_back(name);
+			new_active_params.push_back(name);
 
 			// Store this param in a dictionary that is saved in the resource file
 			// Initially set with default value
@@ -1172,6 +1172,7 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
 			}
 		}
 	}
+	_active_params = new_active_params;
 
 	// Populate Godot's property list
 	Array keys = grouped_params.keys();
--- a/src/constants.h
+++ b/src/constants.h
@@ -3,6 +3,8 @@
 #ifndef CONSTANTS_CLASS_H
 #define CONSTANTS_CLASS_H
 
+#include <functional>
+
 // GDExtension uses the godot namespace, custom modules do not.
 #if defined(GDEXTENSION) && !defined(GODOT_MODULE)
 using namespace godot;


TokisanGames:
@Dj1nus please test this build.

TokisanGames:
Thanks for working on this.

It seems that change in Godot 4.6 uses threads for the inspector much more aggressively and none of our function are designed to be thread safe.

The first commit that builds a local variable is fine. This PR could be simplified to just this commit.

For the second commit with the full fix, std::atomic might be a more appropriate class than safenumeric. However, a better approach would be to redesign the function without mutable variables and no atomics at all. So:
* active_params and shader_params updated with _refresh_shader_params()
* update_shader() and set_shader_override() call _refresh_shader_params()
* _get_property_list() processes const read only variables.

This approach should be thread safe and faster. If you want to tackle this, that could go in a separate PR.

Hrothdel:
That makes sense, I removed the second commit from this PR.

About the function redesign, I'm too unfamiliar with the code for this to be trivial for me, so I'll leave that to anyone else that's interested. I might get around to it in a few days if no one does it in the mean time, but I don't really have a lot of time for this at the moment, sorry.

Dj1nus:
This pull request fixed the issue for me. Over 200 saves, I haven't had a single crash.

src/terrain_3d_material.cpp
```diff
@@ -1173,6 +1173,10 @@ void Terrain3DMaterial::_get_property_list(List<PropertyInfo> *p_list) const {
 		}
 	}
 
+	if (_active_params != new_active_params) {
```
TokisanGames: This doesn't do a full dictionary comparison. It compares the pointers. We already know they are different because we just made it. So we can just assign every time, which is faster than doing a full dict compare.

Hrothdel: These were pointers only in the second (removed) commit, now they are plain instances, and I'm pretty sure that operator actually [does](https://github.com/godotengine/godot/blob/4a919adccf8e398aceca75399c539078c54fe97f/core/variant/array.cpp#L133) a recursive check of the array elements.
I was trying to minimize writes to the shared variable as much as possible, but I've just tested a build that always does the assignment and I can't seem to reproduce the crash, so I guess this should be fine.

TokisanGames: Ok, I didn't know that. In that case, [assigning every time](https://github.com/godotengine/godot/blob/4a919adccf8e398aceca75399c539078c54fe97f/core/variant/array.cpp#L210-L215) will definitely be much faster than comparing every element. But I suppose comparing each means the times it does the assignment will be less frequent. 🤔

TokisanGames:
@Hrothdel thanks for figuring out the solution for us.
@Dj1nus thanks for the continued testing.

I've created a new ticket to fully address the race condition if anyone wants to take it on. #964

Issue #416, opened 2024-07-09
Fireflies on terrain


TokisanGames:
### Description

With a light background and dim light, it's possible to see white dots or fireflies from the background through the terrain, even though the meshes are perfectly aligned. It changes based on camera position, angle, and viewport aspect ratio.

This issue was discovered and worked around in Godot 3. I hadn't been able to reproduce it in GD4 until today. It is easily reproducable in the demo by turning off the light and brightening up the ground color in Sky and moving the camera around.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/d43525fe-8e82-496f-bea9-ed4df51027d1)

I was most strongly seeing it along the highlighted areas.

![image](https://github.com/TokisanGames/Terrain3D/assets/632766/4363a189-9800-48d7-9348-2427b234381c)

Fortunately, the same fix works. 
https://github.com/godotengine/godot/issues/35067#issuecomment-1058617216

It's an issue in the godot renderer. The most effective solution is to avoid using the default model space to view space transform in one operation (MODELVIEW_MATRIX) by adding the `skip_vertex_transform` render mode. Then we do our own conversion in two separate steps at the end of vertex().

```glsl
	VERTEX = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
	VERTEX = (VIEW_MATRIX * vec4(VERTEX, 1.0)).xyz;
```

cc: @Xtarsia

TokisanGames:
Fixed in https://github.com/TokisanGames/Terrain3D/commit/1a312cc4ea11d3404f584008869a668ea331e9a6

PR #967, opened 2026-04-15
Fix double precision building

### LINKED ISSUE #977: Can't build with double precision
### Terrain3D version

ecb53d1, v1.0.2, at least those are the ones I tested today, but v1.0.1 and v1.0.0 also had issues

### System information

Windows 10, multiple Godot versions

Not applicable

### Issue description

I've been trying to build it 2 months ago but couldn't get it to work and gave up, as I'm not a C++/low level guy.
I tried it again today, same results, I dug a bit deeper and I managed to build it and run it on https://github.com/TokisanGames/Terrain3D/commit/1b686902193b66b25bd89065a50a0bd4a17824ad.
So I guess it's not me, just that some source code doesn't handle it correctly as it's a bit forgotten.

Also I noticed that main and tagged versions have some differences, and I'm not sure, are those intentional?
Should there be such difference when building from main vs v1.0.2-stable that can be seen in the logs?

### Logs

Both built with scons precision=double and custom_api_file= pointing to extension_api.json that was built with a working Godot 4.6.3 double precision .exe, same way as https://github.com/TokisanGames/Terrain3D/commit/1b686902193b66b25bd89065a50a0bd4a17824ad + --ignore-errors
<br>

Built from main (https://github.com/TokisanGames/Terrain3D/commit/ecb53d142ef37a25e514a554cab2c5691e55f2e3):
<details>

```
src\terrain_3d_mesh_asset.cpp(492): error C2672: 'differs': no matching overloaded function found
D:\Terrain3D\src\terrain_3d_util.h(314): note: could be 'bool differs(const T &,const T &)'
src\terrain_3d_mesh_asset.cpp(492): note: 'bool differs(const T &,const T &)': template parameter 'T' is ambiguous
src\terrain_3d_mesh_asset.cpp(492): note: could be 'T'
        with
        [
            T=real_t
        ]
src\terrain_3d_mesh_asset.cpp(492): note: or       'float'
src\terrain_3d_mesh_asset.cpp(492): note: 'bool differs(const T &,const T &)': could not deduce template argument for 'const T &' from 'T'
        with
        [
            T=real_t
        ]
scons: *** [src\terrain_3d_mesh_asset.windows.template_debug.double.x86_64.obj] Error 2



src\terrain_3d_editor.cpp(782): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(782): note: while trying to match the argument list '(const real_t)'
src\terrain_3d_editor.cpp(782): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(782): note: while trying to match the argument list '(real_t)'
src\terrain_3d_editor.cpp(782): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(782): note: while trying to match the argument list '(real_t)'
src\terrain_3d_editor.cpp(782): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(782): note: while trying to match the argument list '(real_t)'
src\terrain_3d_editor.cpp(782): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(782): note: while trying to match the argument list '(real_t)'
src\terrain_3d_editor.cpp(784): error C2668: 'get_blend': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(245): note: could be 'uint8_t get_blend(const float)'
D:\Terrain3D\src\terrain_3d_util.h(244): note: or       'uint8_t get_blend(const uint32_t)'
src\terrain_3d_editor.cpp(784): note: while trying to match the argument list '(const real_t)'
scons: *** [src\terrain_3d_editor.windows.template_debug.double.x86_64.obj] Error 2



src\terrain_3d_collision.cpp(60): error C2672: 'godot::Math::clamp': no matching overloaded function found
godot-cpp\include\godot_cpp/core/math.hpp(385): note: could be 'T godot::Math::clamp(T,T,T)'
src\terrain_3d_collision.cpp(60): note: 'T godot::Math::clamp(T,T,T)': template parameter 'T' is ambiguous
src\terrain_3d_collision.cpp(60): note: could be 'float'
src\terrain_3d_collision.cpp(60): note: or       'godot::real_t'
src\terrain_3d_collision.cpp(60): note: 'T godot::Math::clamp(T,T,T)': could not deduce template argument for 'T' from 'float'
src\terrain_3d_collision.cpp(61): error C2672: 'godot::Math::clamp': no matching overloaded function found
godot-cpp\include\godot_cpp/core/math.hpp(385): note: could be 'T godot::Math::clamp(T,T,T)'
src\terrain_3d_collision.cpp(61): note: 'T godot::Math::clamp(T,T,T)': template parameter 'T' is ambiguous
src\terrain_3d_collision.cpp(61): note: could be 'float'
src\terrain_3d_collision.cpp(61): note: or       'godot::real_t'
src\terrain_3d_collision.cpp(61): note: 'T godot::Math::clamp(T,T,T)': could not deduce template argument for 'T' from 'float'
scons: *** [src\terrain_3d_collision.windows.template_debug.double.x86_64.obj] Error 2



LINK : fatal error LNK1181: cannot open input file 'src\terrain_3d_collision.windows.template_debug.double.x86_64.obj'
scons: *** [project\addons\terrain_3d\bin\libterrain.windows.debug.x86_64.dll] Error 1181
scons: done building targets.
  ```
</details>

Built from v1.0.2-stable (https://github.com/TokisanGames/Terrain3D/commit/0077405b52e353c5e5dc3a094e7ede49833ba6fe)
With godot-cpp updated to 10.0.0-rc1 as it was outdated and didn't work with doubles.
Same error appears when building official single precision with godot-cpp 10.0.0-rc1.
<details>

```
src\terrain_3d_texture_asset.cpp(84): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_texture_asset.cpp(84): note: while trying to match the argument list '(int32_t)'
src\terrain_3d_texture_asset.cpp(84): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_texture_asset.cpp(84): note: while trying to match the argument list '(int32_t)'
src\terrain_3d_texture_asset.cpp(105): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_texture_asset.cpp(105): note: while trying to match the argument list '(int32_t)'
src\terrain_3d_texture_asset.cpp(105): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_texture_asset.cpp(105): note: while trying to match the argument list '(int32_t)'
scons: *** [src\terrain_3d_texture_asset.windows.template_debug.double.x86_64.obj] Error 2



src\terrain_3d_region.cpp(168): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_region.cpp(168): note: while trying to match the argument list '(int32_t)'
src\terrain_3d_region.cpp(168): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int32_t>(const T)'
        with
        [
            T=int32_t
        ]
src\terrain_3d_region.cpp(168): note: while trying to match the argument list '(int32_t)'
scons: *** [src\terrain_3d_region.windows.template_debug.double.x86_64.obj] Error 2



src\terrain_3d_data.cpp(116): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<int>(const T)'
        with
        [
            T=int
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<int>(const T)'
        with
        [
            T=int
        ]
src\terrain_3d_data.cpp(116): note: while trying to match the argument list '(int)'
scons: *** [src\terrain_3d_data.windows.template_debug.double.x86_64.obj] Error 2



D:\Terrain3D\src\terrain_3d_util.h(115): error C2668: 'is_power_of_2': ambiguous call to overloaded function
D:\Terrain3D\src\terrain_3d_util.h(95): note: could be 'bool is_power_of_2<T>(const T)'
        with
        [
            T=int
        ]
godot-cpp\include\godot_cpp/core/defs.hpp(144): note: or       'bool godot::is_power_of_2<T>(const T)'
        with
        [
            T=int
        ]
D:\Terrain3D\src\terrain_3d_util.h(115): note: while trying to match the argument list '(T)'
        with
        [
            T=int
        ]
D:\Terrain3D\src\terrain_3d_util.h(115): note: the template instantiation context (the oldest one first) is
src\terrain_3d_collision.cpp(245): note: see reference to function template instantiation 'T int_ceil_pow2<int>(T,T)' being compiled
        with
        [
            T=int
        ]
scons: *** [src\terrain_3d_collision.windows.template_debug.double.x86_64.obj] Error 2



LINK : fatal error LNK1181: cannot open input file 'src\terrain_3d_collision.windows.template_debug.double.x86_64.obj'
scons: *** [project\addons\terrain_3d\bin\libterrain.windows.debug.x86_64.dll] Error 1181
scons: done building targets.
```
</details>

### PR DESCRIPTION
**Pull request by jamonholmgren**
Fixes #977

-------- 

I've been working with Godot's [large world coordinates](https://docs.godotengine.org/en/stable/tutorials/physics/large_world_coordinates.html) build to try to address floating point imprecision issues with the large worlds that my game (Gunship Origins) requires. (Previous attempts include [origin shifting](https://github.com/TokisanGames/Terrain3D/compare/main...jamonholmgren:Terrain3D:jamon/origin-shifting), which comes with its own set of tradeoffs.)

When Godot is built with `precision=double`, `real_t` becomes `double` instead of `float`. This causes three categories of compile errors in Terrain3D:

## 1. Control-map pixel-unpacking helpers

The helpers `get_base`, `get_overlay`, `get_blend`, `get_uv_rotation`, `get_uv_scale`, `is_hole`, `is_nav`, `is_auto` only have `float` overloads. Callers pass `double` values from Godot's Image API, causing errors.

I added `double` overloads that cast to `float`.

## 2. `SET_IF_DIFF` errors

The `differs<T>` template errors when the types mismatch.

I switched to use `std::remove_reference_t`  + `static_cast` so both sides have the same type.

## 3. `Math::clamp` in collision code

`Math::clamp` uses `float` literals (`1e-8f`) but the first argument is `real_t` (now `double`), causing errors. 

I now wrap literals in `real_t()`.

Tested against Godot 4.6-stable built with `precision=double` and godot-cpp master. Terrain3D appears to work correctly, but more testing would be appreciated, whether you're using double precision builds or not.

Terrain3D working in double Godot editor build

My C++ skills aren't as good as they could be; feedback welcome.

### DIFF
--- a/src/terrain_3d_collision.cpp
+++ b/src/terrain_3d_collision.cpp
@@ -57,8 +57,8 @@ Dictionary Terrain3DCollision::_get_shape_data(const Vector2i &p_position, const
 
 		real_t blend_factor = 2.0f + 126.0f * (1.0f - region_blend);
 		Vector2 f = Vector2(uv2.x - Math::floor(uv2.x), uv2.y - Math::floor(uv2.y));
-		f.x = Math::clamp(f.x, 1e-8f, 1.f - 1e-8f);
-		f.y = Math::clamp(f.y, 1e-8f, 1.f - 1e-8f);
+		f.x = Math::clamp(f.x, real_t(1e-8), real_t(1.0 - 1e-8));
+		f.y = Math::clamp(f.y, real_t(1e-8), real_t(1.0 - 1e-8));
 		Vector2 w = Vector2(1.f / (1.f + Math::exp(blend_factor * Math::log((1.f - f.x) / f.x))),
 				1.f / (1.f + Math::exp(blend_factor * Math::log((1.f - f.y) / f.y))));
 		real_t blend = Math::lerp(Math::lerp(d, c, w.x), Math::lerp(a, b, w.x), w.y);
--- a/src/terrain_3d_collision.cpp
+++ b/src/terrain_3d_collision.cpp
@@ -57,8 +57,8 @@ Dictionary Terrain3DCollision::_get_shape_data(const Vector2i &p_position, const
 
 		real_t blend_factor = 2.0f + 126.0f * (1.0f - region_blend);
 		Vector2 f = Vector2(uv2.x - Math::floor(uv2.x), uv2.y - Math::floor(uv2.y));
-		f.x = Math::clamp(f.x, 1e-8f, 1.f - 1e-8f);
-		f.y = Math::clamp(f.y, 1e-8f, 1.f - 1e-8f);
+		f.x = Math::clamp(f.x, real_t(1e-8f), real_t(1.0f - 1e-8f));
+		f.y = Math::clamp(f.y, real_t(1e-8f), real_t(1.0f - 1e-8f));
 		Vector2 w = Vector2(1.f / (1.f + Math::exp(blend_factor * Math::log((1.f - f.x) / f.x))),
 				1.f / (1.f + Math::exp(blend_factor * Math::log((1.f - f.y) / f.y))));
 		real_t blend = Math::lerp(Math::lerp(d, c, w.x), Math::lerp(a, b, w.x), w.y);
--- a/src/terrain_3d_util.h
+++ b/src/terrain_3d_util.h
@@ -7,6 +7,7 @@
 #include <godot_cpp/classes/node.hpp>
 #include <godot_cpp/classes/resource_loader.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
+#include <type_traits>
 
 #include "constants.h"
 #include "generated_texture.h"
@@ -231,46 +232,54 @@ inline uint32_t as_uint(const float p_value) { return *(uint32_t *)&p_value; }
 
 inline uint8_t get_base(const uint32_t p_pixel) { return p_pixel >> 27 & 0x1F; }
 inline uint8_t get_base(const float p_pixel) { return get_base(as_uint(p_pixel)); }
+inline uint8_t get_base(const double p_pixel) { return get_base(static_cast<float>(p_pixel)); }
 inline uint32_t enc_base(const uint8_t p_base) { return (p_base & 0x1F) << 27; }
 inline uint32_t gd_get_base(const uint32_t p_pixel) { return get_base(p_pixel); }
 inline uint32_t gd_enc_base(const uint32_t p_base) { return enc_base(p_base); }
 
 inline uint8_t get_overlay(const uint32_t p_pixel) { return p_pixel >> 22 & 0x1F; }
 inline uint8_t get_overlay(const float p_pixel) { return get_overlay(as_uint(p_pixel)); }
+inline uint8_t get_overlay(const double p_pixel) { return get_overlay(static_cast<float>(p_pixel)); }
 inline uint32_t enc_overlay(const uint8_t p_over) { return (p_over & 0x1F) << 22; }
 inline uint32_t gd_get_overlay(const uint32_t p_pixel) { return get_overlay(p_pixel); }
 inline uint32_t gd_enc_overlay(const uint32_t p_over) { return enc_overlay(p_over); }
 
 inline uint8_t get_blend(const uint32_t p_pixel) { return p_pixel >> 14 & 0xFF; }
 inline uint8_t get_blend(const float p_pixel) { return get_blend(as_uint(p_pixel)); }
+inline uint8_t get_blend(const double p_pixel) { return get_blend(static_cast<float>(p_pixel)); }
 inline uint32_t enc_blend(const uint8_t p_blend) { return (p_blend & 0xFF) << 14; }
 inline uint32_t gd_get_blend(const uint32_t p_pixel) { return get_blend(p_pixel); }
 inline uint32_t gd_enc_blend(const uint32_t p_blend) { return enc_blend(p_blend); }
 
 inline uint8_t get_uv_rotation(const uint32_t p_pixel) { return p_pixel >> 10 & 0xF; }
 inline uint8_t get_uv_rotation(const float p_pixel) { return get_uv_rotation(as_uint(p_pixel)); }
+inline uint8_t get_uv_rotation(const double p_pixel) { return get_uv_rotation(static_cast<float>(p_pixel)); }
 inline uint32_t enc_uv_rotation(const uint8_t p_rotation) { return (p_rotation & 0xF) << 10; }
 inline uint32_t gd_get_uv_rotation(const uint32_t p_pixel) { return get_uv_rotation(p_pixel); }
 inline uint32_t gd_enc_uv_rotation(const uint32_t p_rotation) { return enc_uv_rotation(p_rotation); }
 
 inline uint8_t get_uv_scale(const uint32_t p_pixel) { return p_pixel >> 7 & 0x7; }
 inline uint8_t get_uv_scale(const float p_pixel) { return get_uv_scale(as_uint(p_pixel)); }
+inline uint8_t get_uv_scale(const double p_pixel) { return get_uv_scale(static_cast<float>(p_pixel)); }
 inline uint32_t enc_uv_scale(const uint8_t p_scale) { return (p_scale & 0x7) << 7; }
 inline uint32_t gd_get_uv_scale(const uint32_t p_pixel) { return get_uv_scale(p_pixel); }
 inline uint32_t gd_enc_uv_scale(const uint32_t p_scale) { return enc_uv_scale(p_scale); }
 
 inline bool is_hole(const uint32_t p_pixel) { return (p_pixel >> 2 & 0x1) == 1; }
 inline bool is_hole(const float p_pixel) { return is_hole(as_uint(p_pixel)); }
+inline bool is_hole(const double p_pixel) { return is_hole(static_cast<float>(p_pixel)); }
 inline uint32_t enc_hole(const bool p_hole) { return (p_hole & 0x1) << 2; }
 inline bool gd_is_hole(const uint32_t p_pixel) { return is_hole(p_pixel); }
 
 inline bool is_nav(const uint32_t p_pixel) { return (p_pixel >> 1 & 0x1) == 1; }
 inline bool is_nav(const float p_pixel) { return is_nav(as_uint(p_pixel)); }
+inline bool is_nav(const double p_pixel) { return is_nav(static_cast<float>(p_pixel)); }
 inline uint32_t enc_nav(const bool p_nav) { return (p_nav & 0x1) << 1; }
 inline bool gd_is_nav(const uint32_t p_pixel) { return is_nav(p_pixel); }
 
 inline bool is_auto(const uint32_t p_pixel) { return (p_pixel & 0x1) == 1; }
 inline bool is_auto(const float p_pixel) { return is_auto(as_uint(p_pixel)); }
+inline bool is_auto(const double p_pixel) { return is_auto(static_cast<float>(p_pixel)); }
 inline uint32_t enc_auto(const bool p_auto) { return p_auto & 0x1; }
 inline bool gd_is_auto(const uint32_t p_pixel) { return is_auto(p_pixel); }
 
@@ -336,9 +345,10 @@ _FORCE_INLINE_ bool differs(const T &a, const T &b) {
 }
 
 // Sets A if different from B, otherwise returns
+// Cast b to a's type for precision=double builds.
 #define SET_IF_DIFF(a, b) \
-	if (differs(a, b)) {  \
-		a = b;            \
+	if (differs(a, static_cast<std::remove_reference_t<decltype(a)>>(b))) {  \
+		a = static_cast<std::remove_reference_t<decltype(a)>>(b);            \
 	} else {              \
 		return;           \
 	}
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -737,8 +737,8 @@ void Terrain3DEditor::_apply_undo(const Dictionary &p_data) {
 }
 
 // Returns average of height, blend (as real_t(0-255)), or roughness. Overloaded version handles average color
-real_t Terrain3DEditor::_average(const AverageMode p_mode, const Vector3 &p_global_position, const real_t p_base,
-		const real_t p_nan_val, bool p_alt) const {
+float Terrain3DEditor::_average(const AverageMode p_mode, const Vector3 &p_global_position, const float p_base,
+		const float p_nan_val, bool p_alt) const {
 	IS_DATA_INIT(NAN);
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: src/terrain_3d_collision.cpp [hunk 1]
```diff
@@ -57,8 +57,8 @@ Dictionary Terrain3DCollision::_get_shape_data(const Vector2i &p_position, const
 
 		real_t blend_factor = 2.0f + 126.0f * (1.0f - region_blend);
 		Vector2 f = Vector2(uv2.x - Math::floor(uv2.x), uv2.y - Math::floor(uv2.y));
-		f.x = Math::clamp(f.x, 1e-8f, 1.f - 1e-8f);
-		f.y = Math::clamp(f.y, 1e-8f, 1.f - 1e-8f);
+		f.x = Math::clamp(f.x, real_t(1e-8), real_t(1.0 - 1e-8));
+		f.y = Math::clamp(f.y, real_t(1e-8), real_t(1.0 - 1e-8));
```
TokisanGames: `1e-8f` is a float literal that can be converted to a double for free.
`1e-8` is a double that has a potential loss of precision and a cost to convert to float.
All single and double float literals should have `f`.
If a Godot macro or template complains about mixing types during a double precision build, then that literal can be wrapped in real_t() but still needs the `f`.

src/terrain_3d_util.h
```diff
@@ -231,46 +232,54 @@ inline uint32_t as_uint(const float p_value) { return *(uint32_t *)&p_value; }
 
 inline uint8_t get_base(const uint32_t p_pixel) { return p_pixel >> 27 & 0x1F; }
 inline uint8_t get_base(const float p_pixel) { return get_base(as_uint(p_pixel)); }
+inline uint8_t get_base(const double p_pixel) { return get_base(static_cast<float>(p_pixel)); }
```
TokisanGames: When would a pixel be a double?
Image.get_pixel() returns a Color
Color is exactly 4x 32-bit floats, not real_t, never doubles.
https://github.com/godotengine/godot/blob/master/core/math/color.h#L38-L49

Our control map data format is FORMAT_RF: exactly 1x 32-bit float per pixel, never a double. Those 32-bits are only a container for int data, not floats. It should never be converted or treated as a float. This function assumes the source integer data was converted from an invalid float to an invalid double, then converts it back to an invalid float in order to retrieve the integer data out of it. I would be amazed if that worked without data loss, but regardless this is not safe.

If you have a compiling typing error because of this, let's look at that separately, but this is not the solution.

src/terrain_3d_util.h
```diff
@@ -336,9 +345,10 @@ _FORCE_INLINE_ bool differs(const T &a, const T &b) {
 }
 
 // Sets A if different from B, otherwise returns
+// Cast b to a's type for precision=double builds.
 #define SET_IF_DIFF(a, b) \
-	if (differs(a, b)) {  \
-		a = b;            \
+	if (differs(a, static_cast<std::remove_reference_t<decltype(a)>>(b))) {  \
+		a = static_cast<std::remove_reference_t<decltype(a)>>(b);            \
```
TokisanGames: This change allows us to compare mixed types. When was that occurring? In the repo code we only use this macro on setters, where the incoming parameter and the member variable already have the exact same type. So what prompted you to make this change? 

Double precision builds were working before, but occasionally we add code that mistakenly breaks it. Usually that just means wrapping something in real_t. If you ran into compiler errors that prompted you to do this, I'd like to know what they are. 

As explained above sometimes a float isn't actually used as a float and shouldn't be converted. If we have an explicit float compared against a double, or an int compared with a float, this will now pass it when it should have broken the compile so we could fix it.

TokisanGames:
Neither differs nor the utilities should be changed; at least not for this. I have applied the proper fixes. If someone can report a successful build, I'll merge it.


This is one more report added to others that origin shifting is not worth the trouble and double precision is the better fix.

kampe1234:
Looks like it's working. Built and tested.

Although it looks like terrain doesn't really benefit from it. Not sure if it should. Couldn't find any clear information on that.
Textures and geometry flicker the same way as in single precision, for geometry I see that it's using skip_vertex_transform.
I managed to revert fixes in https://github.com/TokisanGames/Terrain3D/issues/416 and geometry got stable, unfortunately fireflies are still there so it's a trade-off I guess.
But texture flickering is much more noticeable than geometry, can it be fixed? Is it possible? Are there any plans for that?

I also noticed that Overlays don't work properly because of https://github.com/godotengine/godot/issues/85711,
I think I can open a PR for that if it's not a problem, but I don't know a lot about shaders.

TokisanGames:
Thanks for testing the build.

> But texture flickering is much more noticeable than geometry, can it be fixed? Is it possible? Are there any plans for that?

Shaders are only 32-bit. What do you think should be done? We don't control the rendering pipeline.

> I also noticed that Overlays don't work properly... I think I can open a PR for that if it's not a problem,

I wasn't aware of that. If you can make a PR, that would be great.

kampe1234:
> Shaders are only 32-bit. What do you think should be done? We don't control the rendering pipeline.

Wouldn't it be possible to handle it like Godot already does normally somehow? Since it works with multiple meshes internally, right? Couldn't textures work the same way that geometry does?

> I wasn't aware of that. If you can make a PR, that would be great.

Will do tomorrow 👍

TokisanGames:
> Wouldn't it be possible to handle it like Godot already does normally somehow? Since it works with multiple meshes internally, right? Couldn't textures work the same way that geometry does?

You mean with double precision emulation? lol. That would require an invasive change that would significantly slow down single precision. This is an exercise best left to a user who needs to support double precision throughout the pipeline. It will be easier to support once we have an RVT on a GPU workflow, as the shader will be much simpler.

Ataraxia-Mechanica:
> You mean with double precision emulation? lol. That would require an invasive change that would significantly slow down single precision. This is an exercise best left to a user who needs to support double precision throughout the pipeline. It will be easier to support once we have an RVT on a GPU workflow, as the shader will be much simpler.

I have not used this addon yet as I'm still learning the basics of godot, but from the looks of [this](https://godotengine.org/article/emulating-double-precision-gpu-render-large-worlds/) article, as long as you don't use ```skip_vertex_transform``` in a double build engine, it automatically handles the transform in a way that increases precision. Just need a way to copy this transform method in the shader for terrain texture uv and the firefly fix, then it would work properly I think.

TokisanGames:
That article describes building double precision emulation for vertex translation only into the renderer. Not every matrix, the MODELVIEW_MATRIX. Not the full matrix, the translation transform only. We use a lot of the built-in matrices, and rotate UVs. We do a lot of work in world space. All of that will be done in 32-bit unless we emulate, which is very slow.

> Just need a way ... then it would work properly I think.

If you can do it, by all means. I will be exceedingly impressed.

Considering emulation post RVT is more achievable.

Ataraxia-Mechanica:
> Considering emulation post RVT is more achievable.

My bad, just checked the shaders to see how many times ```(INV_)VIEW_MATRIX``` is used and how many variables are in global coordinates and it's indeed a ton of work.

jamonholmgren:
Apologies for missing the activity on this. I've been incredibly busy lately.

Thank you for fixing up the PR Cory, and getting this merged. I appreciate your work.

I have not noticed the terrain shaders being an issue, but then again I'm flying at 150 mph and 150 feet above the terrain, so it's probably not very noticeable in that scenario.

PR #983, opened 2026-05-25
[4.6] Make shortcuts configurable in editor settings

elzewyr:
This PR exposes Terrain3D's hotkeys to the Shortcuts tab in the Editor Settings. The keybinds in the toolbar button tooltips update to new values after Terrain3D node is reselected (there is no signal on shortcut change in the editor settings yet). Additionally, although custom keybinds persist between sessions, their entries are not generated automatically, and they had to be re-created on plugin initialization, resulting in there being no 'reset to default' button in those entries.

Screenshot_20260526_002216

### DIFF
--- a/doc/docs/keyboard_shortcuts.md
+++ b/doc/docs/keyboard_shortcuts.md
@@ -1,7 +1,7 @@
 Keyboard Shortcuts
 =================
 
-The following mouse and keyboard shortcuts or hotkeys are available.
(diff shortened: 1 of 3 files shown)


TokisanGames:
Thanks for this. This PR uses a Godot 4.6 feature and we still support 4.5, so care needs to be used before merging. This might need to be gated behind a version check and support both versions for a while.

elzewyr:
Updated to maintain compatibility with 4.5 and earlier

TokisanGames:
Note to self, before merge: 
* Update overlays xml showing hotkey defaults
* Probably drop instancer grid from default keys

PR #990, opened 2026-06-29
Add multiselect for Terrain3DMeshAssets

aidandavey:
This adds the ability to select multiple Mesh Assets and paint them in a single operation. I changed the selected_id int to an Array<int> 

Left clicking a tile in the asset dock will add it to the selection, or deselect it if it is already selected. 

Right clicking a tile will add it to the selection, but will not remove from the selection. 

Density is independent for each MA, so you can paint grass at high density with lower density trees. 

Instance removal is similarly multi-select, it will remove instances of the MAs that are currently selected. 

Search - because we are selectively showing tiles here, we get a mismatch between the tile id and the asset id. To fix it, I have changed to a system whereby the selected_ids array contains the asset_ids, rather than tile ids. This means we can make a selection, perform a search, add a new MA to the selection, clear the search without losing our original selection.

Fixes #739

### DIFF
--- a/project/addons/terrain_3d/src/asset_dock.gd
+++ b/project/addons/terrain_3d/src/asset_dock.gd
@@ -204,7 +204,8 @@ func _on_godot_window_entered() -> void:
 func set_selected_by_asset_id(p_id: int) -> void:
 	search_box.text = ""
 	_on_search_text_changed()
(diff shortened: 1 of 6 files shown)


aidandavey:
Forgot I had a PR already open for this feature, I've pushed into that branch instead. Closing this one.

PR #991, opened 2026-06-29
Asset picker highlights

### LINKED ISSUE #986: Live highlight for asset picking
When picking assets, dynamically highlight the to-be-picked asset so it highlights on the ground.

aidandavey:
Adds a live highlighter for the mesh asset picker tool

Fixes #986


TokisanGames:
Very good, thank you

PR #998, opened 2026-07-12
Add Windows on Arm build config, extension libraries

NJRBailey:
This PR adds Windows on Arm support in Terrain3D. The existing workflows use MinGW, but WoA support for that is still [in progress](https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28802842658/MinGW+GNU+Toolchain). Therefore we use the same solution as Godot Engine itself, and use LLVM-MinGW to perform the build.

Godot Engine WoA workflow files for reference:
[build-containers](https://github.com/godotengine/build-containers/blob/main/Dockerfile.windows#L8)
[godot-build-scripts](https://github.com/godotengine/godot-build-scripts/blob/main/build-windows/build.sh#L46)

- [x] Tested workflow runs successfully
- [x] Tested in Godot 4.5.2
- [x] Tested Release in Godot 4.7
- [x] Tested Debug in Godot 4.7

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        identifier: [linux, windows, macos, ios, android-arm32, android-arm64, web-nothreads]
+        identifier: [linux, windows, windows-arm64, macos, ios, android-arm32, android-arm64, web-nothreads]
         target: [debug, release]
         include:
           - identifier: linux
@@ -33,6 +33,12 @@ jobs:
             name: 🪟 Windows
             runner: ubuntu-22.04
             flags: arch=x86_64
+
+          - identifier: windows-arm64
+            platform: windows
+            name: 🪟 Windows Arm64
+            runner: ubuntu-22.04
+            flags: arch=arm64 use_mingw=yes use_llvm=yes
 
           - identifier: macos
             platform: macos
@@ -87,6 +93,17 @@ jobs:
           cache-name: ${{ matrix.identifier }}-${{ matrix.target }}
         continue-on-error: true
 
+      - name: Setup LLVM-MinGW (Windows Arm64)
+        if: ${{ matrix.identifier == 'windows-arm64' }}
+        shell: sh
+        run: |
+          export LLVM_MINGW_VERSION=20251118
+          export LLVM_MINGW_NAME=llvm-mingw-${LLVM_MINGW_VERSION}-ucrt-ubuntu-22.04-x86_64
+          curl -LO https://github.com/mstorsjo/llvm-mingw/releases/download/${LLVM_MINGW_VERSION}/${LLVM_MINGW_NAME}.tar.xz
+          tar xf ${LLVM_MINGW_NAME}.tar.xz
+          rm -f ${LLVM_MINGW_NAME}.tar.xz
+          echo "$(pwd)/${LLVM_MINGW_NAME}/bin" >> $GITHUB_PATH
+
       - name: Build Terrain3D
         env:
             SCONS_CACHE: "${{ github.workspace }}/.scons-cache/"
@@ -95,12 +112,16 @@ jobs:
         run: |
             scons target=$TARGET platform='${{ matrix.platform }}' ${{ matrix.flags }} debug_symbols=no -j2
 
-      - name: Strip Libraries (Windows/Linux)
+      - name: Strip Libraries (Windows / Linux)
         if: ${{ matrix.platform == 'windows' || matrix.platform == 'linux' }}
         shell: sh
         run: |
+          STRIP="strip"
+          if [ "${{ matrix.identifier }}" = "windows-arm64" ]; then
+            STRIP="llvm-strip"
+          fi
           ls -l project/addons/terrain_3d/bin/
-          strip project/addons/terrain_3d/bin/libterrain.*
+          $STRIP project/addons/terrain_3d/bin/libterrain.*
           ls -l project/addons/terrain_3d/bin/
 
       - name: Prepare Files
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         platform: [windows]
         target: [debug, release]
-        arch: [x86_64]
+        arch: [x86_64, arm64]
 
     steps:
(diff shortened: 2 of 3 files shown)


TokisanGames:
I ran the workflow but arm is broken.

NJRBailey:
> I ran the workflow but arm is broken.

Thanks so much 🙏 I missed the identifier list on line 22, have updated now if you could please re-run it?

NJRBailey:
I downloaded the [artifact](https://github.com/TokisanGames/Terrain3D/actions/runs/29187778812/artifacts/8258671025) from the [merge step of the CI](https://github.com/TokisanGames/Terrain3D/actions/runs/29187778812/job/86638058351?pr=998#step:2:113) and put it in a new project on Godot 4.5.2, and the demo scene ran fine without any errors 🙌 Checked the `bin` folder and the DLLs are there for both arm and x64. I've attached the Godot console logs for my test run. Is there anything else you'd like me to check?

[Godot Terrain3D WoA 4.5.2 Logs CI Build.txt](https://github.com/user-attachments/files/29937866/Godot.Terrain3D.WoA.4.5.2.Logs.CI.Build.txt)

TokisanGames:
Please test both debug and release builds. Also test Godot 4.7. You can squash all commits.

NJRBailey:
Okay, can I just check that running it in-editor is testing the Debug build, and running an exported .exe tests the Release build? I've run it both ways using Forward+:

## Release .exe
No errors or anything in logs when using the exported `.console.exe` file, all seems to work fine and Windows recognises it as arm64:
image

## In-editor run (Debug)
No errors or anything in logs when running the demo scene:
image

.github/workflows/build.yml
```diff
@@ -87,6 +93,17 @@ jobs:
           cache-name: ${{ matrix.identifier }}-${{ matrix.target }}
         continue-on-error: true
 
+      - name: Setup LLVM-MinGW (Windows Arm64)
+        if: ${{ matrix.identifier == 'windows-arm64' }}
+        shell: sh
+        run: |
+          export LLVM_MINGW_VERSION=20251118
```
TokisanGames, line 100: Why this version instead of anything newer?

.github/workflows/build.yml
```diff
@@ -95,14 +112,22 @@ jobs:
...
-        if: ${{ matrix.platform == 'windows' || matrix.platform == 'linux' }}
+      - name: Strip Libraries (Windows x86_64 / Linux)
+        if: ${{ matrix.identifier == 'windows' || matrix.platform == 'linux' }}
         shell: sh
         run: |
           ls -l project/addons/terrain_3d/bin/
           strip project/addons/terrain_3d/bin/libterrain.*
           ls -l project/addons/terrain_3d/bin/
 
+      - name: Strip Libraries (Windows Arm64)
```
TokisanGames: Lets combine these steps like you did in the windows build.

.github/workflows/build.yml
```diff
@@ -87,6 +93,17 @@ jobs:
...
+      - name: Setup LLVM-MinGW (Windows Arm64)
+        if: ${{ matrix.identifier == 'windows-arm64' }}
+        shell: sh
+        run: |
+          export LLVM_MINGW_VERSION=20251118
+          export LLVM_MINGW_NAME=llvm-mingw-${LLVM_MINGW_VERSION}-ucrt-ubuntu-22.04-x86_64
+          curl -LO https://github.com/mstorsjo/llvm-mingw/releases/download/${LLVM_MINGW_VERSION}/${LLVM_MINGW_NAME}.tar.xz
+          tar xf ${LLVM_MINGW_NAME}.tar.xz
+          rm -f ${LLVM_MINGW_NAME}.tar.xz
+          echo "$(pwd)/${LLVM_MINGW_NAME}/bin" >> $GITHUB_PATH
```
TokisanGames, line 105: Why are you writing to this file?

TokisanGames:
Editor and export w/ debug use the debug build. An export w/o debug uses the release build.

.github/workflows/build.yml
NJRBailey: It's the version Godot uses [here](https://github.com/godotengine/build-containers/blob/main/Dockerfile.windows#L6) but can try the latest version if you'd prefer (looks like [20260616](https://github.com/mstorsjo/llvm-mingw/releases/))?

.github/workflows/build.yml
NJRBailey: Okay sure, will update now and do another squash

.github/workflows/build.yml
NJRBailey: This step installs LLVM-MinGW, but it's used later in the `Build Terrain3D` and `Strip Libraries (Windows Arm64)` steps, so we need to add it to the $GITHUB_PATH variable to make it available in the subsequent steps ([documentation here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-system-path))

NJRBailey:
> Editor and export w/ debug use the debug build. An export w/o debug uses the release build.

Okay I'll re-check without the debug export

NJRBailey:
> Editor and export w/ debug use the debug build. An export w/o debug uses the release build.

Yeah have double checked, I can confirm I exported without debug and no errors in the console.

image

.github/workflows/build.yml
TokisanGames: No, the justification is fine.

.github/workflows/build.yml
TokisanGames: Weird syntax. Ok.

TokisanGames:
Thank you.

TokisanGames:
@NJRBailey Any reason why you chose v20251118, which wasn't even the latest in 21.x? I'm upgrading to the latest stable 20260616 and 22.1.8 in #1023.

NJRBailey:
> @NJRBailey Any reason why you chose v20251118, which wasn't even the latest in 21.x? I'm upgrading to the latest stable 20260616 and 22.1.8 in #1023.

Yep, it was the version which matched Godot Engine's CI - you mentioned it at the time but we decided to use 20251118 (for clarity it looks like Godot have updated their version since then) https://github.com/TokisanGames/Terrain3D/pull/998#discussion_r3566418892

TokisanGames:
Got it, thank you.

Issue #491, opened 2024-09-15
Implement Region Streaming


TokisanGames:
### Description

* Load and unload regions as they enter or exit camera visible bounds.
* Change region map into a fixed size array with loaded & visible regions, which it already is, but make it a region map relative to the camera, not relative to 0, 0, 0.

Empt-y:
## Region streaming (#1020) + a scoping note on the rest of this issue

#1020 implements the first bullet of this issue for real: `Terrain3DData::load_directory()` currently loads every region file in a directory unconditionally, and there was no way to load/unload by distance at all. `update_streaming(target_positions, radius)` fixes that — regions load in and (after saving, if modified) unload as they cross `radius` of any tracked target, with the union computed across *all* targets before anything is evicted, so a region a second player/camera still needs never gets pulled out from under them.

**What that PR deliberately does not attempt**, and why — surfacing this now rather than after starting, since it affects how the rest of the issue should be scoped:

### `get_region_id()` is the GPU shader array, not just similar to it

The "region map relative to camera, not (0,0,0)" and "possible loadable bounds unlimited" asks both run into the same wall: `get_region_id()` — the function every collision, instancer, editor brush, and gameplay `get_height()`/`get_pixel()` call resolves through — reads directly from `_region_map`, and `_region_map` is the *literal array uploaded to the shader*:

```cpp
inline int Terrain3DData::get_region_id(const Vector2i &p_region_loc) const {
    int map_index = get_region_map_index(p_region_loc);
    if (map_index >= 0) {
        int region_id = _region_map[map_index] - 1;
        ...
```

and in `main.glsl` / `displacement_buffer.glsl`:
```glsl
uniform int _region_map[1024];
```

Both a fixed-size, compile-time GLSL array declaration. So "adjustable window size" isn't a C++-side resize — it's a shader change (a larger fixed array with an "active size" uniform, or a different GPU-side structure like a texture-based lookup). And "unbounded loadable bounds" can't just mean widening this one array, because it's simultaneously the answer to "is there a region here" for every non-rendering consumer too. Decoupling those — an unbounded CPU-side lookup for `get_region_id()`, a separate fixed-size windowed array rebuilt for the GPU each time the render camera moves — is real, careful work on the single hottest path in the codebase, and IMO deserves a design decision from a maintainer before a contributor attempts it, given how much depends on it.

### "Relative to the camera" and "multiple players" are in tension as written

[...]

PR #1004, opened 2026-07-17
Add runtime region streaming

Niekvdm:
Partially addresses #491.

This PR adds optional runtime region streaming. Regions around the clipmap target are loaded on background threads and evicted as the target moves away, so memory scales with the loaded area instead of the whole terrain. Our test world has 700 regions at 512px and costs about 2.3 GB RAM + 2.3 GB VRAM fully resident. With streaming at distance 4 it stays around 170 MB, and in play you can't tell the difference.

**How it works**

Normally region_id doubles as the texture array layer and the position in _region_locations, so any membership change re-enumerates ids and rebuilds the full texture arrays. That makes add/remove at runtime far too expensive to do per frame. While streaming, a region checks out a stable slot from fixed-capacity pooled arrays instead: the region map stores slot + 1, add/evict touch one texture layer and one region map entry, and the arrays are allocated once. The shader's location table is slot-indexed through get_shader_region_locations().

The runtime logic lives in its own subsystem class, Terrain3DStreamer, owned by Terrain3D the same way Terrain3DCollision and Terrain3DInstancer are, with the streaming_* node properties forwarding as aliases like the collision settings do. Terrain3D itself only creates it, calls step() once per physics frame, and exposes the aliases. The slot pool stays in Terrain3DData since it manipulates the region containers and generated texture arrays directly.

The streamer runs a small state machine each physics frame:
- Threaded loads via ResourceLoader (request/poll/get, abandoned requests are always consumed). Nearest regions first, ties broken toward the direction of travel. Corrupt files are never re-requested.
- Insertions are capped per frame (streaming_loads_per_frame), since the texture uploads and signal cascade are the main thread cost. Loads that fell out of range while still loading get dropped.
- Eviction starts one region beyond streaming_distance for hysteresis, up to 8 per frame so teleports recover quickly. Regions modified at runtime are saved back to the data directory first.
- Collision. While streaming, region_map_changed is not wired to build(), because that destroys and recreates the physics body under the player. Insertions instead force an in-place dynamic window update() when the region can intersect it.
- Region size is adopted from the first region file on disk before the pool is built, same as load_directory() does.

**Scope and safety**

Off by default, and is_streaming_active() also checks !Engine.is_editor_hint(), so the editor always keeps full residency and behaves exactly as before. With streaming disabled the load path is untouched and the demo runs clean on the new build.

New inspector group Streaming: streaming_enabled, streaming_shape (square or circle loaded area), streaming_mode (Disk, or RAM Resident where region bodies stay cached in memory and only the GPU layers follow the loaded area, with background prefetch of the rest of the world), streaming_distance, streaming_slots, streaming_concurrent_loads, streaming_loads_per_frame. Everything can be toggled at runtime; the test covers live enable/disable in both directions. All documented in doc_classes. There's also get_streaming_stats() for debug overlays.

The first commit is a separate fix that this feature exposed. Terrain3DMaterial::update() tested p_flags & FULL_REBUILD, which matches every non-zero flag because FULL_REBUILD is a superset mask, so every maps_changed regenerated and recompiled the whole shader. You rarely notice it classically since maps_changed only fires on edits, but with per-frame region traffic it turned into a hitch storm. A dedicated SHADER_REBUILD bit fixes it. Worst streaming frame in our churn test went from ~900 ms to ~17 ms.

**Testing**

- [x] Headless test in project/tests/test_region_streaming.gd (27 asserts: ring residency, eviction, height correctness under churn, slot leak check, runtime-edit save on eviction, circular shape, RAM mode round trip, live enable/disable toggling), green
- [x] Demo boots clean with streaming off
- [x] Soak tested in our driving game on the 700 region world, sweeping 24 km of terrain, ~99% of frames under 20 ms during ring churn
- [ ] Windows/macOS builds untested, Linux only here

**Possible improvements**

Things I left out to keep this reviewable, can follow up on any of them:
- Threaded write-back on eviction. A modified region currently saves synchronously on the main thread before dropping.
- Splitting a single insertion across frames (upload height, then control/color, then flip the region map entry) to shrink the per-insertion cost on large region sizes.
- Instancer support is minimal. MMIs are refreshed per add/evict but instance data streaming hasn't been profiled.
- Some kind of far-field hook so the horizon doesn't pop where regions end. We ship a baked low-res far mesh in our game, but a generic hook probably fits upstream better.
- streaming_slots could be derived from streaming_distance plus in-flight headroom instead of being user-set.

One behavior worth knowing: streamed regions get set_version() on insertion like load_region() does, so files from an older data version are marked modified and get rewritten (upgraded) on eviction. That means the first drive across an old world rewrites its regions one by one. It self-heals after one pass, but if you'd rather not have streaming upgrade files implicitly I can skip the version bump instead.

**Disclosure**

Parts of this PR were developed with AI assistance. All of it was designed, reviewed and tested by hand: the feature has been soak tested in our game for extended play sessions, and the headless test suite plus the demo regression check above were run against every revision.


Saul2022:
Great, will gladly test on Mac once I have  some free time.

Although being pretty critic, you should add a full disclosure on how much AI was used and where just
,so reviewers can focus mostly on  those parts since they will be the most dangerous ones and the ones more prone to make bugs or corrupt the terrain itself.

Saul2022:
Testing on my macbook air m3 8gb of ram  rn in 4.7.1 rc 2 i found some funky behaviour, though the ram savings are so big:

●When you set distance to 15 , in the places the terrain region  don't load back up and stay  as if there wasn't a region done there.



●Moreover , as expected since the load /unload is executed syncronomously , there are  stutters( specially  when the distance is at minimum for obvious reason)

●Adding up to my first statement, when you set the minimum distsnce to 1  and unload regions , sometimes the instances don't disappear along with it's region. 



●Atm from what i can tell this is only working in game only as the editor does not react to changing the streaming values. 


● Aside from the distance stuff and the regions sometimes not loading back up, region streaming has a base cost  of 10% in my computer, which i suppose it's fine and won't matter much when runtime virtual texture is added  , tough  i thought it would give more fps since it's unloading far away regions..

Overall i think this is looking  pretty good( though i insist to put more effort in reviewing the AI code  for safety measures). Also my ram went from like 910 to 220mb when i set distance to 4, which is a great save(  though when setting the  distance  to 1 it has  190mb )

Image-unavailable:
> region streaming has a base cost of 10% in my computer, which i suppose it's fine and won't matter much when runtime virtual texture is added

Actually, the RVT would make the terrain rendering cheaper, which would in turn make the streaming overhead comparatively larger.

I'd be kinda curious to know where the extra frame time is being spent, a visual profiler screenshot would help in that regard.

Saul2022:
> Actually, the RVT would make the terrain rendering cheaper, which would in turn make the streaming overhead comparatively larger.


What do you mean by that? That rvt wouldn’t save fps with  region streaming  on?

Image-unavailable:
> What do you mean by that? That rvt wouldn’t save fps with region streaming on?

I would expect it to save FPS just the same.

It's just that if you experience a +10% frame time increase with streaming on a 100% frame time terrain, and RVT lowers it to say 50%, then the 10% streaming overhead would be a larger chunk of the total cost (50% + 10%).

At least this if it's a GPU bottleneck. If streaming is CPU bottlenecked then that'd be different.

PR #1005, opened 2026-07-18
Move region blend code to Terrain3DData::get_height()

### LINKED ISSUE #949: get_height() needs region blend
### Description

The recent changes to region blend #899 changes the height for collision. get_height() also needs to be updated. The code should probably live in one place.

### LINKED ISSUE #999: Instancer should respect region blend
Instancer needs to respect the region blend, like collision does. Should it rebuild if changed? Probably not.

aidandavey:
Admin edit

* get_height() is now a simple wrapper for get_pixel()
* get_surface_height() adds the material region blend and ground level and interpolates among adjacent vertices
* The UI shows both on the live info panel. The first is necessary for accurate picking
* Added Terrain3DData::world_to_vgrid() as a universal global_position to descaled, floored vertex grid converter which should be used for all functions for consistency in internal use.
* added get_pixel_descaled() and get_modified_height() which take a vgrid for internal use
* Collision, instancer, nav baking, mesh baking call these various new functions
* Improved importer detection and messaging
* Instancer on_collision start height now allows larger than 200m

----------

fixes #949
fixes #999

Moved the region blending code to Terrain3DData::get_height() so that it can be shared when generating collision, or checking the height for any other purpose including instance painting.

Collision generation still passes in the exact vertex so it doesn't need interpolation.  

image

### DIFF
--- a/src/terrain_3d_data.cpp
+++ b/src/terrain_3d_data.cpp
@@ -43,6 +43,45 @@ void Terrain3DData::_copy_paste_dfr(const Terrain3DRegion *p_src_region, const R
 	_terrain->get_instancer()->copy_paste_dfr(p_src_region, p_src_rect, p_dst_region);
 }
 
+Error Terrain3DData::_save_export_image(const MapType p_map_type, const Ref<Image> &p_img, const String &p_path,
+		const String &p_ext) const {
+	if (p_map_type == TYPE_HEIGHT && (p_ext == "r16" || p_ext == "raw")) {
+		Ref<FileAccess> file = FileAccess::open(p_path, FileAccess::WRITE);
+		if (file.is_null()) {
+			return ERR_CANT_OPEN;
+		}
+		Vector2 minmax = _master_height_range;
+		if (minmax.x >= minmax.y) {
+			// Global height range wasn't calculated, fall back to range for this image
+			minmax = Util::get_min_max(p_img);
+			LOG(MESG, "  Height range (per image fallback): ", vformat("%.2f", minmax.x), " to ", vformat("%.2f", minmax.y));
+		} else {
+			LOG(MESG, "  Height range (global scale): ", vformat("%.2f", minmax.x), " to ", vformat("%.2f", minmax.y));
+		}
+		float range = minmax.y - minmax.x;
+		real_t hscale = 65535.0 / (range != 0.f ? range : 1.f);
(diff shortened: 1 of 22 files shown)


src/terrain_3d_data.cpp
```diff
@@ -727,6 +733,40 @@ bool Terrain3DData::is_in_slope(const Vector3 &p_global_position, const Vector2
 	return (slope_range.x <= slope_angle_degrees) && (slope_angle_degrees <= slope_range.y);
 }
 
+real_t Terrain3DData::get_region_blend(const Vector3 &p_global_position) const {
+	auto check_region = [&](const Vector2 &uv2) -> real_t {
```
TokisanGames: This function already exists as get_region_map_index()

src/terrain_3d_data.cpp
```diff
@@ -673,8 +674,13 @@ real_t Terrain3DData::get_height(const Vector3 &p_global_position) const {
 		real_t ht10 = get_pixel(TYPE_HEIGHT, pos10).r;
 		Vector3 pos11 = pos00 + Vector3(step, 0.f, step);
 		real_t ht11 = get_pixel(TYPE_HEIGHT, pos11).r;
-		return bilerp(ht00, ht01, ht10, ht11, pos00, pos11, pos);
+		height = bilerp(ht00, ht01, ht10, ht11, pos00, pos11, pos);
 	}
+
+	if (!std::isnan(height) && _terrain->get_material()->get_world_background() != Terrain3DMaterial::WorldBackground::NONE) {
+		height = Math::lerp(height, _terrain->get_material()->get("ground_level"), smoothstep(0.f, 1.f, get_region_blend(pos)));
```
TokisanGames: if ground_level or region_blend aren't in the shader, this returns Variant::NIL which turns into 0, which may be a problem for the math. It's better to not do the math at all.

TokisanGames (commented):
This worked well, under most cases, but there was a bug when using vertex spacing that is not a multiple of 0.25. and would cause weird artifacts; get_height() returned slightly wrong values in stripes causing the collision to appear a little lower than the terrain.

Doing it with this implementation introduces a minor performance hit as it has to scale up and then descale again, which can add up. That floating point precision and rounding likely introduced the problem above.

I have instead made descaled versions of the functions that avoid both of these circumstances.

I've taken this one over.

TokisanGames:
Good work. I've updated Op with notes.

PR #1008, opened 2026-07-26
Fix regions silently failing to save after an interrupted brush stroke

### LINKED ISSUE #34: Scene Builder clobbers Terrain3D's hotkeys, causing data loss due to a confused undo stack
Please remove SceneBuilder while editing Terrain3D terrain for now. As it is, SceneBuilder will occasionally cause your edited work to not save.

sci-comp:
If a brush stroke ends without `stop_operation()` running, the affected regions are left permanently flagged as edited. This means `backup_region()` skips them, which means they are never marked modified, never written to disk, and never recorded for undo operations.

This bug will help resolve https://github.com/sci-comp/scene-builder/issues/34 in my level design asset, which is how I found this bug (it's easy to intercept Terrain3D operations when another plugin is listening for hotkeys).

### DIFF
--- a/src/terrain_3d_editor.cpp
+++ b/src/terrain_3d_editor.cpp
@@ -906,16 +906,14 @@ void Terrain3DEditor::set_operation(const Operation p_operation) {
 // Called on mouse click
 void Terrain3DEditor::start_operation(const Vector3 &p_global_position) {
 	IS_DATA_INIT_MESG("Terrain isn't initialized", VOID);
+	// In case mouse-up was intercepted (by a modal dialog, focus change, or a raycast miss, etc...)
+	stop_operation();
 	LOG(INFO, "Setting up undo snapshot");
 	_undo_data.clear();
 	_undo_data["region_locations"] = _terrain->get_data()->get_region_locations().duplicate();
 	_is_operating = true;
-	_original_regions = TypedArray<Terrain3DRegion>(); // New pointers instead of clear
-	_edited_regions = TypedArray<Terrain3DRegion>();
-	_added_removed_locations = TypedArray<Vector2i>();
 	// Reset counter at start to ensure first click places an instance
 	_terrain->get_instancer()->reset_density_counter();
-	_terrain->get_data()->clear_edited_area();
 	_operation_position = p_global_position;
 	_operation_movement = V3_ZERO;
 }


TokisanGames:
Our environment artist has reported an elusive bug where sometimes regions get in a state where they won't save. We haven't been able to find the cause, so this is a very welcome contribution.

TokisanGames:
Is it better to clear the flag, or is it better to call stop_operation() which will clear the flag and store the undo? 🤔

sci-comp:
would it be this simple? (updated the commit)

src/terrain_3d_editor.cpp
```diff
@@ -906,6 +906,12 @@ void Terrain3DEditor::set_operation(const Operation p_operation) {
 // Called on mouse click
 void Terrain3DEditor::start_operation(const Vector3 &p_global_position) {
 	IS_DATA_INIT_MESG("Terrain isn't initialized", VOID);
+
+	// In case mouse-up was intercepted (by a modal dialog, focus change, or a raycast miss, etc...)
+	if (_is_operating) {
+		stop_operation();
+	}
+
```
TokisanGames: stop_operation() already checks if is_operating.  And it does most of the things below, which can all be removed. Keep only what this does that stop_operation() doesn't.

Also please test it. You can trigger the bug on demand, right? Please make sure doing this mints the undo and doesn't cause other strange things.

sci-comp: Okay, I removed what look like the duplicate lines, and then I compiled and tested. It seems to work! The bug is easy to reproduce,

1) Create a scene with two regions and at least 2 textures. Paint between the seam of the two regions.
2) While mouse is held down, mid brush stroke: press Ctrl+Shift+P to open Godot's command pallete. This will intercept the operation and trigger the failed state.
3) Any Ctrl+Z will now undo all changes on one of the two regions.

TokisanGames: Ok, looks good.
* Please add this back in, `_undo_data.clear();`  right before the other _undo_data line. It's redundant, but will be more clear.
* Remove the space before the `// in case`
* Then squash all your commits into one and I'll merge it.

sci-comp: Alright, I added that line back in, removed the white space, and squashed.

TokisanGames:
Excellent, thank you!

PR #1010, opened 2026-07-29
Fix dummy texture RID leak on reinitialization

TheXenOne:
Terrain3DMaterial::initialize() can run repeatedly when assets or the data directory change at runtime. Each call recreated the dummy texture array and orphaned its previous RID. Reuse the existing valid dummy texture instead.

### DIFF
--- a/src/terrain_3d_material.cpp
+++ b/src/terrain_3d_material.cpp
@@ -827,7 +827,8 @@ void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
 	}
 	_shader.instantiate();
 	_buffer_shader.instantiate();
-	if (!_generated_dummy.get_rid().is_valid()) { // Create dummy texture array to avoid empty sampler2DArrays
+	// Create dummy texture array to avoid empty sampler2DArrays
+	if (!_generated_dummy.get_rid().is_valid()) {
 		Ref<Image> img = Image::create(1, 1, false, Image::FORMAT_RF);
 		TypedArray<Image> ia = { img };
 		_generated_dummy.create(ia);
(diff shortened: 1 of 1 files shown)


src/terrain_3d_material.cpp
```diff
@@ -827,7 +827,9 @@ void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
 	}
 	_shader.instantiate();
 	_buffer_shader.instantiate();
-	{ // Create dummy texture array to avoid empty sampler2DArrays
+	if (!_generated_dummy.get_rid().is_valid()) {
+		// Runtime asset or data-directory changes reinitialize this material.
+		// Reuse the dummy texture instead of orphaning its previous RID.
```
TokisanGames: Please reinstate my comment which explains the code.
Your comments are helpful for me reviewing this, but don't explain the code so can be removed.

Also why don't you flip the conditional:
`if(_generated_dummy.get_rid().is_null())`

TheXenOne: I've put your comment back in and removed mine.
I believe is_null() is only available on Godot's internal RID type, not through godot-cpp, so I've kept the equivalent !is_valid() check.

src/terrain_3d_material.cpp
```diff
@@ -827,7 +827,7 @@ void Terrain3DMaterial::initialize(Terrain3D *p_terrain) {
 	}
 	_shader.instantiate();
 	_buffer_shader.instantiate();
-	{ // Create dummy texture array to avoid empty sampler2DArrays
+	if (!_generated_dummy.get_rid().is_valid()) { // Create dummy texture array to avoid empty sampler2DArrays
```
TokisanGames:
```suggestion
	// Create dummy texture array to avoid empty sampler2DArrays
	if (!_generated_dummy.get_rid().is_valid()) {
```

TokisanGames:
Great, thank you!

PR #1011, opened 2026-07-30
Quality of Life Features

mochoindiestudio:
- Add texture pair import dialog and matcher for bulk importing albedo/normal textures
- Users can now drag-and-drop multiple files into the textures and meshes
- Added editor's own "Undo/Redo" (history) registration for user changes in these two areas


TokisanGames:
Please clean up your pull request. Remove the 550+ file changes that don't need to be there, and squash your commits. Also you should only do a PR against `main`, unless directed otherwise.

PR #1015, opened 2026-08-03
Add slope point selection hotkeys

levidavidmurray:
Adds shortcuts for individual slope point selection.

When the slope tool is active, `KEY_1` selects the first gradient point for picking, `KEY_2` selects the second.

This PR introduces `OperationBuilder` input/hotkey consumption, which is hopefully a clean enough approach for scoping hotkeys to certain tools and operations.

This version currently overrides the default region grid toggle (`KEY_1`) and region label toggle (`KEY_2`) when the slope tool is selected. I just personally find `KEY_1` and `KEY_2` far too convenient to limit to debug toggles, but this is also something that can be addressed once https://github.com/TokisanGames/Terrain3D/pull/983 is merged.

https://github.com/user-attachments/assets/970d2361-bb70-4037-acd9-e6134685b343

### DIFF
--- a/project/addons/terrain_3d/src/editor_plugin.gd
+++ b/project/addons/terrain_3d/src/editor_plugin.gd
@@ -233,6 +233,7 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) -> 
 				var height: float = terrain.data.get_height(mouse_global_position)
 				ui.brush_data["height"] = height
 				ui.tool_settings.set_setting("height", height)
+				return AFTER_GUI_INPUT_STOP
 				
 			# If adjusting regions
 			if editor.get_tool() == Terrain3DEditor.REGION:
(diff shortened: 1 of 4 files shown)


project/addons/terrain_3d/src/editor_plugin.gd
```diff
@@ -233,6 +233,7 @@ func _forward_3d_gui_input(p_viewport_camera: Camera3D, p_event: InputEvent) ->
 				var height: float = terrain.data.get_height(mouse_global_position)
 				ui.brush_data["height"] = height
 				ui.tool_settings.set_setting("height", height)
+				return AFTER_GUI_INPUT_STOP
```
levidavidmurray, line 236: Oh, I also snuck in this small fix. Current behaviour allows you to sample height via `Ctrl+LMB`, but doing so also affects the terrain. I didn't notice any inverted operations on HEIGHT tool operations, so it seemed safe enough to return early for height sampling.

PR #1017, opened 2026-08-07
Fix ambiguous overload resolution in double-precision (precision=double) builds

Empt-y:
## Summary
Fixes a real compile failure (MSVC error C2668) when building Terrain3D against a Godot editor compiled with `precision=double`, following the steps in the double precision docs.

`get_modified_height()` and `get_texture_id()` in `terrain_3d_data.cpp` call `is_hole()`, `is_auto()`, `get_base()`, `get_overlay()`, and `get_blend()` with a `real_t` control-map value. These utility functions in `terrain_3d_util.h` are overloaded on `float` and `uint32_t`. The control map value is bit-packed into a float32 regardless of engine precision, but when `real_t` is `double` (a `precision=double` build), passing it directly makes the call genuinely ambiguous - the compiler can't choose between the implicit `double`->`float` and `double`->`uint32_t` conversions, and fails to compile.

## Fix
Explicit `static_cast<float>(control)` at each of the six call sites, since the underlying value is always float32-encoded regardless of `real_t`. No behavior change for the default single-precision build, where `real_t` already is `float` (the cast is a no-op there).

## Test plan
- [x] Reproduced the failure building against a from-source `precision=double` Godot 4.7.1 editor (Windows/MSVC) with custom-generated godot-cpp bindings
- [x] Applied the fix, confirmed a clean build (`libterrain.windows.debug.x86_64.dll` links successfully)
- [ ] Maintainers may want to verify runtime behavior in an actual double-precision project, since I don't have a full double-precision test scene set up yet

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -455,8 +455,9 @@ void Terrain3D::_generate_triangle_pair(PackedVector3Array &p_vertices, PackedVe
 		}
 	}
 
-	// Get control pixels
-	real_t val = _data->get_pixel_descaled(TYPE_CONTROL, v1g).r;
+	// Get control pixels. Always float: control map is packed as a float32
+	// Color component regardless of engine precision.
+	float val = _data->get_pixel_descaled(TYPE_CONTROL, v1g).r;
 	uint32_t ctrl1 = (std::isnan(val)) ? UINT32_MAX : as_uint(val);
 	val = _data->get_pixel_descaled(TYPE_CONTROL, v2g).r;
 	uint32_t ctrl2 = (std::isnan(val)) ? UINT32_MAX : as_uint(val);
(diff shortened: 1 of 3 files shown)


TokisanGames:
Thanks for bringing it to my attention. The real fix should be:
```
-real_t control = get_pixel_descaled(TYPE_CONTROL, p_vgrid).r;
+float control = get_pixel_descaled(TYPE_CONTROL, p_vgrid).r;
```

Color r,g,b,a are always floats, and we always want them to be floats. Please review where I've retreived `.r` and assigned it to real_t and you can make the above changes.

Casting a Control float is always the wrong choice as it is a 32-bit memory block and not actually a float. We don't ever want the compiler to take a 4 byte memory block, cast it to an 8 byte memory block and back to a 4 byte via FPU interpretation.

Looks like all of these are wrong. Please fix them all and squash your changes.
```
$ egrep -rIn 'real_t.+[cC][oO][nN][tT][rR][oO][lL].+\.r' src/
src/terrain_3d.cpp:459: real_t val = _data->get_pixel_descaled(TYPE_CONTROL, v1g).r;
src/terrain_3d_data.cpp:709:    real_t control = get_pixel_descaled(TYPE_CONTROL, p_vgrid).r;
src/terrain_3d_data.cpp:827:    real_t control = get_pixel_descaled(TYPE_CONTROL, vgrid).r;
src/terrain_3d_data.h:305:      real_t val = get_pixel(TYPE_CONTROL, p_global_position).r;
```

TokisanGames:
Thank you!

PR #1019, opened 2026-08-07
Add Terrain3D::rebase() for floating origin support

Empt-y:
## Summary
Godot's `real_t`/`Node3D` transforms are float32 in default (single-precision) builds, which loses sub-unit precision at large world coordinates (10s-100s of thousands+ units from origin) — the standard fix is a floating origin, where the host application periodically shifts the camera/player and every other dynamic `Node3D` back near true origin. Terrain3D previously had no way to participate: none of its own position math reads from its own `Node3D` transform, so its baked `PhysicsServer3D` collision shapes, `MultiMeshInstance3D` foliage/rock transforms, and region data all stay anchored to their original absolute coordinates even if the rest of the scene shifts, causing terrain to visibly desync from everything else the moment a host app rebases.

Adds `Terrain3D::rebase(Vector3 delta)` as the participation hook: a host calls this with the same delta it applies to the rest of the scene.

The delta must be a multiple of `region_size * vertex_spacing` on X/Z (Y is unconstrained; height data isn't addressed by world Y — a misaligned delta is rounded to the nearest region and logged). This constraint is what makes the implementation cheap: every region's pixel content is byte-for-byte identical after a region-aligned shift, only its address changes, so `Terrain3DData::rebase()` just relabels region_location keys (duplicating each `Terrain3DRegion` shallowly — shares the underlying map Images, no pixel copy — since `remove_region()`'s soft-delete semantics mean mutating and re-adding the same object resurrects its old dictionary entry too) rather than resampling like `change_region_size()` does for actual size changes.

Collision, instancer, and mesher all turned out to need no bespoke shifting logic: collision shapes are regenerated fresh from height data on `update(V2I_MAX, true)`, `MultiMeshInstance3D` transforms are recomputed from `region_loc * region_size * vertex_spacing` on every `update_mmis()` rebuild with per-instance transforms already stored region-relative, and the clipmap mesher already recenters every `snap()` from the live tracked target position — so region re-keying plus the existing full-rebuild entry points already used by `change_region_size()` cover all of it.

## Test plan
- [x] Real headless run against a built `libterrain.so`, not just a compile check: imported a real 2-region deterministic height field plus a foliage instance, rebased by exactly one region, confirmed every height sample and the instance both landed at the correctly shifted coordinates with the old coordinates now empty, then rebased back and confirmed an exact round-trip restore.
- [x] Confirmed a misaligned, sub-region delta logs a warning rather than silently doing nothing.
- [x] Extensive edge-case coverage: out-of-bounds rebase correctly aborts atomically (no partial state change), 4 sequential cumulative rebases net to the correct position with height matching at every step, save/load round-trip through a rebase (old region's file deleted, new files correct, edit made post-rebase persists through save/reload), and fractional `vertex_spacing` (2.5) with both aligned and misaligned deltas.
- [x] Real demo project verification: real 3-region terrain, real player physics, rebased by 2048 units — visually confirmed via screenshot (different, correct biome content at the new location, no rendering corruption) and confirmed two simultaneously-spawned players' render-space distance matched their true-space distance exactly.

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -743,6 +743,50 @@ void Terrain3D::snap() {
 	}
 }
 
+// Shifts the whole terrain (region data, collision, instanced meshes, render mesh) by p_delta, so a
+// host application can implement floating origin: periodically move the camera/player and every other
+// dynamic Node3D back near true origin to avoid float32 precision loss at large world coordinates, and
+// call this so the terrain's own absolute-position state (baked collision shapes, MultiMesh instance
+// transforms, region data) stays correctly positioned relative to everything else that moved.
+//
+// p_delta must be a multiple of region_size * vertex_spacing on X and Z (Y is ignored; height data
+// isn't addressed by world Y). This is what lets region data be relabeled instead of resampled: every
+// region's pixel content is unchanged, only which world location it's addressed at changes. A host
+// wanting an arbitrary, non-region-aligned floating origin threshold should round the shift they apply
+// to the rest of the scene down to this same grid, not shift the terrain by an unaligned amount.
+void Terrain3D::rebase(const Vector3 &p_delta) {
+	if (!_initialized || !_data) {
+		return;
+	}
+	real_t grid_unit = real_t(_region_size) * _vertex_spacing;
+	Vector2i region_offset(
+			int(Math::round(p_delta.x / grid_unit)),
+			int(Math::round(p_delta.z / grid_unit)));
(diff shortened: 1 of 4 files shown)


TokisanGames:
Multiple people have attempted Origin shifting on a pixel basis and gave up because it's unnecessarily complex. But origin shifting by region coordinates is brilliant. To do it effectively, we need region streaming and to remove the region map outbound limits. 

I'm very close to publishing 1.1, so we'll look at this and the others for 1.2.

A similar technique was also implemented in this demo and worth looking at when  this is reviewed.
https://github.com/Inzarcon/Terrain3D-InfiniteProceduralDemo

PR #1020, opened 2026-08-07
Add real disk-based region streaming (issue #491, partial)

Empt-y:
## Summary
`Terrain3DData::load_directory()` currently loads every region file in a directory unconditionally, forever — there was no way to load/unload by distance at all. `update_streaming(target_positions, radius)` fixes that: regions load in and (after saving, if modified) unload as they cross `radius` of any tracked target.

`target_positions` is a list, not a single position, and the union across all of them is computed fully before any eviction happens — this is what keeps it safe for multiple simultaneous players/cameras: a region only unloads if it's out of range of *every* target, never based on evaluating targets independently, which could evict a region a different target still needs. `Terrain3D::_update_streaming()` gathers the usual clipmap/collision/camera targets plus anything registered via the new `add_streaming_target()`/`remove_streaming_target()` (e.g. other connected players on a server) into one call for exactly this reason.

New `unload_region()` is deliberately **not** `remove_region()`: `remove_region()` soft-deletes for the editor's undo/redo-aware Add/Remove Region workflow (keeps the object in `_regions` until the *next* save notices `is_deleted()` and actually erases it), which is wrong for streaming eviction on two counts — it wouldn't actually free memory until some later unrelated save, and it would mark the file for disk deletion, when the file should be left alone. `unload_region()` does a clean, immediate removal from both `_regions` and `_region_locations`, disk file untouched.

Save-before-evict re-checks `is_modified()` *after* calling `save_region()`, not just before: `Terrain3DRegion::save()` only clears that flag on success, so if it's still set the save genuinely failed, and eviction is skipped rather than risking silent data loss — the region stays loaded and gets retried next cycle.

## What this deliberately does not attempt
Region addressing / `REGION_MAP_SIZE` are untouched — a region can only stream in within the same existing ±16 range. I found that removing that bound and making the region map relative to camera position both run into the same deeper problem: `get_region_id()` (used by every collision/instancer/editor/gameplay query) reads directly from the same `_region_map` array that's uploaded as the literal GLSL shader uniform (`main.glsl`/`displacement_buffer.glsl`: `uniform int _region_map[1024]`) — not similar to it, the *same* array serving two different jobs. Decoupling those is real, careful work on the single hottest path in the codebase, and affects every consumer of `Terrain3DData`. I've posted a design writeup on #491 with what I found rather than attempting that unilaterally in this PR.

## Test plan
- [x] Real headless test, four scenarios: (1) no-op with no `data_directory` set, confirming pure in-memory/procedural use (e.g. `CodeGeneratedDemo`) is unaffected.
- [x] (2) Full round trip — edited a region, streamed it out of range, confirmed via an independent fresh disk load (not the same terrain instance) that the edit actually made it to disk before eviction, then streamed back in and confirmed the edit survived.
- [x] (3) Two targets near two different regions, confirming both stay loaded simultaneously (real union, not last-target-wins).
- [x] (4) Regression test for a real bug caught in review — the region-search radius wasn't scaled by `vertex_spacing`, only `region_size`, so at `vertex_spacing < 1.0` (a normal setting) regions genuinely inside the configured radius would be silently missed/evicted. Confirmed the fix by reverting it and rerunning: exactly the new test's check failed, nothing else did; restored the fix and confirmed all 20 checks across all four scenarios pass.

### DIFF
--- a/src/terrain_3d.cpp
+++ b/src/terrain_3d.cpp
@@ -146,6 +146,13 @@ void Terrain3D::__physics_process(const double p_delta) {
 	if (_collision && _collision->is_dynamic_mode()) {
 		_collision->update();
 	}
+	// Streaming doesn't need frame-perfect responsiveness the way LOD/collision
+	// snapping does, and each call can touch disk (FileAccess::file_exists per
+	// candidate region) -- throttled to a modest cadence rather than every
+	// physics tick, to keep that cost bounded regardless of physics tick rate.
+	if (_region_streaming_radius > 0.f && Engine::get_singleton()->get_physics_frames() % 30 == 0) {
+		_update_streaming();
+	}
 }
 
 /**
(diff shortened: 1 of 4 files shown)


TokisanGames:
#1004, #1006 also implement some region streaming. We'll look at it all in 1.2.

PR #1038, opened 2026-09-10
Fix: std::isfinite check wrong var name

jamonholmgren:
In Terrain3DData, I noticed that the `if (!std::isfinite(...)) {` check repeats the `hx` check again instead of using `hz` like it should.

This is pretty minor, but probably worth fixing.


TokisanGames:
Thank you.

PR #1039, opened 2026-09-10
Optimize collision shape generation by 16x or more

jamonholmgren:
With my game, I have long views and distant vehicles that need to drive and interact with the player from many kilometers away (often up to 15 km), so full (game) collision generation is best.

However, it currently takes upwards of 10-25 seconds to generate on map load (at 1024 regions), which isn't great.

Collision generation calls `get_modified_height()` for every generated vertex. For a 1024 region map with 256x256 regions, that amounts to about 67.6 million calls, which repeat the same internal calls to get the same info: resolve the region, height, control maps, material, background mode, ground_level, region_blend.

This PR changes it so fetching that data happens outside the loop (preloads it) for its region and nearby regions, and then generates the vertices. I didn't do anything with background blending (still uses the existing blending calculation).

I benchmarked it and went from 22 seconds to 1 second on my Windows computer, similar ratio on macOS:

```
Machine  Regions  Current ms  Patched ms  Speedup
M1 Max      1,024  11,448.494     401.981   28.48x
Windows     1,024  22,613.607   1,025.145   22.06x
M1 Max          1      10.630       0.359   29.61x
Windows         1       8.310       0.512   16.23x
```

M1 Max = macOS with 64 GB RAM
Windows = Windows 11 liquid-cooled gaming laptop, i9-12900H

One downside is that some of the logic is duplicated, since it's pretty hard to untangle it without complicating the original code.

### DIFF
--- a/src/terrain_3d_collision.h
+++ b/src/terrain_3d_collision.h
@@ -48,6 +48,8 @@ private:
 
 	Vector2i _snap_to_grid(const Vector2i &p_pos) const;
 	Vector2i _snap_to_grid(const Vector3 &p_pos) const;
+	real_t _get_modified_collision_height(const Vector2i &p_vgrid, const Vector2i &p_region_loc, const int p_region_size, const real_t p_region_texel_size,
+			const float *const *p_height_maps, const float *const *p_control_maps, const bool p_blend, const real_t p_ground_level, const real_t p_region_blend) const;
 	Dictionary _get_shape_data(const Vector2i &p_position, const int p_size);
 
 	void _shape_set_disabled(const int p_shape_id, const bool p_disabled);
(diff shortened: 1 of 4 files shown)


src/terrain_3d_collision.h
```diff
@@ -48,6 +48,8 @@ class Terrain3DCollision : public Object {
 
 	Vector2i _snap_to_grid(const Vector2i &p_pos) const;
 	Vector2i _snap_to_grid(const Vector3 &p_pos) const;
+	real_t _get_modified_collision_height(const Vector2i &p_vgrid, const Vector2i &p_region_loc, const int p_region_size, const real_t p_region_texel_size,
+			const float *const *p_height_maps, const float *const *p_control_maps, const bool p_blend, const real_t p_ground_level, const real_t p_region_blend) const;
```
jamonholmgren, line 52: Note: I thought about using a `struct` here to package all this up and modularize the code a bit, but there's a couple things I'm unsure of:

1. It could possibly have a runtime penalty, unless the compiler notices and optimizes it somehow
2. You don't really use that many structs in the code base, and I wanted to stay within the feel of the project
3. It wouldn't really add much outside of making the code read a bit better.

